PR #30, opened 2023-05-16
Touch up release CI job

gordonhart:
In preparation for today's 0.69.0 release (the first to be cut out of this repository!), run the `release.yml` CI through its paces, publishing both `kolena` and `kolena-client` to test PyPI and trunk CodeArtifact. Verified that both work by installing `kolena`/`kolena-client` from test PyPI in a fresh venv.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,7 +33,7 @@ jobs:
           aws-region: us-west-2
 
       - name: Install Poetry
-        run: pip install poetry==1.2.2
+        run: pip install 'poetry>=1.3'
 
       - name: Fetch tags to enable autoversioning
         run: git fetch --prune --unshallow --tags
@@ -50,7 +50,7 @@ jobs:
       - name: Build 'kolena' Python package
         run: poetry build --format=sdist
 
-      - name: Build 'kolena-client' package for backwards compatibility
+      - name: Build 'kolena-client' package for backward compatibility
         run: |
           # update first instance of 'kolena' to 'kolena-client' in pyproject.toml, kolena/__init__.py (package name)
           sed -i '0,/kolena/{s/kolena/kolena-client/}' pyproject.toml kolena/__init__.py
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,9 +1,10 @@
 name: Build
 
 on:
-  push:
+  pull_request:
     branch:
       - gh/pre-release
+  push:
     tags:
       - "*"
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,9 +1,6 @@
 name: Build
 
 on:
-  pull_request:
-    branch:
-      - gh/pre-release
   push:
     tags:
       - "*"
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -69,35 +63,59 @@ jobs:
...
-          twine upload --skip-existing --repository codeartifact ./dist/kolena_cilent*
-
-      - name: Create GitHub release with kolena build as artifact
-        uses: marvinpinto/action-automatic-releases@v1.2.1
-        with:
-          repo_token: ${{ secrets.RELEASE_TOKEN }}
-          prerelease: false
-          files: |
-            ./dist/*.tar.gz
+          # twine upload --skip-existing --repository codeartifact ./dist/kolena_cilent*
```
munkyshi: Should this be uncommented?

gordonhart: Yes, thanks for catching. Uncommented

yifanwu-kolena:
i think we should setup the correct version at [here](https://github.com/kolenaIO/kolena/blob/c78d94972cc4c509aa6d4b72572cf07242ba9cb8/pyproject.toml#L3) as well

gordonhart:
> i think we should setup the correct version at [here](https://github.com/kolenaIO/kolena/blob/c78d94972cc4c509aa6d4b72572cf07242ba9cb8/pyproject.toml#L3) as well

This is set automatically by the [release CI job](https://github.com/kolenaIO/kolena/blob/gh/pre-release/.github/workflows/release.yml#L39):

```yaml
      - name: Update kolena package version to PEP 440-compliant production release tag
        run: poetry version $(git describe --tags --abbrev=0)
```

yifanwu-kolena:
> > i think we should setup the correct version at [here](https://github.com/kolenaIO/kolena/blob/c78d94972cc4c509aa6d4b72572cf07242ba9cb8/pyproject.toml#L3) as well
> 
> This is set automatically by the [release CI job](https://github.com/kolenaIO/kolena/blob/gh/pre-release/.github/workflows/release.yml#L39):
> 
> ```yaml
>       - name: Update kolena package version to PEP 440-compliant production release tag
>         run: poetry version $(git describe --tags --abbrev=0)
> ```

cool, i thought `pyproject.toml` is part of the pypi artifacts, but seems like not, i think it's fine to leave as is.

.github/workflows/release.yml
```diff
@@ -69,35 +63,59 @@ jobs:
...
 
-      - name: Push kolena, kolena-client dists to PyPI
+      - name: Push 'kolena' dist to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           password: ${{ secrets.PYPI_API_TOKEN }}
 
-      - name: Install twine for package distribution on CodeArtifact
-        run: pip3 install twine
+      - name: Create GitHub release with 'kolena' build as artifact
```
ly-kolena, line 77: Nit: do we need this given current release flow? Every release is published to pypi, and currently no non-release build.

.github/workflows/release.yml
```diff
@@ -69,35 +63,59 @@ jobs:
...
+      #
+
+      - name: Build 'kolena-client' package for backward compatibility
+        run: |
+          rm -rf ./dist
+
+          # update first instance of 'kolena' to 'kolena-client' in pyproject.toml, kolena/__init__.py (package name)
+          sed -i '0,/kolena/{s/kolena/kolena-client/}' pyproject.toml kolena/__init__.py
+
+          poetry install
```
ly-kolena, line 96: Should this be run again, or does it need to? It seems I can just run `poetry build` without additional commands?

ly-kolena (approved):

.github/workflows/release.yml
gordonhart: We don't really need the artifact, but I figure it's good to include anyway. We definitely want to keep this action in general so that automated releases with changelogs are created when tags are pushed.

.github/workflows/release.yml
gordonhart: Updating the package name causes `poetry` to use a new environment, as environments are keyed on package name. Had a fun time debugging this one, which was [manifesting as `Command not found: sphinx-apidoc` in the following job step](https://github.com/kolenaIO/kolena/actions/runs/4995591035/jobs/8947648228).

.github/workflows/release.yml
gordonhart: Ended up removing the artifact in #32. It isn't necessary when the client is shipped via PyPI

PR #31, opened 2023-05-16
Use client_state consistently across operations

ly-kolena:

Use client_state, either static module member or contextvar, for connection information. Previously `get_token` access environment variable directly, leading using values different from initialized state (because env variable is updated after client_state is initialized).

Note: this does not address the problem that changing `API_BASE_URL` environment variable after importing kolena will not take effect.

### DIFF
--- a/kolena/_utils/state.py
+++ b/kolena/_utils/state.py
@@ -102,11 +102,11 @@ class _ClientState:
         self.telemetry = False
 
 
-def get_client_base_url() -> str:
+def _get_api_base_url() -> str:
     return os.environ.get(API_BASE_URL_ENV_VAR, API_BASE_URL)
 
 
-_client_state = _ClientState(base_url=get_client_base_url())
+_client_state = _ClientState(base_url=_get_api_base_url())
 
 
 def get_client_state() -> _ClientState:
@@ -160,7 +160,7 @@ def get_token(
 
 @contextlib.contextmanager
 def kolena_session(api_token: str, base_url: Optional[str] = None) -> Iterator[_ClientState]:
-    base_url = base_url or get_client_base_url()
+    base_url = base_url or _get_api_base_url()
     init_response = get_token(api_token, base_url)
     client_state = _ClientState(
         base_url=base_url,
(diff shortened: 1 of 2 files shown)


munkyshi (approved):

kolena/_utils/state.py
```diff
@@ -142,7 +142,7 @@ def get_token(
     base_url: Optional[str] = None,
     proxies: Optional[Dict[str, str]] = None,
 ) -> API.ValidateResponse:
-    base_url = base_url or get_client_base_url()
+    base_url = base_url or get_client_state().base_url
```
ly-kolena, line 145: After fix, the only direct invocation of `get_client_base_url` outside of initializing client_state is in `kolena_session`, which is intended usage.

tests/unit/test_initialize.py
```diff
@@ -75,6 +81,22 @@ def test__initialize__deprecated_positional(clean_client_state: None) -> None:
...
+    mock_response._content = str.encode(json.dumps(dataclasses.asdict(FIXED_TOKEN_RESPONSE)))
+    with patch.dict("os.environ", {API_BASE_URL_ENV_VAR: base_url}, clear=True):
+        with patch("requests.put", return_value=mock_response) as patched:
+            kolena.initialize("random entity", "def")
+            client_state = get_client_state()
+            assert client_state.api_token == "def"
+            assert client_state.base_url == API_BASE_URL
+
+        patched.assert_called_once()
+        assert patched.call_args.args[0] == get_endpoint_with_baseurl(API_BASE_URL, "token/login")
```
ly-kolena: This test would fail before fix as:

```
            patched.assert_called_once()
>           assert patched.call_args.args[0] == get_endpoint_with_baseurl(API_BASE_URL, "token/login")
E           AssertionError: assert 'https://internal-api.kolena.io/v1/token/login' == 'https://api.kolena.io/v1/token/login'
E             - https://api.kolena.io/v1/token/login
E             + https://internal-api.kolena.io/v1/token/login
E             ?         +++++++++

test_initialize.py:98: AssertionError
```

kolena/_utils/state.py
gordonhart: Thoughts on removing `get_client_base_url` so as to not encourage other misusage down the line?

ly-kolena: Marked it as private as an alternative, `_get_api_base_url`, since I think it serves a purpose to avoid remembering env variable name.

PR #32, opened 2023-05-16
Idempotent PyPI publishing in release CI

gordonhart:
Add `skip-existing: true` to each PyPI publish in `release.yml` to make this job rerunnable. Also use stock `GITHUB_TOKEN` for `action-automatic-release`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -65,16 +65,18 @@ jobs:
           aws s3 sync ./build "$BUCKET_PATH"
         working-directory: ./docs
 
-      # - name: Push 'kolena' dist to Test PyPI
-      #   uses: pypa/gh-action-pypi-publish@release/v1
-      #   with:
-      #     password: ${{ secrets.TEST_PYPI_API_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -68,16 +68,18 @@ jobs:
         with:
           password: ${{ secrets.TEST_PYPI_API_TOKEN }}
           repository-url: https://test.pypi.org/legacy/
+          skip-existing: true
```
ly-kolena, line 71: C&P for visibility, in case there's additional considerations I missed.

> Pro tip: try to avoid enabling this setting where possible. If you have steps for publishing to both PyPI and TestPyPI, consider only using it for the latter, having the former fail loudly on duplicates.

If we're scoped to only run the workflow on release, risk is low?

gordonhart: I think the risk is low in our case. This "protip" makes me scratch my head — it's really difficult to put together automations when actions aren't idempotent.

ly-kolena (approved):

PR #37, opened 2023-05-17
Fix result order before assertion

ly-kolena:

### DIFF
--- a/tests/integration/detection/test_model.py
+++ b/tests/integration/detection/test_model.py
@@ -87,9 +87,9 @@ def test__load_inferences__no_inferences(detection_test_data: TestData) -> None:
     # test-cases[0] has sample (sample_0, sample_1, sample_2, sample_4), gt (gt_0, None, gt_2, None,)
     # test-cases[1] has sample (sample_2, sample_2, sample_3), gt (gt_1, gt_2, gt_4,)
     # test sample #3 across test suite, should have different ground_truths
-    sample, _ = inferences[test_case_id_0][2]
+    sample, _ = sorted(inferences[test_case_id_0], key=lambda x: x[0].locator)[2]
     assert len(sample.ground_truths) == 1
-    sample, _ = inferences[test_case_id_1][0]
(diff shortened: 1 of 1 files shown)


tests/integration/detection/test_model.py
```diff
@@ -87,9 +87,9 @@ def test__load_inferences__no_inferences(detection_test_data: TestData) -> None:
     # test-cases[0] has sample (sample_0, sample_1, sample_2, sample_4), gt (gt_0, None, gt_2, None,)
     # test-cases[1] has sample (sample_2, sample_2, sample_3), gt (gt_1, gt_2, gt_4,)
     # test sample #3 across test suite, should have different ground_truths
-    sample, _ = inferences[test_case_id_0][2]
+    sample, _ = sorted(inferences[test_case_id_0], key=lambda x: x[0].locator)[2]
     assert len(sample.ground_truths) == 1
```
ly-kolena, line 91: Yesterday I've seen a couple test failures at this assertion
```
=========================== short test summary info ============================
FAILED tests/integration/detection/test_model.py::test__load_inferences__no_inferences - AssertionError: assert 0 == 1
 +  where 0 = len([])
 +    where [] = TestImage(locator='https://fake-locator/HKSVUW391V7V/detection/base/4.png', dataset='HKSVUW391V7V fake-data-set', metadata={'example': 'metadata', 'i': 4}, ground_truths=[]).ground_truths
============= 1 failed, 22 passed, 4 warnings in 673.75s (0:11:13) =============
```

Suspect non-deterministic ordering could be the culprit.

gordonhart: Should we also fix this upstream?

gordonhart (approved):

tests/integration/detection/test_model.py
ly-kolena: I have not looked in detail for this particular case. However, from others I've seen during migration, the order issue most often was from that id sequence does not necessarily match to the insertion order. Typical example is the test_sample/image. I think, even with the measures we put in place, there is no way enforce sequence id matches the insertion order.

PR #38, opened 2023-05-17
Logging improvements

gordonhart:
- Log `created` before `edited` when creating+populating test cases/test suites
- Only use `log.success` (green) on actions that modify stored state — not for loads
- Improve clarity of test run `reset=False` logs
- Consistently include version in test case/test suite logs across workflows

### DIFF
--- a/kolena/detection/_internal/test_suite.py
+++ b/kolena/detection/_internal/test_suite.py
@@ -116,7 +116,7 @@ class BaseTestSuite(ABC, Frozen, WithTelemetry):
         krequests.raise_for_status(res)
         data = from_dict(data_class=CoreAPI.TestSuite.EntityData, data=res.json())
         obj = cls._create_from_data(data)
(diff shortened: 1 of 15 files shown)


kolena/detection/_internal/test_suite.py
```diff
@@ -289,7 +289,7 @@ def edit(self, reset: bool = False) -> Iterator[Editor]:
             log.info("no op: nothing edited")
             return
 
-        log.info(f"edited test suite '{self.name}'")
+        log.info(f"edited test suite '{self.name}' (v{self.version})")
```
ly-kolena: `editing`?

ly-kolena (approved):

kolena/detection/_internal/test_suite.py
gordonhart: Good catch, updated.

yifanwu-kolena (approved):

PR #39, opened 2023-05-17
Add age estimation example integration

gordonhart:
Porting over the [LFW](http://vis-www.cs.umass.edu/lfw/) age estimation demo integration, including automated tests. [Example on app.kolena.io](https://app.kolena.io/sdk-integ/results?modelIds=2&testSuiteId=4&workflow=Age+Estimation&evaluators=N4IgLCBcoExQjAX0UA).

PR #40, opened 2023-05-18
Disregard GT ordering in builtin classification/detection TestImage equality

gordonhart:
Addresses the red tests surfaced in #39. The ordering of classification `labels` and detection `ground_truths` is not critical, so it's acceptable to treat two test images with the same contents in a different order as equivalent.

This is intended to be a short-term fix while we work to ensure that this order is preserved upstream.

### DIFF
--- a/kolena/detection/test_image.py
+++ b/kolena/detection/test_image.py
@@ -83,13 +83,14 @@ class TestImage(BaseTestImage):
 
     # TODO: remove implementation in favor of Frozen.__eq__ once ground_truth ordering is ensured upstream
     def __eq__(self, other: Any) -> bool:
-        def sort_ground_truths(ground_truths: List[GroundTruth]) -> List[GroundTruth]:
-            return sorted(ground_truths, key=lambda gt: json.dumps(gt._to_dict(), sort_keys=True))
-
         return isinstance(other, type(self)) and {
             **self.__dict__,
-            "ground_truths": sort_ground_truths(self.ground_truths),
-        } == {**other.__dict__, "ground_truths": sort_ground_truths(other.ground_truths)}
+            "ground_truths": self._sort_ground_truths(self.ground_truths),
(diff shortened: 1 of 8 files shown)


kolena/detection/test_image.py
```diff
@@ -80,6 +81,16 @@ def _from_record(cls, record: Any) -> "TestImage":
     def _to_record(cls, image: "TestImage") -> Tuple[str, Optional[str], List[Dict[str, Any]], Dict[str, Any]]:
         return (image.locator, image.dataset, [gt._to_dict() for gt in image.ground_truths], _to_dict(image.metadata))
 
+    # TODO: remove implementation in favor of Frozen.__eq__ once ground_truth ordering is ensured upstream
+    def __eq__(self, other: Any) -> bool:
+        def sort_ground_truths(ground_truths: List[GroundTruth]) -> List[GroundTruth]:
```
ly-kolena: Nit: this function does not look to depend on enclosure, can be moved to its own method instead of created per invocation.

kolena/detection/test_image.py
```diff
@@ -80,6 +81,16 @@ def _from_record(cls, record: Any) -> "TestImage":
...
+            return sorted(ground_truths, key=lambda gt: json.dumps(gt._to_dict(), sort_keys=True))
```
ly-kolena: Did a quick google and didn't find warning bells using json and floating point comparison, looks to be good. Still checking with you, any edge case you can think of?

ly-kolena (approved):

kolena/detection/test_image.py
gordonhart: Updated

kolena/detection/test_image.py
gordonhart: Good flag, we should be safe here under the assumption that `json.dumps(...)` has consistent behavior on a given system.

PR #42, opened 2023-05-18
Conditional integration testing

ly-kolena:

Conditionally run integration/example tests based on what files are changed. The logic is:

* for each workflow, files under the matching directory in `kolena` and `tests/integration` changed, the integration tests under the directory would run
* files change elsewhere would trigger all integration and example tests to run
* for example tests, it would also run if generic workflow needs re-testing or codes under `examples` changed.
* ci/unit-test are always run (for now, unless we feel the need to further trim down)

example tests can potentially be further refined, but leaving it out from this PR.

**NOTE**: the one uncertainty is the impact to code-coverage report, may need a couple iteration. I'm using `carryforward` to `backfill` un-run tests from existing data.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,236 +1,71 @@
 version: 2.1
 
+setup: true
+
 orbs:
-  codecov: codecov/codecov@3.2.4
+  continuation: circleci/continuation@0.3.1
+  path-filtering: circleci/path-filtering@0.1.1
 
 parameters:
-  python-version:
-    type: string
-    default: "3.9"
+  manual:
+    type: boolean
+    default: false
 
 jobs:
-  ci-base:
-    parameters:
-      python-version:
-        type: string
-        default: "3.9"
-    docker:
-      - image: cimg/python:<< parameters.python-version >>
+  setup:
+    executor: path-filtering/default
     resource_class: small
-    environment:
-      POETRY_CACHE_DIR: /home/circleci/project/.poetry
     steps:
       - checkout
-      - restore_cache:
-          key: &ci-base-cache ci-cache-<< parameters.python-version >>-{{ checksum "pyproject.toml" }}
-      - run: |
-          poetry config experimental.new-installer false
-          poetry config installer.max-workers 10
-          poetry install --no-ansi
-      - save_cache:
-          key: *ci-base-cache
-          paths:
-            - /home/circleci/project/.poetry/virtualenvs
-            - poetry.lock
-
-  unit-test:
-    parameters:
-      python-version:
-        type: string
-        default: "3.9"
-    docker:
-      - image: cimg/python:<< parameters.python-version >>
-    resource_class: small
-    environment:
-      POETRY_CACHE_DIR: /home/circleci/project/.poetry
-    steps:
-      - checkout
-      - restore_cache:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -1,238 +1,50 @@
...
-          poetry install --no-ansi
-      - run:
-          name: Run pre-commit checks
-          command: |
-            poetry run pre-commit run -a
+          name: Print parameters
+          command: cat /tmp/pipeline-parameters.json
       - run:
-          name: Run << parameters.subproject >> (Python << parameters.python-version >>) integration tests
+          name: Generate config
```
ly-kolena, line 38: This section is the black magic to replace integration test `pytest-group` matrix with actual values from base `continue_config.yml`.

This is the compromise from a few approaches I tried. Keeping it shell commands for now, instead of a separate shell or python executable.

munkyshi: I've definitely preferred python scripts in the the past -- bash is pretty unmaintainable, and it makes this dense and hard to understand for anyone trying to debug or edit things.

.circleci/continue_config.yml
```diff
@@ -0,0 +1,235 @@
...
+      POETRY_CACHE_DIR: /home/circleci/project/.poetry
+    parallelism: << parameters.parallelism >>
+    steps:
+      - when:
+          condition:
+            not: << parameters.enabled >>
+          steps:
+            - run:
+                name: skip job
+                command: circleci-agent step halt
```
munkyshi, line 122: Have we tested that the job is properly skipped when not scoped? If so, I'm willing to approve based on the results alone.

munkyshi (approved):

.circleci/continue_config.yml
ly-kolena: i have tested several combinations.

.circleci/config.yml
```diff
@@ -1,236 +1,66 @@
...
+      - path-filtering/set-parameters:
+          base-revision: ${BASE_REVISION}
+          mapping: |
+            kolena/classification/.* classification true
+            tests/integration/classification/.* classification true
+            kolena/detection/.* detection true
+            tests/integration/detection/.* detection true
+            kolena/fr/.* fr true
+            tests/integration/fr/.* fr true
+            kolena/workflow/.* generic true
```
gordonhart, line 27: Non-blocking, but I've been meaning to change the `generic` integration tests to `workflow` to reflect this external naming convention.

gordonhart: https://github.com/kolenaIO/kolena/pull/46

.circleci/continue_config.yml
```diff
@@ -0,0 +1,235 @@
...
+    default: false
+  fr:
+    type: boolean
+    default: false
+  generic:
+    type: boolean
+    default: false
+  misc:
+    type: boolean
+    default: false
```
yifanwu-kolena, line 21: where do we use this parameter?

ly-kolena: Path-filter will output the parameters into a json file. It is then used at step `Generate Config` to manipulate the continued yaml.

yifanwu-kolena: 🤔 i didn't see it in the mapping, maybe i misunderstand something

ly-kolena: Understood your question now, will follow up.

PR #43, opened 2023-05-18
Only smoke test example integrations in CI

gordonhart:
Pare down the datasets used by the `text_summarization` and `age_estimation` automated tests to just a handful of samples. This are really just smoke tests, and have been renamed as such.

### DIFF
--- a/examples/text_summarization/tests/test_text_summarization.py
+++ b/examples/text_summarization/tests/test_text_summarization.py
@@ -29,7 +29,7 @@ def with_init() -> Iterator[None]:
 
 
 def test__seed_test_suite() -> None:
-    args = Namespace(dataset_csv="s3://kolena-public-datasets/CNN-DailyMail/metadata/CNN_DailyMail_metadata_5.csv")
+    args = Namespace(dataset_csv="s3://kolena-public-datasets/CNN-DailyMail/metadata/metadata.tiny1.csv")
     seed_test_suite_main(args)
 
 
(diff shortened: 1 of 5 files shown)


examples/text_summarization/tests/test_text_summarization.py
```diff
@@ -28,12 +28,12 @@ def with_init() -> Iterator[None]:
...
-    args = Namespace(dataset_csv="s3://kolena-public-datasets/CNN-DailyMail/metadata/CNN_DailyMail_metadata_100.csv")
+def test__seed_test_suite__smoke() -> None:
+    args = Namespace(dataset_csv="s3://kolena-public-datasets/CNN-DailyMail/metadata/metadata.tiny1.csv")
     seed_test_suite_main(args)
 
 
 @pytest.mark.depends(on=["test__seed_test_suite"])
-def test__seed_test_run() -> None:
+def test__seed_test_run__smoke() -> None:
     args = Namespace(model="ada", test_suite="CNN-DailyMail :: text length", local_csv=None)
```
ly-kolena, line 28: Do we need to add random prefix to these tests?

ly-kolena (approved):

examples/text_summarization/tests/test_text_summarization.py
gordonhart: These tests are idempotent due to `reset=True` (which is why they haven't been failing). This does mean that we're running from a different starting state depending on how many times these jobs have run, but I think that's preferable over drilling a `prefix` concept into these example scripts for the purposes of testing. I'm open to being convinced otherwise though!

PR #45, opened 2023-05-19
add circleCI job for different resource class

marko-polo-cheno:
Might be a simpler alternative to running tests on different resource sizes without excludes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -163,7 +163,7 @@ jobs:
             - store_test_results:
                 path: test-results
 
-  example-test:
+  example-test-large: &example-test-large
     parameters:
       python-version:
         type: string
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -222,20 +225,15 @@ workflows:
               pytest-group: [ classification, misc ]
           requires:
             - ci-base-<< matrix.python-version >>
-      - example-test:
+      - example-test-large:
```
ly-kolena: Nit: I think this can be simplified into

```
workflows:
  ci:
    jobs:
      ...
      - example-test:
             name: example-test-large
             <unchanged>
       - example-test:
              name: example-test-small
              <unchanged>
```

with
```
jobs
  example-test:
     <unchanged-from-example-test-large>
```

ly-kolena (approved):

PR #48, opened 2023-05-24
Migrate to new api_base envvar

ly-kolena:
Introduce new envvar `KOLENA_API_URL` to configure api url

### DIFF
--- a/kolena/_utils/state.py
+++ b/kolena/_utils/state.py
@@ -34,15 +34,16 @@ from kolena.errors import UnauthenticatedError
 from kolena.errors import UninitializedError
 
 API_VERSION = "v1"
-API_BASE_URL = "https://api.kolena.io"
-API_BASE_URL_ENV_VAR = "KOLENA_CLIENT_BASE_URL"
+API_URL = "https://api.kolena.io"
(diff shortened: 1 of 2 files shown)


kolena/_utils/state.py
```diff
@@ -34,15 +34,16 @@
 from kolena.errors import UninitializedError
 
 API_VERSION = "v1"
-API_BASE_URL = "https://api.kolena.io"
-API_BASE_URL_ENV_VAR = "KOLENA_CLIENT_BASE_URL"
+API_URL = "https://api.kolena.io"
+CLIENT_BASE_URL_ENV_VAR = "KOLENA_CLIENT_BASE_URL"
```
gordonhart: Is there a particular reason to preserve this compatibility? This environment variable has thus far not been external-facing.

ly-kolena: Internal CI/CD or developers may be impacted.

gordonhart: We'd have to migrate to the new environment variable when this release is picked up, but that will be an intentional change anyway as existing usages of this library pin its version. Not a big deal either way but I'd rather rip this band-aid off now than preserve the old environment variable indefinitely

gordonhart (approved):

PR #49, opened 2023-05-25
Support multiple distributions per test case in `kolena.workflow.Histogram`

gordonhart:
This PR extends the `kolena.workflow.Histogram` API object to support specifying multiple frequency distributions corresponding to e.g. different classes within a test case.


gordonhart:
Screen Shot 2023-05-25 at 10 34 31 AM

marko-polo-cheno (approved):
Might need to update `visualizing a histogram per test case` because now it can handle more than one histogram per test case, otherwise LGTM.

y27choi (approved):

ly-kolena (approved):

PR #50, opened 2023-05-25
Add confusion matrix to plots

marko-polo-cheno:
Fixes KOL-2297

### DIFF
--- a/kolena/workflow/metrics/_plots.py
+++ b/kolena/workflow/metrics/_plots.py
@@ -1,0 +1,63 @@
+# Copyright 2021-2023 Kolena Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from collections import defaultdict
+from typing import Dict
+from typing import List
+from typing import Optional
+from typing import Set
+from typing import Union
+
+from kolena.workflow.evaluator import ConfusionMatrix
+from kolena.workflow.evaluator import Plot
+from kolena.workflow.metrics._geometry import InferenceMatches
+from kolena.workflow.metrics._geometry import MulticlassInferenceMatches
+
+
+def compute_test_case_confusion_matrix(
+    all_matches: List[Union[MulticlassInferenceMatches, InferenceMatches]],
+    test_case_name: str,
+    plot_title: str,
+) -> Optional[Plot]:
+    labels: Set[str] = set()
+
+    confusion_matrix: Dict[str, Dict[str, int]] = defaultdict(lambda: defaultdict(int))
+    for match in all_matches:
+        for gt, _ in match.matched:
+            actual_label = gt.label
+            confusion_matrix[actual_label][actual_label] += 1
+            labels.add(actual_label)
+
+        for gt, inf in match.unmatched_gt:
+            actual_label = gt.label
+            labels.add(actual_label)
+            if inf is not None:
+                predicted_label = inf.label
+                confusion_matrix[actual_label][predicted_label] += 1
+                labels.add(predicted_label)
+
+        labels.update(inf.label for inf in match.unmatched_inf)
+
+    if len(labels) < 2:
+        print(f"skipping confusion matrix for {test_case_name}: single label test case")
+        return None
+
+    if len(labels) > 10:
+        print(f"skipping confusion matrix for {test_case_name}: too many labels")
+        return None
+
+    ordered_labels = sorted(labels)
+    matrix = []
+    for actual_label in ordered_labels:
+        matrix.append([confusion_matrix[actual_label][predicted_label] for predicted_label in ordered_labels])
+    return ConfusionMatrix(title=plot_title, labels=ordered_labels, matrix=matrix)
--- a/kolena/workflow/metrics/_plots.py
+++ b/kolena/workflow/metrics/_plots.py
@@ -22,10 +22,17 @@ from kolena.workflow.evaluator import ConfusionMatrix
 from kolena.workflow.metrics._geometry import MulticlassInferenceMatches
 
 
-def compute_test_case_confusion_matrix(
+def compute_object_detection_test_case_confusion_matrix(
     all_matches: List[MulticlassInferenceMatches],
     plot_title: str = "Confusion Matrix",
 ) -> Optional[ConfusionMatrix]:
+    """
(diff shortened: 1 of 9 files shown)


kolena/workflow/metrics/_plots.py
```diff
@@ -0,0 +1,63 @@
...
+from typing import Union
+
+from kolena.workflow.evaluator import ConfusionMatrix
+from kolena.workflow.evaluator import Plot
+from kolena.workflow.metrics._geometry import InferenceMatches
+from kolena.workflow.metrics._geometry import MulticlassInferenceMatches
+
+
+def compute_test_case_confusion_matrix(
+    all_matches: List[Union[MulticlassInferenceMatches, InferenceMatches]],
```
gordonhart: It looks like we just want `MulticlassInferenceMatches` here, as we're referencing `gt.label` and `inf.label` which don't exist on the GT/Inf types in `InferenceMatches`.

kolena/workflow/metrics/_plots.py
```diff
@@ -0,0 +1,62 @@
...
+        labels.update(inf.label for inf in match.unmatched_inf)
+
+    if len(labels) < 2:
+        print(f"skipping confusion matrix for {test_case_name}: single label test case")
+        return None
+
+    # TODO: Remove limits when labels of plots can be changed by the user
+    if len(labels) > 10:
+        print(f"skipping confusion matrix for {test_case_name}: too many labels")
+        return None
```
gordonhart: Looks like `test_case_name` is only used for these two debug messages (aside: should use [`log.info`](https://github.com/kolenaIO/kolena/blob/trunk/kolena/_utils/log.py#L57)). In general, function signatures should be kept as small as possible — suggest refactoring to remove `test_case_name` from this one.

kolena/workflow/metrics/_plots.py
```diff
@@ -0,0 +1,62 @@
...
+
+from kolena.workflow.evaluator import ConfusionMatrix
+from kolena.workflow.evaluator import Plot
+from kolena.workflow.metrics._geometry import MulticlassInferenceMatches
+
+
+def compute_test_case_confusion_matrix(
+    all_matches: List[MulticlassInferenceMatches],
+    test_case_name: str,
+    plot_title: str,
```
gordonhart: Is there a sensible default value here? `Confusion Matrix`?

kolena/workflow/metrics/_plots.py
```diff
@@ -0,0 +1,61 @@
...
+# limitations under the License.
+from collections import defaultdict
+from typing import Dict
+from typing import List
+from typing import Optional
+from typing import Set
+
+from kolena._utils import log
+from kolena.workflow.evaluator import ConfusionMatrix
+from kolena.workflow.metrics._geometry import MulticlassInferenceMatches
```
y27choi, line 22: Should `MulticlassInferenceMatches` be available under `kolena.workflow.metrics`?  It's currently missing from the API doc.

kolena/workflow/metrics/_plots.py
```diff
@@ -0,0 +1,61 @@
...
+                confusion_matrix[actual_label][predicted_label] += 1
+                labels.add(predicted_label)
+
+        labels.update(inf.label for inf in match.unmatched_inf)
+
+    if len(labels) < 2:
+        log.info(f"skipping confusion matrix for a single label: {labels}")
+        return None
+
+    # TODO: Remove limits when labels of plots can be changed by the user
```
y27choi: Now that we support a larger # of labels on a confusion matrix as of release 0.69.0, do we still want to skip plotting when there are too many labels?

kolena/workflow/metrics/_plots.py
```diff
@@ -0,0 +1,61 @@
...
+
+        for gt, inf in match.unmatched_gt:
+            actual_label = gt.label
+            labels.add(actual_label)
+            if inf is not None:
+                predicted_label = inf.label
+                confusion_matrix[actual_label][predicted_label] += 1
+                labels.add(predicted_label)
+
+        labels.update(inf.label for inf in match.unmatched_inf)
```
y27choi, line 53: Do we need to add `unmatched_inf`'s label to `labels`?

kolena/workflow/metrics/_plots.py
```diff
@@ -0,0 +1,61 @@
...
+from typing import List
+from typing import Optional
+from typing import Set
+
+from kolena._utils import log
+from kolena.workflow.evaluator import ConfusionMatrix
+from kolena.workflow.metrics._geometry import MulticlassInferenceMatches
+
+
+def compute_test_case_confusion_matrix(
```
y27choi: Should we add a API-doc friendly header here?

y27choi: This plot function seems to be implemented specifically for OD workflow.  Do we need one for Classification workflow?  If we are not intending on supporting Classification ConfusionMatrix in this PR, we should probably make it clear that this is for multi-class OD.

kolena/workflow/metrics/_plots.py
marko-polo-cheno: nice catch

kolena/workflow/metrics/_plots.py
marko-polo-cheno: I think it is better for conf matrices to be explicit, otherwise it may seem like there is missing information

y27choi: > conf matrices to be explicit

Can you elaborate more on this?

One thing I'm concerned about is that if there happens to be labels that were not part of any of the `matches` but are in `unmatched_inf`, [line 60](https://github.com/kolenaIO/kolena/pull/50/files/dd019b7598e23a70de0f5dea431d5fdadb5fe97e#diff-41991ae41fb36a2427881ebdff528b82d89b8daa07a6b1712d7945dcf6cd1144R60) will fail.

marko-polo-cheno: I think the confusion matrix should be complete, containing all non-ignored labels, and should be a square (n by n). If a label was only in the `unmatched_inf`s, a user would gain more information from seeing a 0-column/row than having it absent from view. It might be misleading if `conf_mat(labels=[a,b,c])` and `conf_mat(labels=[a,b], ignore=[c])` produced the exact same matrix. (`ignore` is implicit from stratification / test case design)

`defaultdict` should handle line 60, but definitely good to add a test for that 👍

kolena/workflow/metrics/_plots.py
```diff
@@ -0,0 +1,63 @@
...
+from typing import List
+from typing import Optional
+from typing import Set
+
+from kolena._utils import log
+from kolena.workflow.evaluator import ConfusionMatrix
+from kolena.workflow.metrics._geometry import MulticlassInferenceMatches
+
+
+def compute_object_detection_test_case_confusion_matrix(
```
munkyshi, line 25: Same comment as with the other pull request on naming to make things less verbose -- does the user need to think that that this is for `test_case` specifically? Any reason not to just have this as `compute_object_detection_confusion_matrix`?

On a side note, is `object_detection` necessary / should we split out OD metrics code into its own package? It feels as if it's implicitly assumed for other tools in `metrics`, but explicitly mentioned here.

PR #52, opened 2023-05-30
Support nested aggregate metrics in MetricsTestCase

gordonhart:
Add the ability to nest `List[MetricsTestCase]` inside a `MetricsTestCase` object, supporting e.g. per-class metrics reported for multiple classes within a test case.

Fixes KOL-2408

### DIFF
--- a/kolena/workflow/evaluator.py
+++ b/kolena/workflow/evaluator.py
@@ -71,7 +71,7 @@ class MetricsTestCase(DataObject, metaclass=ABCMeta):
     .. code-block:: python
 
         @dataclass(frozen=True)
-        class PerClassMetrics(DataObject):
+        class PerClassMetrics(MetricsTestCase):
             Precision: float
             Recall: float
             F1: float
@@ -430,8 +430,8 @@ def _validate_metrics_test_sample_type(metrics_test_sample_type: Type[MetricsTes
 def _validate_metrics_test_case_type(metrics_test_case_type: Type[DataObject]) -> None:
     validate_data_object_type(
         metrics_test_case_type,
-        supported_field_types=[_SCALAR_TYPES, List[DataObject]],
-        allow_lists=False,
+        supported_field_types=_SCALAR_TYPES,
+        supported_list_types=[MetricsTestCase],
     )
 
 
--- a/kolena/workflow/evaluator.py
+++ b/kolena/workflow/evaluator.py
@@ -26,16 +26,18 @@ from pydantic.dataclasses import dataclass
 from pydantic.typing import Literal
 
 from kolena._api.v1.generic import TestRun as API
+from kolena._utils.datatypes import get_args
+from kolena._utils.datatypes import get_origin
 from kolena._utils.validators import ValidatorConfig
 from kolena.workflow import GroundTruth
 from kolena.workflow import Inference
 from kolena.workflow import TestCase
 from kolena.workflow import TestSample
 from kolena.workflow import TestSuite
-from kolena.workflow._datatypes import _SCALAR_TYPES
 from kolena.workflow._datatypes import DataObject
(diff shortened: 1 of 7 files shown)


gordonhart:
Screen Shot 2023-05-31 at 12 25 48 PM

kolena/workflow/evaluator.py
```diff
@@ -63,6 +66,26 @@ class MetricsTestCase(DataObject, metaclass=ABCMeta):
...
+    contains multiple classes. Example usage:
+
+    .. code-block:: python
+
+        @dataclass(frozen=True)
+        class PerClassMetrics(MetricsTestCase):
+            Precision: float
+            Recall: float
+            F1: float
+            AP: float
```
ly-kolena, line 81: Nit: class name?

kolena/workflow/evaluator.py
```diff
@@ -407,7 +430,21 @@ def _validate_metrics_test_sample_type(metrics_test_sample_type: Type[MetricsTes
...
+
+    # validate that there is only one level of nesting
+    for field_name, field_type in get_data_object_field_types(metrics_test_case_type).items():
+        origin = get_origin(field_type)
+        if origin is not list:  # only need to check lists, as MetricsTestCase is only allowed in lists
+            continue
+        # expand e.g. List[Union[MetricsA, MetricsB]] into [MetricsA, MetricsB]
+        list_args = [t for arg_type in get_args(field_type) for t in get_args(arg_type) or [arg_type]]
+        for arg_type in list_args:
+            if issubclass(arg_type, MetricsTestCase):
```
ly-kolena: Nit: use `safe_issubclass` to prevent unexpected usage, as `issubclass` can throw exception?

ly-kolena (approved):

kolena/workflow/evaluator.py
gordonhart: Thanks for flagging — adding `Class`.

tests/unit/workflow/test_evaluator.py
```diff
@@ -150,12 +151,89 @@ class DictTester(base):
...
+        a: int
+
+    @dataclasses.dataclass(frozen=True)
+    class NestedTester(MetricsTestCase):
+        a: int
+        b: float
+        c: List[Nested]
+        d: Optional[List[Nested]]
+        e: Union[List[Nested], List[Nested2]]
+        f: List[Union[Nested, Nested2]]
```
munkyshi, line 188: What's the desired behavior for `List[Optional[Nested]]`? My assumption is that we should forbid it -- can we add a check for that?

Questions for how the webapp will handle this:
- How does the webapp know to treat the object as a `List[Nested]`? As is, my initial approach would be to check the value type, ensure that it isn't an asset or annotation, then assume that it's a nested metric if it's an Object. That's a bit messy, am I missing anything?
- How will the webapp handle `List[Union[Nested, Nested2]]`? Should it read all of the entries, and then use that to determine the columns to display (seems like there's no way to avoid loading all of the data, which would prevent chunking it out, but that's probably okay)?

gordonhart: > How does the webapp know to treat the object as a `List[Nested]`?

No real magic or special handling required here. Test case metrics values are either scalars or lists of scalar-only objects. No need to check for things like assets or annotations.

> How will the webapp handle `List[Union[Nested, Nested2]]`?

Yeah, as suggested, union the columns. We can do this on the frontend or on the backend if chunking is required.

gordonhart: > What's the desired behavior for `List[Optional[Nested]]`? My assumption is that we should forbid it

👍 Added a test and an extra validation condition to disallow this.

munkyshi (approved):

PR #54, opened 2023-05-31
Add object detection test case plots

marko-polo-cheno:
PR + F1 plots


kolena/workflow/metrics/_plots.py
```diff
@@ -0,0 +1,119 @@
...
+def _compute_sklearn_arrays(
+    all_matches: List[Union[MulticlassInferenceMatches, InferenceMatches]],
+) -> Tuple[List[int], List[int], Dict[str, Tuple[List[int], List[int]]]]:
+    arrays_by_label: defaultdict[str, Tuple[List[int], List[int]]] = defaultdict(lambda: ([], []))
+    y_true: List[int] = []
+    y_score: List[int] = []
+    for image_bbox_matches in all_matches:
+        for _, bbox_inf in image_bbox_matches.matched:  # TN (if above threshold)
+            arrays_by_label[bbox_inf.label][0].append(1)
+            arrays_by_label[bbox_inf.label][1].append(bbox_inf.score)
```
munkyshi: nit: The structure and naming for `arrays_by_label` makes it hard to read what's going on (What is `arrays`, what is the indexing in the tuple?).

Thoughts on:
```
y_true_by_label[bbox_inf.label].append(1)
y_score_by_label[bbox_inf.label].append(bbox_inf.score)
```

kolena/workflow/metrics/_plots.py
```diff
@@ -0,0 +1,119 @@
...
+        )
+        precisions.append(precision)
+        recalls.append(recall)
+        f1s.append(f1)
+    return Curve(x=thresholds, y=f1s, label=label), Curve(x=recalls, y=precisions, label=label)
+
+
+def compute_object_detection_test_case_plots(
+    all_matches: List[MulticlassInferenceMatches],
+    test_case_name: str,
```
munkyshi: Is this test case specific, or just used for naming of the curves? It seems like we'd like these metrics to be as general as possible, and the more we can avoid prescribing the use case, the better (would `plot_name` or `curve_name` work just as well?).

kolena/workflow/metrics/__init__.py
```diff
@@ -15,8 +15,10 @@
 from ._geometry import iou
 from ._geometry import match_inferences
 from ._geometry import match_inferences_multiclass
+from ._plots import compute_object_detection_test_case_plots
 
 __all__ = [
+    "compute_object_detection_test_case_plots",
```
munkyshi, line 21: Naming is a bit of a mouthful here, and doesn't tell a user what plots are actually being computed -- anything we can do to improve this?

`compute_pr_f1_plots`?

kolena/workflow/metrics/_plots.py
```diff
@@ -0,0 +1,119 @@
...
+            y_score.append(bbox_inf.score)
+    return y_true, y_score, dict(arrays_by_label)
+
+
+def _compute_threshold_curves(
+    y_true: np.ndarray,
+    y_score: np.ndarray,
+    label: str,
+) -> Tuple[Curve, Curve]:
+    thresholds = list(np.linspace(min(abs(y_score)), max(y_score), min(401, len(y_score))))[:-1]
```
munkyshi: Curious about this piece:
```
min(401, len(y_score))
```
`len(y_score)` is strange to linspace with -- if you're going to have a point for every y_score, why not just use the threshold values (instead of linspacing)?

On the flip side, one consideration that we had for builtin OD was that linspacing gave us more points when values were sparse, but less detail than we would have liked when values were dense. Not a blocker, but worth pointing out that you could have 100 points between (0.9525, 0.9550) and only 10 points between (0, 0.5) -- linspacing smudges that out.

Last thing to point out -- there's a chance that you're missing the f1 optimal point if you linspace. Do we want to ensure that it gets surfaced?

munkyshi (commented):
Haven't looked too deeply at the logic -- did we get a chance to compare this to outputs from existing OD plotting?

General question: Is the immediate use case for these metrics the generic OD implementation? Are we still planning on surfacing things like `F1-Optimal`?

nit: Can you add your name or initials to the branch you create for these? It helps with pruning / deleting old code (something like `mc/add-object-detection-test-case-plots` lets us know who's working on the branch / who to talk to if we want to clean up old branches)

kolena/workflow/metrics/_plots.py
marko-polo-cheno: I made more progress on the entire workflow for a better understanding of the [full picture](https://github.com/kolenaIO/kolena/pull/56#discussion_r1214373355). Here, plot utils do not have to be at the test case level, but for the OD workflow, they will be per test case. I think a good solution to address missing the optimal threshold (for sparse cases) is doubling the size of the linspace.

marko-polo-cheno: @y27choi is doubling the size of the linspace from `min(401, len(y_score))` to `min(401, 2 * len(y_score))` a good idea?

y27choi: I'm not sure how doubling the size of the linspace can ensure including the f1 optimal threshold.  Can you explain?

In general, I agree with @munkyshi's comment.  I think we can definitely improve the way we plot PR / F1 curves, but I don't think it's a blocker.  Having said that, let's check popular implementations to see if there is a standard way of extracting the right amount of threshold data points.

And instead of doubling, what do you think of manually adding the f1 optimal threshold if it's not already included in `thresholds`?

kolena/workflow/test_sample.py
munkyshi, line 1: Is this diff intended? It looks like it undoes some of the changes that @gordonhart recently put in.

pyproject.toml
```diff
@@ -83,3 +83,6 @@ norecursedirs = "kolena"
 python_classes = []
 # only collect functions starting with 'test__'
 python_functions = ["test__*"]
+markers = [
+    "metrics: tests that extra dependencies for metrics",
```
munkyshi, line 87: This sentence feels incomplete.

tests/unit/workflow/test_evaluator.py
munkyshi, line 1: Is this diff intended? It doesn't look like these tests are reintroduced anywhere.

kolena/_experimental/object_detection/utils.py
```diff
@@ -0,0 +1,125 @@
...
+
+def _compute_sklearn_arrays(
+    all_matches: List[MulticlassInferenceMatches],
+) -> Tuple[List[int], List[int], Dict[str, List[int]], Dict[str, List[int]]]:
+    y_true_by_label: defaultdict[str, List[int]] = defaultdict(lambda: [])
+    y_score_by_label: defaultdict[str, List[int]] = defaultdict(lambda: [])
+    y_true: List[int] = []
+    y_score: List[int] = []
+    for image_bbox_matches in all_matches:
+        for _, bbox_inf in image_bbox_matches.matched:  # TN (if above threshold)
```
munkyshi, line 41: nit: should comment be `TP`?

kolena/workflow/_validators.py
munkyshi, line 1: Also feels like this is an unintended diff?

kolena/workflow/evaluator.py
munkyshi, line 1: Unintended diff?

kolena/workflow/metrics/__init__.py
munkyshi: Did this get resolved?

munkyshi:
Changes look good, just a few small comments, and it seems like an unintended diff got added.

kolena/_experimental/object_detection/utils.py
```diff
@@ -0,0 +1,125 @@
...
+
+import numpy as np
+
+from kolena._extras.metrics.sklearn import sklearn_metrics
+from kolena.workflow.evaluator import Curve
+from kolena.workflow.evaluator import CurvePlot
+from kolena.workflow.metrics import MulticlassInferenceMatches
+
+
+def threshold_key(label: str) -> str:
```
y27choi, line 28: Is this used anywhere?

kolena/_experimental/object_detection/utils.py
```diff
@@ -0,0 +1,125 @@
...
+    sanitized = re.sub(r"\W+", "_", label)
+    return f"threshold_{sanitized}"
+
+
+def _compute_sklearn_arrays(
+    all_matches: List[MulticlassInferenceMatches],
+) -> Tuple[List[int], List[int], Dict[str, List[int]], Dict[str, List[int]]]:
```
y27choi, line 35: Don't scikit-learn arrays need to be `np.array` type?  Is there reason why we are using `List` instead of `np.array` here?

kolena/_experimental/object_detection/utils.py
```diff
@@ -0,0 +1,125 @@
...
+    y_true_by_label: defaultdict[str, List[int]] = defaultdict(lambda: [])
+    y_score_by_label: defaultdict[str, List[int]] = defaultdict(lambda: [])
```
y27choi, line 37: shouldn't `y_score` be float?

kolena/_experimental/object_detection/utils.py
```diff
@@ -0,0 +1,125 @@
...
+    y_true: List[int] = []
+    y_score: List[int] = []
```
y27choi, line 39: shouldn't `y_score` be float?

kolena/_experimental/object_detection/utils.py
y27choi, line 1: Are these utility functions only for multi-class OD?  If so, what's the plan for single class OD?

kolena/_experimental/object_detection/utils.py
```diff
@@ -0,0 +1,125 @@
...
+    y_true: np.ndarray,
+    y_score: np.ndarray,
+    label: str,
+) -> Tuple[Curve, Curve]:
+    thresholds = list(np.linspace(min(abs(y_score)), max(y_score), min(401, len(y_score))))[:-1]
+    precisions: List[float] = []
+    recalls: List[float] = []
+    f1s: List[float] = []
+    for threshold in thresholds:
+        y_pred = [1 if score > threshold else 0 for score in y_score]
```
y27choi, line 69: Should it be counted as a positive prediction if `score > threshold` or `score >= threshold`?

kolena/_experimental/object_detection/utils.py
```diff
@@ -0,0 +1,125 @@
...
+    return Curve(x=thresholds, y=f1s, label=label), Curve(x=recalls, y=precisions, label=label)
+
+
+def compute_pr_f1_plots(
+    all_matches: List[MulticlassInferenceMatches],
+    curve_label: str = "baseline",
+) -> List[CurvePlot]:
+    """
+    Creates a PR (precision and recall) curve and F1-threshold (confidence threshold) curve for the multiclass object
+    detection workflow. For `n` labels, each plot has `n+1` curves. One for the test case, and one per label.
```
y27choi, line 88: Should update this now that we don't include per label plots.

kolena/_experimental/object_detection/utils.py
```diff
@@ -0,0 +1,125 @@
...
+
+    :param all_matches: a list of multiclass or single class matching results.
```
y27choi, line 90: Does it include `InferenceMatches`?

kolena/_experimental/object_detection/utils.py
```diff
@@ -0,0 +1,125 @@
...
+    :param curve_label: the label of the main curve.
+    :return: Two :class:`CurvePlot`s for the PR curves and F1-threshold curves for the test case and each label.
```
y27choi, line 92: Update here as well.

tests/unit/_experimental/test_plots.py
```diff
@@ -0,0 +1,1229 @@
...
+    for c1, c2 in zip(c1.curves, c2.curves):
+        assert_curve_equal(c1, c2)
+
+
+@pytest.mark.metrics
+@pytest.mark.parametrize(
+    "test_name, matchings, f1_curve, pr_curve",
+    [
+        (
+            "zeros with unmatched gt and unmatched inf",
```
y27choi, line 57: This is an interesting case.  What happens if there are no valid matches?  Should we still return a plot with a single point at the origin?  @marko-polo-cheno do you know what it would look like on the webapp?

tests/unit/_experimental/test_plots.py
```diff
@@ -0,0 +1,1229 @@
...
+                ],
+                x_config=None,
+                y_config=None,
+            ),
+            CurvePlot(
+                title="Precision vs. Recall",
+                x_label="Recall",
+                y_label="Precision",
+                curves=[
+                    Curve(x=[0.0, 0.0], y=[0.0, 0.0], label="zeros, but one match for label a"),
```
y27choi, line 158: Hmm shouldn't precision be `1.0` at threshold of `0`?  I guess this will depend on whether to include or exclude the matched instance right @ the threshold.

tests/unit/_experimental/test_plots.py
```diff
@@ -0,0 +1,1229 @@
...
+                    ],
+                    unmatched_inf=[],
+                ),
+            ],
+            CurvePlot(
+                title="F1-Score vs. Confidence Threshold",
+                x_label="Confidence Threshold",
+                y_label="F1-Score",
+                curves=[
+                    Curve(x=[1.0], y=[0.0], label="zeros, but b is confused with a"),
```
y27choi, line 216: I'm curious why `x = 1.0` here but all the previous cases have `x = 0.0`.

tests/unit/_experimental/test_plots.py
```diff
@@ -0,0 +1,1229 @@
...
+                    ],
+                    unmatched_inf=[],
+                ),
+            ],
+            CurvePlot(
+                title="F1-Score vs. Confidence Threshold",
+                x_label="Confidence Threshold",
+                y_label="F1-Score",
+                curves=[
+                    Curve(x=[0.0, 0.0, 0.0], y=[0.0, 0.0, 0.0], label="no confusion, one TP per label"),
```
y27choi, line 285: nit: probably doesn't really matter, but I think `np.linspace(0, 0, 4)[:-1]` results in duplicate thresholds of `0` — this is why we see `(4-1)` `(0, 0)` points.  Should we drop dup threshold?

tests/unit/_experimental/test_plots.py
```diff
@@ -0,0 +1,1229 @@
...
+                            ScoredLabeledBoundingBox((10.0, 10.0), (20.0, 20.0), "cow", 0.9),
+                        ),
+                    ],
+                    unmatched_gt=[
+                        (
+                            LabeledPolygon(
+                                points=[(1, 1), (1, 2), (2, 2), (2, 1)],
+                                label="cow",
+                            ),
+                            ScoredLabeledPolygon(
```
y27choi, line 601: Shouldn't this be also added to `unmatched_inf`?  
Currently this pair is only counted as 1 FN, but shouldn't this be 1 FN and 1 FP?

y27choi (commented):
Left some comments.  I will also double check the numbers in the unit tests.

y27choi (commented):
Left some comments.  I will also double check the numbers in the unit tests.

kolena/_experimental/object_detection/utils.py
gordonhart: To add to this, if using NumPy arrays the type hint should be `np.ndarray`

tests/unit/_experimental/test_plots.py
y27choi: It seems like sklearn uses only the unique thresholds.  [sklearn link](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.precision_recall_curve.html)
image

tests/unit/_experimental/test_plots.py
```diff
@@ -0,0 +1,1229 @@
...
+                    ],
+                ),
+            ],
+            CurvePlot(
+                title="F1-Score vs. Confidence Threshold",
+                x_label="Confidence Threshold",
+                y_label="F1-Score",
+                curves=[
+                    Curve(
+                        x=[
```
y27choi, line 1020: The current implementation produces `len(y_score)` (when there are < 401), but here `y_score` includes TP, FP, FN.  Does this make sense?  FNs don't have any valid scores (i.e. we are using `-1` here) so shouldn't there be only # of TP + FP thresholds?

tests/unit/_experimental/test_plots.py
```diff
@@ -0,0 +1,1229 @@
...
+                x_config=None,
+                y_config=None,
+            ),
+            CurvePlot(
+                title="Precision vs. Recall",
+                x_label="Recall",
+                y_label="Precision",
+                curves=[
+                    Curve(
+                        x=[
```
y27choi, line 1120: I tried running sklearn's `precision_recall_curve` and compared the results. 
It seems like precision and recall are incorrect @ threshold of `0.01`.  Could you please take a look at this @marko-polo-cheno ?

[pr_curve.py.zip](https://github.com/kolenaIO/kolena/files/11682132/pr_curve.py.zip)

kolena/_experimental/object_detection/utils.py
marko-polo-cheno: Will add `InferenceMatches` for pr and f1 plots

tests/unit/_experimental/test_plots.py
marko-polo-cheno: In the new PR, I think I've addressed all the comments. Most are due to a logic change. This case now matches perfectly, with the exception of the threshold @ -1, which I omit.

The two naive but massive issues are non-distinct thresholds with the `>=` vs `>` choice.

PR #55, opened 2023-06-01
Move plot definitions to `kolena.workflow.plot`

gordonhart:
Previously these lived in `kolena.workflow.evaluator`, which was getting quite crowded. This new location should improve browsability of API docs. The definitions are also accessible through `kolena.workflow.evaluator` for backcompat, as verified by unit tests.


examples/text_summarization/text_summarization/evaluator.py
```diff
@@ -26,11 +26,11 @@
 from text_summarization.workflow import TestSampleMetric
 from text_summarization.workflow import TestSuiteMetric
 
+from kolena.workflow import AxisConfig
+from kolena.workflow import Curve
+from kolena.workflow import CurvePlot
```
ly-kolena, line 31: `kolena/_experimental/object_detection/utils.py` looks to have left-over to update too.

ly-kolena (approved):

examples/text_summarization/text_summarization/evaluator.py
gordonhart: Thanks for catching. Updated the new experimental OD imports in source and tests.

PR #56, opened 2023-06-02
[WIP] Create experimental object detection workflow

marko-polo-cheno:
Moving all OD plot changes into this PR (was easier to code the evaluator having all parts available together)


kolena/_experimental/object_detection/utils.py
```diff
@@ -0,0 +1,204 @@
...
+            y_score.append(bbox_inf.score)
+    return y_true, y_score, dict(y_true_by_label), dict(y_score_by_label)
+
+
+def _compute_threshold_curves(
+    y_true: np.ndarray,
+    y_score: np.ndarray,
+    label: str,
+) -> Tuple[Curve, Curve]:
+    thresholds = list(np.linspace(min(abs(y_score)), max(y_score), min(401, len(y_score))))[:-1]
```
marko-polo-cheno, line 70: Here, we cap the maximum number of points a plot can have to be 401. In the worst case, we miss the optimal threshold by about 1/802 (0.001). When confidences are dense, we do not have to extend the threshold ranges since extra potential PR curve points will be dups, and extra F1 plot points are outside of the possible area where the max F1 might exist.
Should we increase 401 to something higher? 1001?

PR #57, opened 2023-06-02
Added sklearn as an extra and optional dependancy

marko-polo-cheno:
Adds optional dependancy (mainly for sklearn)


kolena/_utils/scikit-learn.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2021-2023 Kolena Inc.
```
ly-kolena: What is the purpose of this file?

README.md
```diff
@@ -23,6 +23,21 @@ model behaviors and take the mystery out of model development. Kolena helps you:
 This `kolena` package contains the Python client library for programmatic interaction with the Kolena ML testing
 platform.
 
+## Setup
```
ly-kolena, line 26: This looks like instruction for development. Should we include the instruction for users wanting to install kolena through `pip`?

kolena/_utils/scikit-learn.py
marko-polo-cheno: To check if sklearn is installed or not in advance, [similar to this](https://github.com/kolenaIO/kolena/blob/trunk/kolena/_utils/datatypes.py#L33). I might have misunderstood what @gordonhart had in mind.

ly-kolena: I think this may have to be part of your object detection PR, which would use the code here to check sklearn is installed if users want to use (import) anything in `_experimental.object_detection`.

README.md
ly-kolena: Example 7 [here](https://pip.pypa.io/en/stable/cli/pip_install/#examples)

.circleci/continue_config.yml
```diff
@@ -187,16 +222,19 @@ workflows:
   ci:
     jobs:
       - ci-base:
-          name: ci-base-<< matrix.python-version >>
+          name: ci-base-<< matrix.python-version >><< matrix.extras >>
```
ly-kolena: I would change this to `ci-base-<< matrix.python-version >>-<< matrix.extras >>` and remove the `-` from `extras` value. (or `ci-base-<< matrix.extras >>-<< matrix.python-version >>`, to me, display name is less of a concern compared with dealing with extra dash in the value)

note: references also need to be updated.

README.md
```diff
@@ -23,6 +23,36 @@ model behaviors and take the mystery out of model development. Kolena helps you:
...
+
+Client builds can be installed directly from PyPI using any Python package manager such as pip:
+
+```zsh
+pip install kolena
+```
+
+Advanced use cases (eg. metrics computation) may require extra dependencies which can be installed by running:
+```zsh
+pip install `kolena[metrics]`
```
gordonhart: Remove backticks — quotes around this shouldn't be necessary in common shells, and backticks are separate thing entirely.

kolena/_extras/metrics/sklearn.py
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+try:
+    import sklearn.metrics as sklearn_metrics
+
+    sklearn_metrics.__name__
```
gordonhart: I would remove this in favor of suppressing the error that this exists to quiet — something like `# noqa: F401` on the import line

kolena/_extras/metrics/sklearn.py
```diff
@@ -0,0 +1,19 @@
...
+except ImportError:
+    raise ImportError("Package 'scikit-learn' not found; install 'metrics' extras with `pip install 'kolena[metrics]'`")
```
gordonhart, line 18: Are the quotes around `kolena[metrics]` required?

tests/unit_extras/__init__.py
gordonhart, line 1: @ly-kolena what do you think of this unit testing setup?

.circleci/continue_config.yml
```diff
@@ -187,16 +222,19 @@ workflows:
...
+              extras: ["", "-metrics"]
       - unit-test:
+          name: unit-test-<< matrix.python-version >><< matrix.extras >>
           matrix:
             parameters:
               python-version: [ "3.7", "3.8", "3.9", "3.10", "3.11" ]
+              extras: ["", "-metrics"]
           requires:
-            - ci-base-<< matrix.python-version >>
+            - ci-base-<< matrix.python-version >><< matrix.extras >>
```
ly-kolena: For consistency, please also update the `requires` and/or `matrix.extras` to integration test and other jobs that also depends on `ci-base`.

gordonhart (approved):
Looks nice! We'll want to split `kolena/_extras/metrics/__init__.py` apart at some point such that we can lazily import libraries as required, but this is great to get started.

Will want to get approval from @ly-kolena on the testing approach before merging.

tests/unit/_extras/test_extras_fail.py
```diff
@@ -0,0 +1,31 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import numpy as np
+import pytest
+
+
+@pytest.mark.parametrize(
```
ly-kolena: Nit: since this is to test import fail check, can remove these parametrizations.

kolena/_extras/metrics/sklearn.py
marko-polo-cheno: maybe not - I think at some point I ran into an error (didn't appear again) asking me to use quotes or `kolena\[metrics\]`

tests/unit_extras/__init__.py
ly-kolena: What do people think of using marker `metrics` to highlight tests that need matching extras dependency?

If things get tedious, we can use some pytest helper function to mark whole directory/test-file with a marker.

ly-kolena: Unless it's a concern to restructure unit-tests later, I'm also fine with refining the structure later when we integrate `sklearn` into new and/or existing workflows.

gordonhart: > What do people think of using marker metrics to highlight tests that need matching extras dependency?

I like this a lot. Easily extensible to additional extras when those become necessary.

ly-kolena: If going with the marker route, unit-test structure could look like this:

```
├── tests/
│   ├── unit/
│       ├── _extras/
│          ├── __init__.py
│          └── test_metrics.py
```

and we can put tests from `test_extras_fail.py` and `test_extras_metrics.py` together into `test_metrics.py`
```
def test__extras__metrics_import_check() -> None:
       with pytest.raises(ImportError):
        from kolena._extras.metrics import sklearn_metrics

        sklearn_metrics.precision_recall_fscore_support(np.array(["cat", "dog"]), np.array(["cat", "pig"]),
            average="macro")

@pytest.mark.metrics
@pytest.mark.parametrize(
    "y_true, y_pred, precision, recall",
    [
        (
            np.array(["cat", "dog", "pig", "cat", "dog", "pig"]),
            np.array(["cat", "pig", "dog", "cat", "cat", "dog"]),
            2 / 9,
            1 / 3,
        ),
    ],
)
def test__extras__sklearn__pr(y_true: np.ndarray, y_pred: np.ndarray, precision: float, recall: float) -> None:
    prec, rec, _, _ = sklearn_metrics.precision_recall_fscore_support(y_true, y_pred, average="macro")
    assert precision == pytest.approx(prec, abs=1e-5)
    assert recall == pytest.approx(rec, abs=1e-5)
```

CI can be adjusted
```
poetry run pytest -m "not metrics"
poetry run pytest -m "metrics"
```
respectively.

.circleci/continue_config.yml
```diff
@@ -187,36 +224,41 @@ workflows:
...
           requires:
-            - ci-base-<< matrix.python-version >>
+            - ci-base-<< matrix.python-version >>-<< matrix.extras >>
       - integration-test:
-          name: integration-test-misc-<< matrix.python-version >>
+          name: integration-test-misc-<< matrix.python-version >>-<< matrix.extras >>
           matrix:
             parameters:
               python-version: [ "3.9" ]
+              extras: ["", "metrics"]
```
ly-kolena: Nit: do we need both or one of them suffice?

marko-polo-cheno: ah I see, `requires` structures change, but not the actual key

ly-kolena (approved):

.circleci/continue_config.yml
gordonhart: Is this dangling hyphen a result of this change?

Screen Shot 2023-06-07 at 9 47 45 AM

marko-polo-cheno: Apologies for the delay! Yes, because the extra-input is either `""` or `"metrics"` and the key is like `f"ci-base-{version}-{extra}"`

gordonhart: I know that @ly-kolena has pointed in the other direction, but I'd like to avoid this dangling hyphen in these CI jobs. It's a low priority for sure

PR #59, opened 2023-06-06
Remove unused methods from kolena.workflow.Workflow

gordonhart:
Instance methods `kolena.workflow.Workflow.{get_evaluator,register_evaluator}` do not appear to be used within the codebase, and do not have external utility (interacted with via CLI). Removing these methods for clarity.


ly-kolena:
It was intended to be used by kolena users for use-cases that users may not want to manually interact with evaluator endpoints through CLI.

ly-kolena (approved):

gordonhart:
Closing — the impact of publicly exporting these functions is minimal, and removing them would be a breaking change. Not worth it.

PR #60, opened 2023-06-06
Update pyproject.toml

marko-polo-cheno:
Add markers


ly-kolena (approved):

gordonhart:
Why didn't this break CI on #57?

marko-polo-cheno:
> Why didn't this break CI on #57?

[from this](https://docs.pytest.org/en/7.1.x/example/markers.html#registering-markers), it seems like it is optional, but recommended

PR #62, opened 2023-06-07
Remove 'experimental.new-installer' from CI

gordonhart:
Seems that this `experimental.new-installer` option was deprecated in Poetry 1.4 and dropped from the [Poetry 1.5](https://python-poetry.org/blog/announcing-poetry-1.5.0/) version that now ships with the `cimg/python` container image. This PR removes it to bring CI green again — better to keep with the times than pin to an older Poetry version.

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -157,7 +157,7 @@ jobs:
                 command: circleci-agent step halt
       - checkout
       - restore_cache:
-          key: ci-cache-<< parameters.python-version >>-<< parameters.extras >>-{{ checksum "pyproject.toml" }}
+          key: ci-cache-<< parameters.python-version >><< parameters.extras >>-{{ checksum "pyproject.toml" }}
       - run:
           name: Run << parameters.pytest-group >> integration tests
           command: |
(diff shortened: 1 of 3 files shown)


.circleci/continue_config.yml
```diff
@@ -224,41 +221,39 @@ workflows:
...
-              extras: ["", "metrics"]
+              extras: ["", "-metrics"]
       - unit-test:
-          name: unit-test-<< matrix.python-version >>-<< matrix.extras >>
+          name: unit-test-<< matrix.python-version >><< matrix.extras >>
           matrix:
             parameters:
               python-version: [ "3.7", "3.8", "3.9", "3.10", "3.11" ]
-              extras: ["", "metrics"]
+              extras: ["", "-metrics"]
```
ly-kolena: In Mark's original PR, I asked him to remove the `-`, so that later handling of values, we don't have to think about it. Potentially we can use the value in `poetry install -- extras <extras>`. To me, the display of naming is less important than intuitive values.

gordonhart: I would much rather have the CI YAML miss a hyphen here and there than have to look at `ci-base-3.7-` on every commit — but this PR isn't the place for that change. Reverted

ly-kolena (approved):

gordonhart:
As a nice bonus, the new installer appears to deliver on its promise of speeding up installation. Sample times from the `text_summarization` example:

| Before | After |
| ------ | ------- |
| [4m13s](https://app.circleci.com/pipelines/github/kolenaIO/kolena/487/workflows/040beac0-2a8b-45c5-84a6-f9f34685703c/jobs/5722) | [2m25s](https://app.circleci.com/pipelines/github/kolenaIO/kolena/503/workflows/b8af3670-3322-4f97-9922-0c565aadc9ae/jobs/6002) |

PR #67, opened 2023-06-07
Create object detection PR and F1 plots

marko-polo-cheno:
PR and F1 plots for OD


kolena/_experimental/object_detection/utils.py
```diff
@@ -16,12 +16,120 @@
...
+    pr_curves.append(pr_curve)
+
+    # TODO: Uncomment when per class metrics are desired
+    # classes = sorted(y_true_by_label.keys())
+    # for label in classes:
+    #     y_true, y_score = y_true_by_label[label], y_score_by_label[label]
+    #     if len(y_true) > 0:
+    #         f1_curve, pr_curve = _compute_threshold_curves(np.array(y_true), np.array(y_score), label)
+    #         f1_curves.append(f1_curve)
+    #         pr_curves.append(pr_curve)
```
gordonhart: Would prefer not to land this commented-out code block

kolena/_experimental/object_detection/utils.py
```diff
@@ -16,12 +16,120 @@
...
+            y_score.append(bbox_inf.score)
+    return y_true, y_score, dict(y_true_by_label), dict(y_score_by_label)
+
+
+def _compute_threshold_curves(
+    y_true: np.ndarray,
+    y_score: np.ndarray,
+    label: str,
+) -> Tuple[Curve, Curve]:
+    if len(y_score) >= 501:
```
marko-polo-cheno, line 62: We do not want to have too many points, as it would effect the front end's performance.
By generating these points, we are guaranteed to get the optimal confidence threshold within an absolute distance of 0.001.

When there are fewer unique confidences, we know one of them must be the so-called "optimal".

kolena/_extras/metrics/sklearn.py
```diff
@@ -15,4 +15,4 @@
     import sklearn.metrics as sklearn_metrics  # noqa: F401
 
 except ImportError:
-    raise ImportError("Package 'scikit-learn' not found; install 'metrics' extras with `pip install 'kolena[metrics]'`")
+    raise ImportError("Package 'scikit-learn' not found; install 'metrics' extras with `pip install kolena[metrics]`")
```
marko-polo-cheno: no need to write as `pip install <quote>`

kolena/_experimental/object_detection/utils.py
```diff
@@ -16,12 +16,120 @@
...
+def compute_pr_f1_plots(
+    all_matches: List[MulticlassInferenceMatches],
+    curve_label: str = "baseline",
+) -> List[CurvePlot]:
+    """
+    Creates a PR (precision and recall) curve and F1-threshold (confidence threshold) curve for the multiclass object
+    detection workflow. For `n` labels, each plot has `n+1` curves. One for the test case, and one per label.
+
+    :param all_matches: a list of multiclass or single class matching results.
+    :param curve_label: the label of the main curve.
```
gordonhart: Nit: consistent capitalization, `A list` and `The label`

kolena/_extras/metrics/sklearn.py
gordonhart: Nit: `extras` should be singular, `extra`

tests/unit/_experimental/test_plots.py
marko-polo-cheno, line 1: please peek imports `import compute_pr_f1_plots` and metric markers - I've manually tested each test.

kolena/_experimental/object_detection/utils.py
```diff
@@ -16,12 +16,120 @@
...
+        for _, bbox_inf in image_bbox_matches.matched:  # TP (if above threshold)
+            y_true_by_label[bbox_inf.label].append(1)
+            y_score_by_label[bbox_inf.label].append(bbox_inf.score)
+            y_true.append(1)
+            y_score.append(bbox_inf.score)
+        for bbox_gt, _ in image_bbox_matches.unmatched_gt:  # FN
+            y_true.append(1)
+            y_score.append(-1)
+            y_true_by_label[bbox_gt.label].append(1)
+            y_score_by_label[bbox_gt.label].append(-1)
```
gordonhart: This causes strange plots for all-negative test cases. Is there a reason to prefer `-1` over `0`?

kolena/_experimental/object_detection/utils.py
```diff
@@ -16,12 +16,120 @@
...
+        for bbox_inf in image_bbox_matches.unmatched_inf:  # FP (if above threshold)
+            y_true_by_label[bbox_inf.label].append(0)
+            y_score_by_label[bbox_inf.label].append(bbox_inf.score)
+            y_true.append(0)
+            y_score.append(bbox_inf.score)
+    return y_true, y_score, dict(y_true_by_label), dict(y_score_by_label)
```
gordonhart: Are the dicts necessary? They don't seem to be used by the implementation.

kolena/_experimental/object_detection/utils.py
```diff
@@ -16,12 +16,120 @@
...
+            y_true,
+            y_pred,
+            average="binary",
+            zero_division=0,
+        )
+
+        precisions.append(precision)
+        recalls.append(recall)
+        f1s.append(f1)
+    return Curve(x=thresholds, y=f1s, label=label), Curve(x=recalls, y=precisions, label=label)
```
gordonhart: Minor, but feels like this interface should return lists, and the `Curve` should be created in the same context where the `CurvePlot` is created.

kolena/_experimental/object_detection/utils.py
```diff
@@ -16,12 +16,120 @@
...
+    )
+
+    pr = CurvePlot(
+        title="Precision vs. Recall",
+        x_label="Recall",
+        y_label="Precision",
+        curves=pr_curves,
+    )
+
+    return [threshold_f1, pr]
```
gordonhart: It's a weird public API to have the creation of the F1-vs-T plot and the PR curve coupled like this. Is there a way to keep the perf benefits of coupling them without jamming both into a single external-facing function?

kolena/_experimental/object_detection/utils.py
marko-polo-cheno: Old code, forces FNs to always be under every threshold.
Effect of making it 0 will ensure a y-intercept on a F1-threshold curve, which is nice.

marko-polo-cheno: Update on this, I'm keeping it to be -1 to force `unmatched_gt`s to always be under the threshold, even if the threshold is 0. When the threshold is 0, and I set an `unmatched_gt`'s `y_score` to be 0 and filter with `>=`, then `y_true` is 1 and `y_pred` is 1, which is a TP.
What's different now is that every threshold is guaranteed to be non-negative, so no `-1`s exist in the x-axis anymore.

kolena/_experimental/object_detection/utils.py
```diff
@@ -16,21 +16,133 @@
...
+    return np.array(y_true), np.array(y_score)
+
+
+def _compute_threshold_curves(
+    y_true: np.ndarray,
+    y_score: np.ndarray,
+    label: str,
+) -> List[Curve]:
+    if len(y_score) >= 501:
+        thresholds = list(np.linspace(min(abs(y_score)), max(y_score), 501))
```
y27choi, line 63: We used to ignore `max(y_score)` as a threshold by adding `[:-1]` to the list, and if I remember it correctly, it was to avoid plotting `(0, 0)`.   Was this dropped intentionally?

kolena/_experimental/object_detection/utils.py
```diff
@@ -16,21 +16,133 @@
...
+    :param curve_label: The label of the curve.
+    :param plot: The specified plot type to return (`pr`, `f1`, or `all`). All plots returned by default.
+    :return: :class:`CurvePlot`s for the PR curve and/or F1-threshold curve respectively.
+    """
+    y_true, y_score = _compute_sklearn_arrays(all_matches)
+    curves = _compute_threshold_curves(y_true, y_score, curve_label)
+    pr_curves = [curves[0]]
+    f1_curves = [curves[1]]
+
+    if plot == "pr":
```
y27choi: Instead of having duplicated lines for these `CurvePlot` instantiation, how about create them once and return what has been asked by `plot`?

tests/unit/_experimental/test_plots.py
```diff
@@ -12,18 +12,1561 @@
...
+                        ScoredLabeledBoundingBox((1, 1), (2, 2), "a", 0),
+                    ],
+                ),
+            ],
+            CurvePlot(
+                title="F1-Score vs. Confidence Threshold",
+                x_label="Confidence Threshold",
+                y_label="F1-Score",
+                curves=[
+                    Curve(x=[0.0], y=[0.0], label="zeros with unmatched gt and unmatched inf"),
```
y27choi: How do people feel about a PR / F1 plot with a single point at the origin (e.g. when there are no valid matches)?
I don't know how it will look like on the webapp but I think we should skip plotting these.

tests/unit/_experimental/test_plots.py
```diff
@@ -546,7 +2127,9 @@ def test__confusion__matrix(
 )
 def test__confusion__matrix__fails(
     test_name: str,
-    matchings: MulticlassInferenceMatches,
+    matchings: List[MulticlassInferenceMatches],
 ) -> None:
+    from kolena._experimental.object_detection.utils import compute_confusion_matrix_plot
```
y27choi, line 1239: It seems like this import is moved within the scope of each confusion matrix test functions.  What's the reason behind this change?

tests/unit/_experimental/test_plots.py
```diff
@@ -12,18 +12,1561 @@
...
+
+@pytest.mark.metrics
+@pytest.mark.parametrize(
+    "plot, test_name, matchings, curveplot",
+    [
+        (
+            "pr",
+            "tps and fps and fns pr plot",
+            [
+                MulticlassInferenceMatches(
```
y27choi: Can't we avoid this copypasta blocks by making this into a variable?

y27choi (approved):
Confirmed that my comments from [previous PR](https://github.com/kolenaIO/kolena/pull/54/files) have been addressed.  I left some questions / comments / suggestions, but the logic looks good to me.

y27choi (approved):
Confirmed that my comments from [previous PR](https://github.com/kolenaIO/kolena/pull/54/files) have been addressed.  I left some questions / comments / suggestions, but the logic looks good to me.

kolena/_experimental/object_detection/utils.py
```diff
@@ -16,21 +16,133 @@
...
+        recalls.append(recall)
+        f1s.append(f1)
+
+    return [Curve(x=recalls, y=precisions, label=label), Curve(x=thresholds, y=f1s, label=label)]
+
+
+def compute_pr_f1_plots(
+    all_matches: List[Union[MulticlassInferenceMatches, InferenceMatches]],
+    curve_label: str = "baseline",
+    plot: Literal["pr", "f1", "all"] = "all",
```
gordonhart: Let's hop on a call to hash out this API tomorrow

kolena/_experimental/object_detection/utils.py
marko-polo-cheno: This drop is intentional as we now check filter with `>=` instead of `>`. The end effect is similar, and we no longer need to delete the largest score.

tests/unit/_experimental/test_plots.py
marko-polo-cheno: Do you mean the curve should not be made, or the front end will especially flag these curves?

tests/unit/_experimental/test_plots.py
marko-polo-cheno: `kolena/_experimental/object_detection/utils.py` imports sklearn, so if the import is outside of the pytest metrics markers, then it will fail with import errors

tests/unit/_experimental/test_plots.py
y27choi: I'm wondering if we should omit these plots.

kolena/_experimental/object_detection/utils.py
```diff
@@ -82,51 +77,58 @@ def _compute_threshold_curves(
     return [Curve(x=recalls, y=precisions, label=label), Curve(x=thresholds, y=f1s, label=label)]
 
 
-def compute_pr_f1_plots(
+def compute_pr_plot(
     all_matches: List[Union[MulticlassInferenceMatches, InferenceMatches]],
-    curve_label: str = "baseline",
-    plot: Literal["pr", "f1", "all"] = "all",
-) -> List[CurvePlot]:
+    curve_label: str = "",
```
gordonhart: `Optional[str] = None`

tests/unit/_experimental/test_plots.py
marko-polo-cheno: I think we should, similar to how we omit confusion matrices with only 1 label.
I've adjusted it so that these types of plots aren't created.

gordonhart (approved):
💪

PR #68, opened 2023-06-08
Add optimal F1 threshold util

marko-polo-cheno:
branched from https://github.com/kolenaIO/kolena/pull/67


marko-polo-cheno:
[CI fail](https://app.circleci.com/pipelines/github/kolenaIO/kolena/551/workflows/be4634ba-27b6-4360-8921-710b7ef2700a/jobs/7054) was a result of sklearn version diff

[New sklearn](https://github.com/scikit-learn/scikit-learn/blob/364c77e047ca08a95862becf40a04fe9d4cd2c98/sklearn/metrics/_ranking.py#L797)
[Old sklearn](https://github.com/scikit-learn/scikit-learn/blob/baf828ca126bcb2c0ad813226963621cafe38adb/sklearn/metrics/_ranking.py#L779) for py 3.7.1

kolena/_experimental/object_detection/utils.py
```diff
@@ -99,7 +99,18 @@ def compute_optimal_f1(
         _, _, y_true_by_label, y_score_by_label = _compute_sklearn_arrays_by_class(all_bbox_matches)
         for label in sorted(y_true_by_label.keys()):
             y_true, y_score = y_true_by_label[label], y_score_by_label[label]
+
+            # python 3.7.1's sklearn is different
+            if np.all(y_true == 0):
+                optimal_thresholds[label] = 0
+                continue
```
gordonhart: Would prefer not to have version-specific handling in the implementation like this. Is there a version pin we can update to ensure that users of 3.7 have a `scikit-learn` package with consistent behavior?

marko-polo-cheno: sklearn1.1+ requires python3.8+
For python3.7, we have [sklearn1.0](https://scikit-learn.org/1.0/modules/generated/sklearn.metrics.precision_recall_curve.html#sklearn.metrics.precision_recall_curve). The code here is different, and causes some unit tests to fail. I can remove the comment on 3.7.1, as `np.all(y_true == 0) => optimal_threshold == 0` should be true for any case, not version specific. Was noted to explain why a random new piece of logic existed, but not sure if it is the best solution.

kolena/_experimental/object_detection/utils.py
```diff
@@ -177,3 +177,85 @@ def compute_confusion_matrix_plot(
...
+            gt_label: str = gt_or_pair[0].label if type(gt_or_pair) is tuple else gt_or_pair.label
+            y_true.append(1)
+            y_score.append(-1)
+            y_true_by_label[gt_label].append(1)
+            y_score_by_label[gt_label].append(-1)
+        for bbox_inf in image_bbox_matches.unmatched_inf:  # FP (if above threshold)
+            y_true.append(0)
+            y_score.append(bbox_inf.score)
+            y_true_by_label[bbox_inf.label].append(0)
+            y_score_by_label[bbox_inf.label].append(bbox_inf.score)
```
gordonhart: Can we reuse the existing `_compute_sklearn_arrays` implementation for this? Reduces testing burden and ensures consistency.

kolena/_experimental/object_detection/utils.py
```diff
@@ -177,3 +177,85 @@ def compute_confusion_matrix_plot(
...
+    f1_scores = 2 * precision * recall / (precision + recall)
+    f1_scores = np.nan_to_num(f1_scores, nan=0)
+    max_f1_index = np.argmax(f1_scores)
+    if f1_scores[max_f1_index] == 0:
+        return 0.0
+    else:
+        return float(thresholds[max_f1_index])
+
+
+def compute_optimal_f1(
```
gordonhart: There isn't much utility in providing a single function for both single-class and multi-class here, and it forces users of this to deal with an annoying union return type. Would prefer to export separate `compute_optimal_f1_score` and `compute_optimal_f1_score_multiclass` functions

kolena/_experimental/object_detection/utils.py
```diff
@@ -177,3 +177,85 @@ def compute_confusion_matrix_plot(
...
+
+    return np.array(y_true), np.array(y_score), y_true_by_label_np, y_score_by_label_np
+
+
+def _compute_optimal_f1_with_arrays(
+    y_true: np.ndarray,
+    y_score: np.ndarray,
+) -> float:
+    if np.all(y_true == 0):
+        return 0.0
```
gordonhart: Can you add a comment here explaining what this condition means in terms of behavior?

kolena/_experimental/object_detection/utils.py
```diff
@@ -244,18 +238,32 @@ def _compute_optimal_f1_with_arrays(
...
-    all_bbox_matches: List[Union[MulticlassInferenceMatches, InferenceMatches]],
-) -> Union[float, Dict[str, float]]:
+def compute_optimal_f1_threshold(
+    all_matches: List[Union[MulticlassInferenceMatches, InferenceMatches]],
+) -> float:
+    """
+    Computes the optimal F1 threshold for matchings.
+
+    :param all_matches: A list of matching results.
+    :return: A the optimal F1 threshold value.
```
gordonhart: Nit: `A the`

kolena/_experimental/object_detection/utils.py
```diff
@@ -182,25 +182,18 @@ def compute_confusion_matrix_plot(
...
             gt_label: str = gt_or_pair[0].label if type(gt_or_pair) is tuple else gt_or_pair.label
-            y_true.append(1)
-            y_score.append(-1)
             y_true_by_label[gt_label].append(1)
             y_score_by_label[gt_label].append(-1)
         for bbox_inf in image_bbox_matches.unmatched_inf:  # FP (if above threshold)
-            y_true.append(0)
-            y_score.append(bbox_inf.score)
             y_true_by_label[bbox_inf.label].append(0)
             y_score_by_label[bbox_inf.label].append(bbox_inf.score)
```
gordonhart: Structurally this still looks very similar to `_compute_sklearn_arrays` — how can we guarantee that `y_true` and `y_true_by_label` are equivalent when one is created upstream and the other is created here? Logic can be further refactored to share implementation.

marko-polo-cheno: Good call!

kolena/_experimental/object_detection/utils.py
```diff
@@ -177,3 +177,97 @@ def compute_confusion_matrix_plot(
...
+                unmatched_inf=[inf for inf in match.unmatched_inf if inf.label == label],
+            )
+            for match in all_matches
+        ]
+
+        y_true, y_score = _compute_sklearn_arrays(filtered_matchings)
+        y_true_by_label[label] = y_true
+        y_score_by_label[label] = y_score
+
+    y_true, y_score = _compute_sklearn_arrays(all_matches)
```
gordonhart: Do we need to compute and return these? They are not used by the only caller of this function.

gordonhart (approved):

kolena/_experimental/object_detection/utils.py
marko-polo-cheno: If someone were to plot n+1 pr curves for n classes in a test case, returning might be useful. I'll take it out for now.

kolena/_experimental/object_detection/utils.py
```diff
@@ -177,3 +177,97 @@ def compute_confusion_matrix_plot(
...
+
+
+def compute_optimal_f1_threshold(
+    all_matches: List[Union[MulticlassInferenceMatches, InferenceMatches]],
+) -> float:
+    """
+    Computes the optimal F1 threshold for matchings.
+
+    :param all_matches: A list of matching results.
+    :return: The optimal F1 threshold value.
```
y27choi: It would be nice to add here what to expect in case of invalid F1 cases.

kolena/_experimental/object_detection/utils.py
```diff
@@ -208,8 +208,7 @@ def _compute_sklearn_arrays_by_class(
         y_true_by_label[label] = y_true
         y_score_by_label[label] = y_score
 
-    y_true, y_score = _compute_sklearn_arrays(all_matches)
-    return y_true, y_score, y_true_by_label, y_score_by_label
+    return y_true_by_label, y_score_by_label
```
gordonhart: As one last non-blocking nit, this return type seems like it would be more ergonomic than the current one:

```python
Dict[str, Tuple[np.ndarray, np.ndarray]]
```

kolena/_experimental/object_detection/utils.py
```diff
@@ -265,8 +263,8 @@ def compute_optimal_f1_threshold_multiclass(
     """
     optimal_thresholds: Dict[str, float] = {}
 
-    y_true_by_label, y_score_by_label = _compute_sklearn_arrays_by_class(all_matches)
-    for label in y_true_by_label.keys():
-        y_true, y_score = y_true_by_label[label], y_score_by_label[label]
+    y_true_score_by_label = _compute_sklearn_arrays_by_class(all_matches)
+    for label in y_true_score_by_label.keys():
+        y_true, y_score = y_true_score_by_label[label][0], y_true_score_by_label[label][1]
```
gordonhart: `for label, (y_true, y_score) in arrays_by_label.items()`

kolena/_experimental/object_detection/utils.py
```diff
@@ -177,3 +177,96 @@ def compute_confusion_matrix_plot(
...
+
+def _compute_optimal_f1_with_arrays(
+    y_true: np.ndarray,
+    y_score: np.ndarray,
+) -> float:
+    # optimal threshold is 0 if there is no relevant or true label
+    if np.all(y_true == 0):
+        return 0.0
+
+    precision, recall, thresholds = sklearn_metrics.precision_recall_curve(y_true, y_score)
```
y27choi, line 220: I believe if `y_true` and `y_score` is both empty then this call will fail.  Can we add a check at the top and add a unit test for this case?

marko-polo-cheno: thanks for catching this 👍

y27choi (approved):
LGTM!

PR #69, opened 2023-06-09
Use 'insider' features from mkdocs-material and mkdocstrings

gordonhart:
Kolena sponsors the [mkdocs-material](https://github.com/squidfunk/mkdocs-material) and [mkdocstrings](https://github.com/mkdocstrings/mkdocstrings) projects and uses various "insider" features from each. This PR sets up the dependency management scheme to support building with and without access to these repos, and updates the existing documentation to leverage "insider" features. See `CONTRIBUTING.md` for details.

Fixes KOL-2486


gordonhart:
Should be ready to review. Looking for a sanity check from @ly-kolena and @munkyshi on the devx here. All necessary information should be in `CONTRIBUTING.md`.

ly-kolena (approved):

.circleci/continue_config.yml
```diff
@@ -61,13 +61,46 @@ jobs:
...
+      - run:
+          name: Install system-level build dependencies
+          command: |
+            sudo apt-get update
+            sudo apt-get install -y libcairo2-dev
+      - run:
+          name: Build documentation without 'insiders' packages
+          command: poetry run mkdocs build
+      - add_ssh_keys:  # add github.com SSH key fingerprint to access private repo forks
+          fingerprints: ["15:8b:d4:ac:1f:cd:2f:d3:92:a7:4c:aa:46:81:0b:7d"]
```
munkyshi, line 97: Not familiar with what the fingerprint corresponds to -- does this leak anything / should this be a secret envvar?

CONTRIBUTING.md
```diff
@@ -13,3 +16,20 @@ To build static documentation:
...
+declared in [`pyproject.toml`](pyproject.toml) reference the publicly available package sources.
+
+To build the documentation with "insider" features, run the [`setup_insiders.sh`](docs/setup_insiders.sh) script:
+
+```
+./docs/setup_insiders.sh
+```
+
+Note that this script requires an SSH key in your environment with access to the [kolenaIO](https://github.com/kolenaIO)
+organization.
```
munkyshi, line 32: Is the implication here that anyone working on this from outside of `kolenaIO` is flying half-blind when iterating on docs?

munkyshi (commented):
~What does this mean for someone who isn't part of~

Asked as part of a comment

pyproject.toml
```diff
@@ -64,9 +64,11 @@ pytest = "^7"
 pytest-cov = "^4.0.0"
 pytest-depends = "^1.0.1"
 mkdocs = "^1.4.3"
-mkdocs-material = "^9.1.15"
-mkdocstrings = {extras = ["python"], version = "^0.22.0"}
 cairosvg = "^2.7.0"
+mkdocs-material = "^9.1.15"  # insiders fork installed out-of-band in docs/setup_insiders.sh
+mkdocstrings = { version = ">0.20,<1", extras = ["python"] }
+mkdocs-git-committers-plugin-2 = "^1.1.2"
+mkdocs-git-revision-date-localized-plugin = "^1.2.0"
```
munkyshi, line 71: To be clear, these are not `insiders` features, correct? A `poetry install` will still work, even for someone without `kolenaIO` access?

.circleci/continue_config.yml
gordonhart: It's not private — followed the [official CircleCI setup instructions](https://circleci.com/docs/github-integration/#create-additional-github-ssh-keys), which end with putting this fingerprint in your config file. AFAICT it's effectively an ID for that particular SSH key, allowing you to reference a specific key with specific authorizations from job definitions.

CONTRIBUTING.md
gordonhart: No — the docs build fine without access to the insider repos. Is that not clear enough in `CONTRIBUTING.md`?

> In order to preserve the capability to build documentation without access to these private "insider" repos, the docs dependencies declared in [`pyproject.toml`](pyproject.toml) reference the publicly available package sources.

pyproject.toml
gordonhart: Neither of these repos are private (note that the pins don't point anywhere special, and will thus just install the packages from PyPI).

`mkdocs-git-revision-date-localized-plugin` is a non-insider feature that is enabled in `mkdocs.yml` via the `git-revision-date-localized` block in `plugins`.

The other one is an insider feature, which is only enabled when the `git-committers` block is present (see `mkdocs.insiders.yml`). However, since it can be installed publicly, I think it's preferable to have it as a part of the main dev dependency specification in `pyproject.toml`, rather than install as a part of `setup_insiders.sh`.

PR #70, opened 2023-06-09
Make job name looks better

ly-kolena:


munkyshi (approved):

gordonhart (approved):
[Not important] Would prefer to simply omit it rather than have the `-none` suffix but this is definitely better than the dangling hyphen.

PR #71, opened 2023-06-09
Customize docs themes

gordonhart:
Apply Kolena branding on top of the `default` light style and `slate` dark style. Including a branded header bar:

Screen Shot 2023-06-09 at 12 30 09 PM

Fixes KOL-2455


mkdocs.yml
```diff
@@ -51,14 +51,16 @@ theme:
     text: Inter
   palette:
     - media: "(prefers-color-scheme: light)"
-      scheme: default  # TODO: custom kolena-light theme
+      scheme: default
+      accent: cyan
       toggle:
-        icon: material/toggle-switch
+        icon: material/lightbulb
```
ly-kolena: Non-blocking preference discussion: personally lightbulb is a little un-intuitive to me. I've seen sun/moon icon, what do others think of that if they are available?

ly-kolena (approved):

mkdocs.yml
gordonhart: Updated to use sun/moon iconography.

PR #73, opened 2023-06-10
bulk testcase processing

ly-kolena:
#### _[NOT SELECTED]_ Interface design 1
```
# usage pattern 1
with TestCase.bulk_process():
    complete_test_case = TestCase("LIUYUAN TEST ImageNet-1k :: complete (1k labels)", test_samples=test_samples,
        reset=True)
    sub_test_cases = [TestCase(
        f"LIUYUAN TEST ImageNet-1k :: split-{i} (1k labels)",
        test_samples=test_samples[i * per_test_case:(i+1) * per_test_case],
        reset=True
    ) for i in range(split)]

test_suite = TestSuite("LIUYUAN TEST ImageNet-1k :: complete (1k labels)",
    test_cases=[complete_test_case, *sub_test_cases],
    reset=True)

# usage pattern 2
with TestCase.bulk_process():
    with test_suite.test_cases[0].edit() as editor:
        editor.description("foo")
        editor.add(TestSample(locator="hello world"))

    with test_suite.test_cases[1].edit() as editor:
        editor.description("bar")
        editor.add(TestSample(locator="hello world"))

# usage pattern 3
with TestCase.bulk_process():
    for test_case in test_suite.test_cases:
        with test_case.edit() as editor:
            editor.description(f"{test_case.description}-prime")
            editor.remove(test_samples[0])
```
The critical piece to make clear is that, in the `with`-context, each `test_case` object remains as the conceptual "old" entity. They are only updated to the "new" identity (same object) after the `with`-context completes. This applies to both load/init and edit. 

* Pros: the usage pattern looks similar to existing single-test-case pattern.
* Cons: the note above.

#### **[SELECTED]** Interface design 2

```
test_cases = TestCase.init_many(
        [
            TestCaseEditData(name="test-case 1", description="desc 1"),
            TestCaseEditData(name="test-case 2" description="desc 2", test_samples=test_samples[3:6]),
            TestCaseEditData(name="test-case 3", description="desc 3", test_samples=test_samples[6:9]),
        ],
    )
```

### DIFF
--- a/kolena/workflow/test_case.py
+++ b/kolena/workflow/test_case.py
@@ -160,14 +160,6 @@ class TestCase(Frozen, WithTelemetry, metaclass=ABCMeta):
             self.description = other.description
             self.workflow = other.workflow
 
-    def _populate_from_data(self, data: CoreAPI.EntityData) -> None:
-        assert_workflows_match(self.workflow.name, data.workflow)
-        with self._unfrozen():
-            self._id = data.id
-            self.name = data.name
-            self.version = data.version
-            self.description = data.description
-
     @classmethod
     def _create_from_data(cls, data: CoreAPI.EntityData) -> "TestCase":
         assert_workflows_match(cls.workflow.name, data.workflow)
@@ -206,7 +198,6 @@ class TestCase(Frozen, WithTelemetry, metaclass=ABCMeta):
         :return: The newly created test case.
         """
         cls._validate_test_samples(test_samples)
-
         request = CoreAPI.CreateRequest(name=name, description=description or "", workflow=cls.workflow.name)
         res = krequests.post(endpoint_path=API.Path.CREATE.value, data=json.dumps(dataclasses.asdict(request)))
         krequests.raise_for_status(res)
--- a/kolena/workflow/test_case.py
+++ b/kolena/workflow/test_case.py
@@ -13,7 +13,9 @@
 # limitations under the License.
 import dataclasses
 import json
+import time
 from abc import ABCMeta
+from collections import defaultdict
 from contextlib import contextmanager
 from typing import Iterator
 from typing import List
@@ -24,6 +26,7 @@ import pandas as pd
 from pydantic import validate_arguments
 from pydantic.dataclasses import dataclass
 
+from kolena._api.v1.core import BulkProcessStatus
 from kolena._api.v1.core import TestCase as CoreAPI
 from kolena._api.v1.generic import TestCase as API
 from kolena._utils import krequests
@@ -38,6 +41,7 @@ from kolena._utils.instrumentation import telemetry
 from kolena._utils.instrumentation import WithTelemetry
 from kolena._utils.serde import from_dict
 from kolena._utils.validators import ValidatorConfig
+from kolena.errors import IncorrectUsageError
 from kolena.errors import NotFoundError
 from kolena.workflow import GroundTruth
 from kolena.workflow import TestSample
@@ -46,6 +50,33 @@ from kolena.workflow._datatypes import TestSampleDataFrame
 from kolena.workflow._validators import assert_workflows_match
 from kolena.workflow.test_sample import _METADATA_KEY
 from kolena.workflow.workflow import Workflow
+
+
+def _to_editor_data_frame(
+    edits: List[Tuple[TestSample, Optional[GroundTruth], bool]],
+    test_case_name: Optional[str] = None,
+) -> TestCaseEditorDataFrame:
+    records = [
+        (
+            test_case_name,
+            test_sample._data_type().value,
+            test_sample._to_dict(),
+            test_sample._to_metadata_dict(),
+            ground_truth._to_dict() if ground_truth is not None else None,
+            remove,
+        )
+        for test_sample, ground_truth, remove, in edits
+    ]
+    columns = [
+        "test_case_name",
+        "test_sample_type",
+        "test_sample",
+        "test_sample_metadata",
+        "ground_truth",
+        "remove",
+    ]
+    df = pd.DataFrame(records, columns=columns)
+    return TestCaseEditorDataFrame(validate_df_schema(df, TestCaseEditorDataFrame.get_schema(), trusted=True))
 
 
 class TestCase(Frozen, WithTelemetry, metaclass=ABCMeta):
@@ -129,6 +160,14 @@ class TestCase(Frozen, WithTelemetry, metaclass=ABCMeta):
             self.description = other.description
             self.workflow = other.workflow
 
+    def _populate_from_data(self, data: CoreAPI.EntityData) -> None:
+        assert_workflows_match(self.workflow.name, data.workflow)
+        with self._unfrozen():
+            self._id = data.id
+            self.name = data.name
+            self.version = data.version
+            self.description = data.description
+
     @classmethod
     def _create_from_data(cls, data: CoreAPI.EntityData) -> "TestCase":
         assert_workflows_match(cls.workflow.name, data.workflow)
@@ -167,6 +206,7 @@ class TestCase(Frozen, WithTelemetry, metaclass=ABCMeta):
         :return: The newly created test case.
         """
         cls._validate_test_samples(test_samples)
+
         request = CoreAPI.CreateRequest(name=name, description=description or "", workflow=cls.workflow.name)
         res = krequests.post(endpoint_path=API.Path.CREATE.value, data=json.dumps(dataclasses.asdict(request)))
         krequests.raise_for_status(res)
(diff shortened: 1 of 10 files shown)


gordonhart:
As far as the interface goes, I like the first option far better.

Are both `batch_init` and `batch_edit` necessary, or can we get away with a single context function?

kolena/workflow/test_case.py
```diff
@@ -48,6 +52,90 @@
...
+
+        with test_case._unfrozen():
+            test_case._enter_bulk_edit()
+        self.data[test_case.name] = (test_case, edit_data, df)
+
+    @staticmethod
+    def check_ignored(test_case: "TestCase", edit_data: CoreAPI.SingleProcessRequest, has_test_samples: bool) -> None:
+        if edit_data.description is not None and test_case.description != edit_data.description and not edit_data.reset:
+            log.warn(f"test case '{test_case.name}' already exists, not updating description when reset=False")
+        # there will be a false negative when target test-case is version 1
```
ly-kolena: Given the information client has, under bulk operation, it cannot tell the difference between
1. version-0 test case being updated, or
2. version-1 test case intended to updated without `reset`, thus ignored by server.

using `test_case.version > 1` will have false negative (no warning shown) for 2), and using `test_case.version > 0` will have false positive for 1).

Any preference?

kolena/workflow/test_case.py
```diff
@@ -48,6 +52,115 @@
 from kolena.workflow.workflow import Workflow
 
 
+_BULK_PROCESSOR = contextvars.ContextVar("batch_loader")
+
+
+def check_edit(method):
```
gordonhart: Should be `_check_edit` to avoid publicly exporting from this module

kolena/workflow/test_case.py
```diff
@@ -141,6 +293,7 @@ def _create_from_data(cls, data: CoreAPI.EntityData) -> "TestCase":
         obj._freeze()
         return obj
 
+    @check_edit
     def _hydrate(self, test_samples: List[Tuple[TestSample, GroundTruth]], description: Optional[str] = None) -> None:
         if len(test_samples) == 0:
             log.warn("no test samples provided, unable to populate test case")
```
gordonhart: How will `create` and `load` work within a bulk editor context?

ly-kolena: Should behave the same as single test-case `__init__`. Any specifics you're asking?

ly-kolena: If you're asking about class methods `create` and `load`, currently they are not included in bulk processing, and therefore will proceed as normal.

kolena/workflow/test_case.py
```diff
@@ -332,3 +493,48 @@ def edit(self, reset: bool = False) -> Iterator[Editor]:
...
+        """
+        if not hasattr(cls, "workflow"):
+            raise NotImplementedError(f"{cls.__name__} must implement class attribute 'workflow'")
+
+        bulk_processor = _BulkProcessor()
+        token = _BULK_PROCESSOR.set(bulk_processor)
+        try:
+            yield
+
+            log.info(f"loading {len(bulk_processor.data)} test cases")
```
gordonhart: Should this be `editing`?

kolena/workflow/test_case.py
```diff
@@ -332,3 +493,48 @@ def edit(self, reset: bool = False) -> Iterator[Editor]:
...
+
+            bulk_process_response = bulk_processor.process(cls.workflow.name)
+            bulk_processor.commit(bulk_process_response.test_cases)
+
+            log.success(f"edited {len(bulk_processor.data)} test cases")
```
gordonhart: When playing around with this locally, this value was zero:

```
kolena> loading 10 test cases
kolena> edited 0 test cases
```

Is that expected?

kolena/workflow/test_case.py
gordonhart: Given all of the constraints here, I'm wondering if using the same `TestCase` definition is still worth it:

- Must exit context before trying to create test suite
- Can't use `create` or `load` in context
- Weirdness around version 0 and version 1
- Uncertainty about how `reset` operates

Would it ultimately be cleaner to expose an API like this?

```python3
class TestCase:
    ...
    def bulk_create(
        test_cases: List[Tuple[str, List[Tuple[TestSample, GroundTruth]]]],
        reset: bool = False,  # throw if any test cases already exist, unless reset=True (overwrite)
    ) -> List[TestCase]:
        ...
```

What do you think?

kolena/workflow/test_case.py
ly-kolena: Thanks for catching, fixed and added test.

kolena/workflow/test_case.py
ly-kolena: fixed

kolena/workflow/test_case.py
y27choi: I like the current API because it’s easy to use.  You just need to remember to add the context `with TestCase.bulk_process():` to utilize the bulk processing.

> Must exit context before trying to create test suite

I personally don’t have any issue with this.  Since the context manager name includes TestCase, I think it makes sense to only create / edit test cases.

> Can’t use create or load in context

I think being able to support this in context is important.  I’m sure some users are still using create and load

> Weirdness around version 0 and version 1

I didn’t understand this part so can’t comment on this.

> Uncertainty about how `reset` operates

Do we know what the uncertainty here is?  Having a `reset` option is crucial when implementing the seeding script.

kolena/workflow/test_case.py
```diff
@@ -332,3 +366,83 @@ def edit(self, reset: bool = False) -> Iterator[Editor]:
...
+    @classmethod
+    def init_many(cls, data: List[TestCaseData], reset: bool = False) -> List["TestCase"]:
+        """
+        **Experimental** Create, load or edit multiple test cases.
+
+        ```python
+        test_cases = TestCase.init_many([
+            ("test-case 1", [(test_sample_1, ground_truth_1), ...])
+            ("test-case 2", [(test_sample_2, ground_truth_2), ...])
+        ]
```
gordonhart: Nit: close parentheses

kolena/workflow/test_case.py
```diff
@@ -48,6 +52,33 @@
 from kolena.workflow.workflow import Workflow
 
 
+def _to_editor_data_frame(
+    edits: List[Tuple[TestSample, Optional[GroundTruth], bool]],
```
gordonhart, line 56: Should the `Optional` around GT be removed?

kolena/workflow/test_case.py
```diff
@@ -332,3 +354,91 @@ def edit(self, reset: bool = False) -> Iterator[Editor]:
...
+        :param data: A list of tuples where each tuple is a test case name and a set of test samples and ground truths
+                     tuples for the test case.
+        :param reset: If a test case of the same name already exists, overwrite with the provided test_samples.
+        :return: The test cases.
+
+
+        """
+
+        if not hasattr(cls, "workflow"):
+            raise NotImplementedError(f"{cls.__name__} must implement class attribute 'workflow'")
```
gordonhart, line 393: Is this necessary?

ly-kolena: I shot myself in the foot when writing test like this `TestCase.init_many()`, when in fact I should use `DummyTestCase.init_many`. The method needs class workflow information to initialize test-case instances.

kolena/workflow/test_case.py
```diff
@@ -332,3 +354,91 @@ def edit(self, reset: bool = False) -> Iterator[Editor]:
...
+        if len({name for name, _ in data}) != len(data):
+            raise IncorrectUsageError("Multiple edits to the same test case")
+
+        for _, test_samples in data:
+            if test_samples:
+                cls._validate_test_samples(test_samples)
+
+        cls._validate_test_samples()
+
+        log.info(f"processing {len(data)} test cases")
```
gordonhart: Minor: since we use the term `init`, should this be `initializing`?

kolena/workflow/test_case.py
```diff
@@ -332,3 +354,91 @@ def edit(self, reset: bool = False) -> Iterator[Editor]:
...
+
+        test_cases = []
+        statuses = defaultdict(int)
+        for test_case_data in bulk_response.test_cases:
+            test_cases.append(cls._create_from_data(test_case_data.data))
+            statuses[test_case_data.status] += 1
+
+        end = time.time()
+        status_msg = ", ".join([f"{s.value} {statuses[s]}" for s in BulkProcessStatus])
+        log.info(f"processed {len(data)} test cases: {status_msg} in {end - start} seconds")
```
gordonhart: Nit: cap timestamp to 3 decimal places: `{end - start:0.3f}`

gordonhart: Also same question about `initialized` over `processed`

gordonhart (approved):
Played around locally, behaves well :+1:

Thank you for iterating here

kolena/workflow/test_case.py
gordonhart: Isn't this validated via `_validate_test_samples`?

ly-kolena: Since `_validate_test_samples` access `cls.workflow` directly instead of checking its existence, relying on it would see the same exception as `TestCase._init_many()`.

PR #74, opened 2023-06-12
Consistent iconography between docs and platform

gordonhart:
This PR brings in various [Blueprint](https://github.com/palantir/blueprint) icons as SVG for usage in documentation. This setup allows us to have consistent iconography between these docs and our app.

Fixes KOL-2494
Fixes PRO-125

PR #72, opened 2023-06-09
Tutorial documentation (Part 1)

gordonhart:
- Quickstart (rewrite using age estimation example in this repo)
- Installing `kolena`
- Building a Workflow
- Packaging for Automated Evaluation
- FAQ (mostly a rewrite rather than a port)

Note that no effort has been made to preserve URL compatibility with the existing pages on docs.kolena.io, as most pages are not exactly 1:1.

This PR also populates various index files.

Fixes KOL-2458

PR #75, opened 2023-06-13
Tutorial documentation (Part 2)

gordonhart:
Following up on #72 and #74, this PR adds tutorial documentation for the various "core concepts" in Kolena (workflow, test case, test suite, model).

Remaining work I'm tracking after this PR and before this week's release:

- Landing page
- Polish pass


docs/core-concepts/test-suite.md
```diff
@@ -3,3 +3,164 @@ icon: kolena/test-suite-16
 ---
 
 # :kolena-test-suite-20: Test Case & Test Suite
```
gordonhart, line 5: Anything to add to this guide? It feels like it could use more content, but I think this is an acceptable starting point.

ly-kolena (approved):

PR #78, opened 2023-06-14
Render documentation social cards; other cleanups

gordonhart:
Fixes KOL-2503
Fixes KOL-2505
Fixes KOL-2506
Fixes KOL-2514
Fixes KOL-2515
Fixes KOL-2512
Fixes KOL-2519
Fixes KOL-2517

### DIFF
--- a/mkdocs.insiders.yml
+++ b/mkdocs.insiders.yml
@@ -26,6 +26,15 @@ theme:
     - content.tooltips
 
 plugins:
+  social:
+    # cache: false
+    # debug: true
+    cards_layout: custom
+    cards_layout_dir: docs/layouts
+    # cards_layout_options:
+    #   font_family: Inter
+    #   background_color: "#6434c1"
+    #   background_image: docs/assets/images/background.svg  # TODO: not properly applied -- debug
   typeset:
   git-committers:
     repository: kolenaIO/kolena
--- a/mkdocs.insiders.yml
+++ b/mkdocs.insiders.yml
@@ -27,14 +27,8 @@ theme:
 
 plugins:
   social:
-    # cache: false
-    # debug: true
     cards_layout: custom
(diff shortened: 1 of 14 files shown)


mkdocs.insiders.yml
```diff
@@ -26,6 +26,15 @@ theme:
...
 plugins:
+  social:
+    # cache: false
+    # debug: true
+    cards_layout: custom
+    cards_layout_dir: docs/layouts
+    # cards_layout_options:
+    #   font_family: Inter
+    #   background_color: "#6434c1"
+    #   background_image: docs/assets/images/background.svg  # TODO: not properly applied -- debug
```
ly-kolena: Are the comments here intentional to be kept?

mkdocs.yml
```diff
@@ -52,7 +52,8 @@ nav:
               - Base Definitions: reference/legacy/built-in/base.md
   - Help & FAQ:
     - Help & FAQ: faq/index.md
-  - Sign in: https://app.kolena.io
+  - Sign in:
+    - Sign in: sign-in/index.md
```
ly-kolena: What's the benefit of this alternative route?

Side note: previous implementation works for me, but new one doesn't until I hit reload button. This is from me running the server locally.

mkdocs.insiders.yml
gordonhart: No, thanks. Removed

mkdocs.yml
gordonhart: This is a kludge to render a link next to the `Sign in` button. It was working originally but now does not seem to work for me either. Might have to rip out and lose the external link icon...

gordonhart: Removed in favor of a more subtle arrow character.

ly-kolena (approved):

PR #80, opened 2023-06-15
Publish docs as kolenabot

gordonhart:
Publish documentation on release to `release/docs` as user [kolenabot](https://github.com/kolenabot), allowing us to set up author-based branch protections on the `release/docs` branch. Using a custom branch instead of the default `gh-pages` branch also gives us a little added protection in case anybody accidentally runs `mkdocs gh-deploy` as kolenabot.

Fixes KOL-2508

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,8 +4,6 @@ on:
   push:
     tags:
       - "*"
-    branches:
-      - gh/docs-publish-bot
 
 jobs:
   release:
@@ -56,10 +54,7 @@ jobs:
           git config user.name 'kolenabot'
           git config user.email 'bot@kolena.io'
           ./docs/setup_insiders.sh
-          poetry run mkdocs gh-deploy --verbose --strict --remote-branch bot/gh-pages --config-file mkdocs.insiders.yml
-
-      - name: Fail
-        run: exit 1
+          poetry run mkdocs gh-deploy --verbose --strict --remote-branch release/gh-pages --config-file mkdocs.insiders.yml
 
       - name: Push 'kolena' dist to Test PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -54,7 +54,7 @@ jobs:
           git config user.name 'kolenabot'
           git config user.email 'bot@kolena.io'
           ./docs/setup_insiders.sh
-          poetry run mkdocs gh-deploy --verbose --strict --remote-branch release/gh-pages --config-file mkdocs.insiders.yml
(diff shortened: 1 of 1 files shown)


gordonhart:
Looks like this [works as expected](https://github.com/kolenaIO/kolena/commits/bot/gh-pages)!

.github/workflows/release.yml
```diff
@@ -51,8 +51,10 @@ jobs:
 
       - name: Build 'kolena' documentation and push to GitHub Pages
         run: |
+          git config user.name 'kolenabot'
```
ly-kolena, line 54: Is this the same entity where we used to configure ssh key in steps above?

munkyshi (approved):
If someone changes their git settings and declares themself to be `kolenabot`, will that go through the branch protection settings? I imagine not, but we should verify.

.github/workflows/release.yml
gordonhart: Yep, the SSH key belongs to kolenabot.

ly-kolena: side question: config here would make tracking better. However, technically we can protect this branch without these username/email config extras, right?

ly-kolena (approved):

gordonhart:
> If someone changes their git settings and declares themself to be kolenabot, will that go through the branch protection settings? I imagine not, but we should verify.

Hmm, the commit might go through or might be rejected if their SSH key doesn't belong to kolenabot. I'm not sure what author verification protections GitHub provides here (if any).

In either case, it doesn't seem too important to me to address — from my perspective, branch protections are useful to protect against accidental mishaps, not intentional malice. The user performing this attack would need to be inside our organization, and would need to configure themselves as kolenabot and run `mkdocs gh-deploy --remote-branch release/docs`. Too many steps to do accidentally.

.github/workflows/release.yml
gordonhart: > technically we can protect this branch without these username/email config extras

What protections do you have in mind? I've set up protections on `release/docs` to only allow commits from kolenabot.

ly-kolena: Same as you have setup. I'm just curious since ssh-key is used to push to the branch, should be able to verify identity **without** `git config user.name` etc. (editted)

Pure discussion, the PR is good in correctness.

gordonhart: Ah, good question (related to behavior mentioned here: https://github.com/kolenaIO/kolena/pull/80#issuecomment-1593407357).

I haven't been able to find relevant documentation about what protections GitHub actually applies when you restrict branches to certain users. I have to believe that they use the underlying creds/SSH key, but also wouldn't be entirely surprised if it somehow relied on the author information. After this PR lands, both will be set up for kolenabot.

PR #88, opened 2023-06-20
Experimental OD Workflow

marko-polo-cheno:
Notes for future improvements: per class plots when class selection is enabled, AP numbers may need revising


kolena/_experimental/object_detection/evaluator.py
```diff
@@ -0,0 +1,343 @@
...
+            count_FP=len(fp),
+            count_FN=len(fn),
+            count_Confused=len(confused),
+            has_TP=len(tp) > 0,
+            has_FP=len(fp) > 0,
+            has_FN=len(fn) > 0,
+            has_Confused=len(confused) > 0,
+            max_confidence_above_t=max(scores) if len(scores) > 0 else None,
+            min_confidence_above_t=min(scores) if len(scores) > 0 else None,
+            **{threshold_key(label): value for label, value in thresholds.items()},
```
gordonhart: Question — should we filter this to only the labels in this test case? How about only the labels in this particular sample? Can be a lot of information for highly multiclass problems.

kolena/_experimental/object_detection/evaluator.py
```diff
@@ -0,0 +1,343 @@
...
+        )
+
+    def compute_f1_optimal_thresholds(
+        self,
+        configuration: ThresholdConfiguration,
+        inferences: List[Tuple[TestSample, GroundTruth, Inference]],
+    ) -> None:
```
gordonhart: This feels like it should return the optimal thresholds, not just compute them and cache them. If not, would update the name to make the fact that this populates the cache explicit

kolena/_experimental/object_detection/evaluator.py
```diff
@@ -0,0 +1,343 @@
...
+            Inferences=tp_count + fp_count,
+            TP=tp_count,
+            FN=fn_count,
+            FP=fp_count,
+            Confused=confused_counter,
+            TPR=tpr_counter / samples if samples > 0 else 0.0,
+            FPR=fpr_counter / samples if samples > 0 else 0.0,
+            Precision=precision,
+            Recall=recall,
+            F1=2 * precision * recall / (precision + recall) if (precision + recall) > 0 else 0.0,
```
gordonhart: Use `kolena.workflow.metrics.f1_score`

kolena/_experimental/object_detection/evaluator.py
```diff
@@ -0,0 +1,343 @@
...
+                tpr_counter += 1
+            if has_fp:
+                fpr_counter += 1
+            if has_tp or has_fn or has_fp:
+                samples += 1
+        tp_count = len(m_matched)
+        fp_count = len(m_unmatched_inf)
+        fn_count = len(m_unmatched_gt)
+        precision = tp_count / (tp_count + fp_count) if (tp_count + fp_count) > 0 else 0
+        recall = tp_count / (tp_count + fn_count) if (tp_count + fn_count) > 0 else 0
```
gordonhart: Use `kolena.workflow.metrics.{precision,recall}`

kolena/_experimental/object_detection/evaluator.py
```diff
@@ -0,0 +1,343 @@
...
+        baseline_pr_plot = compute_pr_plot(all_bbox_matches)
+        baseline_pr_plot = baseline_pr_plot.curves[0] if baseline_pr_plot is not None else None
+        average_precision = (
+            compute_average_precision(baseline_pr_plot.y, baseline_pr_plot.x)
+            if baseline_pr_plot is not None and precision > 0
+            else 0.0
+        )
+
+        # compute nested metrics per class
+        labels = {gt.label for _, gts, _ in inferences for gt in gts.bboxes}
```
gordonhart: This is already performed above — does it need to be redeclared?

kolena/_experimental/object_detection/evaluator.py
```diff
@@ -0,0 +1,343 @@
...
+                iou_threshold=configuration.iou_threshold,
+            )
+            for _, ground_truth, inference in inferences
+        ]
+        self.locators_by_test_case[test_case.name] = [ts.locator for ts, _, _ in inferences]
+        tp_count = sum(im.count_TP for im in metrics)
+        fp_count = sum(im.count_FP for im in metrics)
+        fn_count = sum(im.count_FN for im in metrics)
+        precision = tp_count / (tp_count + fp_count) if (tp_count + fp_count) > 0 else 0
+        recall = tp_count / (tp_count + fn_count) if (tp_count + fn_count) > 0 else 0
```
gordonhart: Use `kolena.workflow.metrics.{precision,recall}`

kolena/_experimental/object_detection/evaluator.py
```diff
@@ -0,0 +1,343 @@
...
+            Objects=tp_count + fn_count,
+            Inferences=tp_count + fp_count,
+            TP=tp_count,
+            FN=fn_count,
+            FP=fp_count,
+            TPR=sum(1 for im in metrics if im.has_TP) / len(metrics) if len(metrics) > 0 else 0.0,
+            FPR=sum(1 for im in metrics if im.has_FP) / len(metrics) if len(metrics) > 0 else 0.0,
+            Precision=precision,
+            Recall=recall,
+            F1=2 * precision * recall / (precision + recall) if (precision + recall) > 0 else 0.0,
```
gordonhart: Use `kolena.workflow.metrics.f1_score`

kolena/_experimental/object_detection/evaluator.py
```diff
@@ -0,0 +1,343 @@
...
+                inf
+                for inf in inference.bboxes
+                if inf.label in labels and inf.score >= max(configuration.min_confidence_score, thresholds[inf.label])
+            ],
+            ignored_ground_truths=ground_truth.ignored_bboxes,
+            mode="pascal",
+            iou_threshold=configuration.iou_threshold,
+        )
+        tp = [inf for _, inf in bbox_matches.matched if inf.label in labels]
+        fp = [inf for inf in bbox_matches.unmatched_inf if inf.label in labels]
```
gordonhart: Is `inf.label in labels` necessary to check? We're already filtering the inputs by label

kolena/_experimental/object_detection/evaluator.py
```diff
@@ -0,0 +1,343 @@
...
+        self,
+        ground_truth: GroundTruth,
+        inference: Inference,
+        configuration: ThresholdConfiguration,
+        labels: Set[str],  # the labels being tested in this test case
+    ) -> TestSampleMetrics:
+        assert configuration is not None, "must specify configuration"
+        thresholds = self.get_confidence_thresholds(configuration)
+        bbox_matches: MulticlassInferenceMatches = match_inferences_multiclass(
+            ground_truth.bboxes,
```
gordonhart, line 69: Do we need to filter these by label?

kolena/_experimental/object_detection/evaluator.py
```diff
@@ -0,0 +1,343 @@
...
+                for inf in inference.bboxes
+                if inf.label in labels and inf.score >= max(configuration.min_confidence_score, thresholds[inf.label])
+            ],
+            ignored_ground_truths=ground_truth.ignored_bboxes,
+            mode="pascal",
+            iou_threshold=configuration.iou_threshold,
+        )
+        tp = [inf for _, inf in bbox_matches.matched if inf.label in labels]
+        fp = [inf for inf in bbox_matches.unmatched_inf if inf.label in labels]
+        fn = [gt for gt, _ in bbox_matches.unmatched_gt if gt.label in labels]
```
gordonhart: Is `gt.label in labels` necessary to check? My expectation is that every box in `gt.bboxes` will have its label in `labels`.

kolena/_experimental/object_detection/evaluator.py
```diff
@@ -0,0 +1,343 @@
...
+
+    def compute_aggregate_label_metrics(
+        self,
+        matchings: List[MulticlassInferenceMatches],
+        label: str,
+        thresholds: Dict[str, float],
+    ) -> ClassMetricsPerTestCase:
+        m_matched = []
+        m_unmatched_gt = []
+        m_unmatched_inf = []
```
gordonhart: Should these be `n_*`? Why `m`?

kolena/_experimental/object_detection/evaluator.py
```diff
@@ -0,0 +1,343 @@
...
+            filter(
+                None,
+                [
+                    compute_f1_plot_multiclass(all_bbox_matches),
+                    compute_f1_plot(all_bbox_matches),
+                    compute_pr_plot_multiclass(all_bbox_matches),
+                    compute_pr_plot(all_bbox_matches),
+                    compute_confusion_matrix_plot(all_bbox_matches),
+                ],
+            ),
```
gordonhart: I'm confused as to what's going on here. We're plotting both `f1_plot` and `f1_plot_multiclass`? Neither have optional return types.

gordonhart (commented):
Good start, should be ready to land after comments are addressed.

Follow-ups I am tracking before this can be promoted out of `kolena._experimental`:

- [ ] Document everything
- [ ] Thoroughly unit test evaluator
- [ ] Bug bash with real results for single-class and multiclass cases

kolena/_experimental/object_detection/evaluator.py
```diff
@@ -0,0 +1,343 @@
...
+    def compute_image_metrics(
+        self,
+        ground_truth: GroundTruth,
+        inference: Inference,
+        configuration: ThresholdConfiguration,
+        labels: Set[str],  # the labels being tested in this test case
+    ) -> TestSampleMetrics:
+        assert configuration is not None, "must specify configuration"
+        thresholds = self.get_confidence_thresholds(configuration)
+        bbox_matches: MulticlassInferenceMatches = match_inferences_multiclass(
```
y27choi, line 81: It looks like `bbox_matches` are computed multiple times for each test case.  Is this something we can cache?

kolena/_experimental/object_detection/evaluator.py
```diff
@@ -0,0 +1,343 @@
...
+            ground_truth.bboxes,
+            [
+                inf
+                for inf in inference.bboxes
+                if inf.label in labels and inf.score >= max(configuration.min_confidence_score, thresholds[inf.label])
```
y27choi: Have you verified that confidence thresholding prior to matching is what the standard is?

kolena/_experimental/object_detection/evaluator.py
```diff
@@ -0,0 +1,343 @@
...
+            ignored_ground_truths=ground_truth.ignored_bboxes,
+            mode="pascal",
+            iou_threshold=configuration.iou_threshold,
+        )
+        tp = [inf for _, inf in bbox_matches.matched if inf.label in labels]
+        fp = [inf for inf in bbox_matches.unmatched_inf if inf.label in labels]
+        fn = [gt for gt, _ in bbox_matches.unmatched_gt if gt.label in labels]
+
+        confused = [inf for gt, inf in bbox_matches.unmatched_gt if inf is not None and gt.label in labels]
+        non_ignored_inferences = [inf for _, inf in bbox_matches.matched] + bbox_matches.unmatched_inf
```
y27choi: `tp + bbox_matches.unmatched_inf`?

kolena/_experimental/object_detection/evaluator.py
```diff
@@ -0,0 +1,343 @@
...
+
+    def compute_test_sample_metrics(
+        self,
+        test_case: TestCase,
+        inferences: List[Tuple[TestSample, GroundTruth, Inference]],
+        configuration: Optional[ThresholdConfiguration] = None,
+    ) -> List[Tuple[TestSample, TestSampleMetrics]]:
+        assert configuration is not None, "must specify configuration"
+        # compute thresholds to cache values for subsequent steps
+        self.compute_f1_optimal_thresholds(configuration, inferences)
```
y27choi: So here we are assuming that the first test case is a complete one.  1. Can we avoid this assumption? 2. If not, how can we communicate this clearly to the users?

kolena/_experimental/object_detection/evaluator.py
marko-polo-cheno: By sample is a good idea 👍

kolena/_experimental/object_detection/evaluator.py
marko-polo-cheno: Renamed to `inf_match_*`

kolena/_experimental/object_detection/evaluator.py
```diff
@@ -0,0 +1,343 @@
...
+            has_fp = False
+            for gt, inf in match.matched:
+                if gt.label == label:
+                    m_matched.append((gt, inf))
+                    has_tp = True
+            for gt, inf in match.unmatched_gt:
+                if gt.label == label:
+                    m_unmatched_gt.append(gt)
+                    has_fn = True
+                    if inf is None:
```
y27choi: Should this be `inf is not None`?

kolena/_experimental/object_detection/evaluator.py
marko-polo-cheno: Just wanted to temporarily inspect test case plots separately. Will fix signatures/logic!

kolena/_experimental/object_detection/evaluator.py
```diff
@@ -0,0 +1,343 @@
...
+            Class=label,
+            Threshold=thresholds[label],
+            TestSamples=samples,
+            Objects=tp_count + fn_count,
+            Inferences=tp_count + fp_count,
+            TP=tp_count,
+            FN=fn_count,
+            FP=fp_count,
+            Confused=confused_counter,
+            TPR=tpr_counter / samples if samples > 0 else 0.0,
```
y27choi: Should `TPR` and `FPR` be based on samples or based on objects?

kolena/_experimental/object_detection/evaluator.py
```diff
@@ -0,0 +1,343 @@
...
+                samples += 1
+        tp_count = len(m_matched)
+        fp_count = len(m_unmatched_inf)
+        fn_count = len(m_unmatched_gt)
+        precision = tp_count / (tp_count + fp_count) if (tp_count + fp_count) > 0 else 0
+        recall = tp_count / (tp_count + fn_count) if (tp_count + fn_count) > 0 else 0
+        all_bbox_matches = [
+            InferenceMatches(matched=m_matched, unmatched_gt=m_unmatched_gt, unmatched_inf=m_unmatched_inf),
+        ]
+        baseline_pr_plot = compute_pr_plot(all_bbox_matches)
```
y27choi: Should we check for `precision > 0` before computing pr plot?

kolena/_experimental/object_detection/evaluator.py
gordonhart: I'd consider omitting TPR — not clear to me what it means, and why it's meaningful to include in addition to Recall.

kolena/_experimental/object_detection/evaluator.py
marko-polo-cheno: I thought it was a good idea before, but I'm not sure how Python's internal memory management works. Little concerned about cacheing a complete test case of 100k samples for 1k test cases, and having that memory space occupied for the entire duration of an evaluation.

kolena/_experimental/object_detection/evaluator.py
marko-polo-cheno: I've taken a look at PASCAL VOC and COCO's evaluation code. They do not have a concept of confidence thresholding.
This doesn't mean that confidence thresholding prior to matching is wrong tho, because eventually matches of poor confidence will be filtered out anyways. The only difference is shorter computation time, and "better" matchings (lower iou if the "best match" of a lower confidence is removed, but the new match would be above the threshold).

kolena/_experimental/object_detection/evaluator.py
marko-polo-cheno: Should be objects. I made it samples-based because it is hard to define how many true negatives there are in `FPR = FP / (FP + TN)`.
Will take both out.

kolena/_experimental/object_detection/evaluator.py
marko-polo-cheno: It's not hard to cache different thresholds per `(testcase.name, configuration.display_name)` but I think users should have a complete test case as a benchmark. If different f1 optimal thresholds are used per test case, it wont be straightforward to compare pr curves or results across different test cases or different models. Do you think outlining this requirement in documentation/guides would cover this issue?

y27choi: Yes, f1 optimal thresholds should be the same for every test case. However, we are assuming that the **first** test case is a complete test case, and f1 optimal thresholds are computed on the **first** test case. Users should be aware of this assumption and seed test suites accordingly.

kolena/_experimental/object_detection/evaluator.py
y27choi: > "better" matchings (lower iou if the "best match" of a lower confidence is removed, but the new match would be above the threshold).

Is this true? Since predictions are sorted by the confidence score, I don't think there should be any difference in the matching results.

kolena/_experimental/object_detection/evaluator.py
y27choi: How long does it take to run evaluation on COCO dataset in comparison to the built-in OD workflow?

kolena/_experimental/object_detection/evaluator.py
marko-polo-cheno: hmm, right. I guess pre-filtering only improves performance then.

gordonhart (approved):
Looks good to land as an initial pass. We'll want to follow up with PRs to address the remaining bug bash items and branch based on single-class/multi-class config, but those would be best to implement and review in follow-ups instead of this branch.

Nice work here!

kolena/_experimental/object_detection/evaluator.py
```diff
@@ -0,0 +1,367 @@
...
+
+        average_precision = 0.0
+        if precision > 0:
+            baseline_pr_curve = compute_pr_curve(all_bbox_matches_for_one_label)
+            if baseline_pr_curve is not None:
+                average_precision = compute_average_precision(baseline_pr_curve.y, baseline_pr_curve.x)
+
+        return ClassMetricsPerTestCase(
+            Class=label,
+            TestSamples=samples_with_label,
```
y27choi: We might need to revisit this later once we have the multi-model support for nested metrics.  I think it makes more sense to use # of objects for statistical significance calculation for OD.

y27choi (approved):
Looks good to me as well!  👍

kolena/_experimental/object_detection/evaluator.py
marko-polo-cheno: Good thought - I'll add this too! Just thought it'd be good to know how many images are relevant to a particular class in a test case.

marko-polo-cheno: Actually, the front end should have enough info with `TP`, `Objects`, and `Inferences`.

kolena/_experimental/object_detection/workflow.py
```diff
@@ -87,7 +87,7 @@ class TestSampleMetrics(MetricsTestSample):
 @dataclass(frozen=True)
 class ClassMetricsPerTestCase(MetricsTestCase):
     Class: str
-    TestSamples: int
+    n_test_samples: int
```
gordonhart: We should follow the same naming convention across all metrics. What would the ideal name for this field be? `nImages`?

marko-polo-cheno: I think ideal naming is `# Images` to follow the test cases, but maybe that should be renamed to `Images`, like `Objects` and `Inferences`. I think `Images` > `nImages`.

PR #92, opened 2023-06-21
Add meta tags to social card layout

gordonhart:
#78 updated the social cards to use a custom layout, but neglected to port the meta tags to this new layout (custom social cards is an [experimental feature](https://squidfunk.github.io/mkdocs-material/setup/setting-up-social-cards/#customization) and its documentation could use fleshing out). This PR fixes that and should unbreak social cards once released.

Fixes KOL-2503

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -86,10 +86,13 @@ jobs:
       - restore_cache:
           key: ci-cache-<< parameters.python-version >>-<< parameters.extras >>-{{ checksum "pyproject.toml" }}
       - run:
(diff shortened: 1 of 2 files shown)


ly-kolena (approved):

ly-kolena:
dumb question, how do you test (verify) social card? I've checked the generated cache image, but didn't see it in live.

gordonhart:
These particular changes can be verified by looking for `<meta property>` tags in the document `<head>`. For other changes, I've had success copy/pasting into Slack or using an online checker. I haven't found a good way to validate local changes without deploying.

PR #93, opened 2023-06-21
Add Datadog RUM to docs.kolena.io

gordonhart:
Tracking will help us understand docs engagement. Note that this, along with all other common user monitoring tools, will be blocked by ad blockers.

Fixes KOL-2500

### DIFF
--- a/docs/overrides/main.html
+++ b/docs/overrides/main.html
@@ -14,7 +14,7 @@
       applicationId: '32c8b59a-335b-4a60-8320-52cf94822079',
       site: 'datadoghq.com',
       service: 'docs.kolena.io',
-      env: 'production',
+      env: window.location.hostname,
       // Specify a version number to identify the deployed version of your application in Datadog
       // version: '1.0.0',
       sessionSampleRate: 100,
--- a/docs/overrides/main.html
+++ b/docs/overrides/main.html
@@ -1,0 +1,31 @@
+{% extends "base.html" %}
+
+{% block extrahead %}
+{{ super() }}
+<script>
+  (function(h,o,u,n,d) {
+    h=h[d]=h[d]||{q:[],onReady:function(c){h.q.push(c)}}
+    d=o.createElement(u);d.async=1;d.src=n
+    n=o.getElementsByTagName(u)[0];n.parentNode.insertBefore(d,n)
+  })(window,document,'script','https://www.datadoghq-browser-agent.com/us1/v4/datadog-rum.js','DD_RUM')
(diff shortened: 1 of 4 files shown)


docs/overrides/main.html
```diff
@@ -0,0 +1,28 @@
+{% extends "base.html" %}
+
+{% block extrahead %}
```
ly-kolena, line 7: for my own learning, why `extrahead` instead of `script`?

ly-kolena (approved):

docs/overrides/main.html
```diff
@@ -0,0 +1,28 @@
...
+{{ super() }}
+<script>
+  (function(h,o,u,n,d) {
+    h=h[d]=h[d]||{q:[],onReady:function(c){h.q.push(c)}}
+    d=o.createElement(u);d.async=1;d.src=n
+    n=o.getElementsByTagName(u)[0];n.parentNode.insertBefore(d,n)
+  })(window,document,'script','https://www.datadoghq-browser-agent.com/us1/v4/datadog-rum.js','DD_RUM')
+  window.DD_RUM.onReady(function() {
+    window.DD_RUM.init({
+      clientToken: 'pub77d1be21718df74d03f852860d7f6b7f',
```
ly-kolena: Non-blocking nit: This looks like some template library. Is it possible to use env variable for these 2 configurations at build time?

gordonhart: Should be possible to read in environment variables. Seeing as this token and app ID are public information (they will be visible to anybody loading the site), is there benefit to doing that?

ly-kolena: Feels cleaner to separate data from checked-in codes. Potentially we can also update api key etc. without update the sdk codes if data is kept out. Small things like this. Although if we follow that path, might need to check how to make local instance render fine without it.

Non-blocking and up to your judgement.

docs/overrides/main.html
gordonhart: Slight preference to include async analytics in the `<head>` (via `extrahead` block) instead of the footer (via the `scripts` block) to get the script running earlier. I've typically seen analytics in the header for this reason. In practice either would work fine here

docs/overrides/main.html
gordonhart: Went ahead and implemented it. These variables are defined as GitHub repo variables (not secrets), loaded into `mkdocs.yml`, and referenced in this template via `{{ config.extra.dd_rum.client_token }}`. Looks like it works.

PR #96, opened 2023-06-22
Advanced usage docs for nested aggregate metrics

gordonhart:
Add guide for when/why/how nested aggregate metrics should be used.


docs/advanced-usage/nesting-aggregate-metrics.md
```diff
@@ -0,0 +1,119 @@
...
+
+Let's consider the case of a multiclass object detection task with objects of type `Airplane`, `Boat`, and `Car`.
+When a test case contains images with each of these three classes, test-case-level metrics are the average (e.g.
+[micro](../metrics/averaging-methods.md#micro-average), [macro](../metrics/averaging-methods.md#macro-average), or
+[weighted](../metrics/averaging-methods.md#weighted-average)) of class-level metrics across each of these three classes.
+
+Using [macro-averaged](../metrics/averaging-methods.md#macro-average) precision, recall, and F1 score, and mean average
+precision score (mAP) across all images in the test case.
+
+| Test Case | # Images | `Precision_macro` | `Recall_macro` | `F1_macro` | `mAP` |
```
marko-polo-cheno: I usually see the averaging method first, before the metric eg macro_Precision because it reads "macro-averaged precision"

docs/advanced-usage/nesting-aggregate-metrics.md
```diff
@@ -0,0 +1,119 @@
...
+    mAP: float
+```
+
+These metrics tell us how well the model performs in `Example Scenario` across all classes, but they don't tell us
+anything about per-class model performance. Within each test case, we'd also like to see precision, recall, F1, and AP
+scores:
+
+| `Class` | `N` | `Precision` | `Recall` | `F1` | `AP` |
+| --- | --- | --- | --- | --- | --- |
+| `Airplane` | 5 | 0.5 | 0.5 | 0.5 | 0.5 |
```
marko-polo-cheno: probably just a github thing, but personally like centred alignment if possible

docs/advanced-usage/nesting-aggregate-metrics.md
```diff
@@ -0,0 +1,119 @@
...
+on the [:kolena-results-16: Results](https://app.kolena.io/redirect/results) page.
+
+### Statistical Significance
+
+When comparing models, Kolena only highlights performance improvements and regressions that are likely to be
+statistically significant. These calculations take the number of samples into account. When reporting nested metrics,
+certain fields, such as `N` in the above `PerClassMetrics` example, are interpreted as population size and used for
+statistical significance calculations.
+
+To ensure that highlighted improvements and regressions in these nested metrics are statistically significant, populate
```
marko-polo-cheno: I think `Statistical Significance` can be a little shorter: Kolena recommends including sample sizes for pretty highlights of red/green using statistical significance calculations

y27choi:
Completed a quick review and the overall direction looks good to me.  It includes exactly what I would need to get started with Nested Metrics.

marko-polo-cheno (approved):

marko-polo-cheno:
Might be nice to see an example/screenshot in the docs to supplement, and could make understanding these concepts seem friendlier.

gordonhart:
> Might be nice to see an example/screenshot in the docs to supplement, and could make understanding these concepts seem friendlier.

I definitely see the motivation for including screenshots. For now, we're avoiding platform screenshots in docs, because styles change and screenshots get stale. Might reevaluate this once the docs are more fleshed out.

gordonhart:
Ready for review — requesting reviews from @ly-kolena and @y27choi.

docs/advanced-usage/nesting-aggregate-metrics.md
```diff
@@ -0,0 +1,118 @@
+---
+icon: kolena/nest-16
```
ly-kolena: really really nit and optional:

None builtin icons I quickly checked totally make sense or intuitive, but any of `Array`, `tree diagram`, or `layout: hierarchy` feels slightly better?

y27choi (approved):
Looks great!

docs/advanced-usage/nesting-aggregate-metrics.md
```diff
@@ -0,0 +1,118 @@
...
+---
+
+# :kolena-nest-20: Nesting Aggregate Metrics
```
ly-kolena: Nit: does using `Composite` vs `Nesting` make a difference?

docs/advanced-usage/nesting-aggregate-metrics.md
```diff
@@ -0,0 +1,118 @@
...
+one `str`-type column. This column, `Class` in the above example, is pinned to the left when displaying nested metrics
+on the [:kolena-results-16: Results](https://app.kolena.io/redirect/results) page.
+
+### Statistical Significance
+
+When comparing models, Kolena highlights performance improvements and regressions that are likely to be statistically
+significant. The number of samples being evaluated factors into these calculations.
+
+For nested metrics, certain fields like `N` in the above `PerClassMetrics` example are used as the population size for
+statistical significance calculations. To ensure that highlighted improvements and regressions in these nested metrics
```
ly-kolena: Do we have a docs page for this computation that can be linked from here? Would also be helpful if there's an example how this is would be computed under per-class case.

ly-kolena (approved):

docs/advanced-usage/nesting-aggregate-metrics.md
gordonhart: `kolena.workflow.Composite` is a separate (and unrelated) concept — I'd prefer to use separate terminology for this feature. Do you have any suggestions over `Nesting`? Definitely room for improvement.

ly-kolena: Trying to find a word that can be related to all 3 highlighted use-cases, multiclass (slice/stratification), ensembles (collection), and pipeline (sequences/trees/chains) and came up short.

`compound`, `combined`, `consolidated`? If none looks like an improvement, `Nesting` it is.

docs/advanced-usage/nesting-aggregate-metrics.md
gordonhart: Updated to `diagram-tree` 👍

docs/advanced-usage/nesting-aggregate-metrics.md
gordonhart: We do not. Will track as a potential follow-up, though it's not clear that this explanation belongs in the developer guide instead of in its current place on the Results page, where you can play around with the slider to see the effect on your data.

docs/advanced-usage/nesting-aggregate-metrics.md
gordonhart: Went with `Nested Test Case Metrics` per discussion on Slack 👍

PR #98, opened 2023-06-22
metrics :: geometry matcher

y27choi:
### Linked issue(s):
[KOL-2533](https://linear.app/kolena/issue/KOL-2533/port-geometry-matcher-doc-to-docskolenaio)

* Adds geometry matcher doc
* Cleaned up IoU doc — breaking up long lines of text (there is no actual edit to the page)
* TODO: Images will need to be updated to support dark mode and to have consistent style 


docs/metrics/geometry-matcher.md
```diff
@@ -0,0 +1,275 @@
+---
+search:
+  exclude: true
+---
+
+# Geometry Matcher
```
gordonhart: Thoughts on `Geometry Matching` instead of `Geometry Matcher`?

docs/metrics/geometry-matcher.md
```diff
@@ -0,0 +1,275 @@
...
+
+Geometry matcher finds the best possible match, given the sets of the ground truth and prediction polygons for each
+image. The geometry matcher utility is a building block for any object detection metrics, such as precision, recall,
+and average precision. In this guide, we will focus on 2D object detection workflow, specifically 2D bounding box
+matching.
+
```
gordonhart: Add `## Overview` header?

docs/metrics/geometry-matcher.md
```diff
@@ -0,0 +1,275 @@
+---
+search:
+  exclude: true
+---
+
+# Geometry Matcher
+
+Geometry matcher finds the best possible match, given the sets of the ground truth and prediction polygons for each
```
gordonhart: We should use the term `inference` instead of `prediction` consistently. Is there any reason to prefer `prediction`?

docs/metrics/geometry-matcher.md
```diff
@@ -0,0 +1,275 @@
...
+image. The geometry matcher utility is a building block for any object detection metrics, such as precision, recall,
+and average precision. In this guide, we will focus on 2D object detection workflow, specifically 2D bounding box
+matching.
```
gordonhart: Suggested opening section:

> Geometry matching is the process of finding the best possible inference for a given ground truth. It is the building block for downstream metrics like precision and recall in any geometry-based workflow, such as 2D and 3D object detection and instance segmentation.
>
> While it sounds simple, geometry matching is surprisingly challenging and full of edge cases! In this guide, we'll focus on 2D object detection—specifically 2D bounding box matching—to learn about geometry matching algorithms.

docs/metrics/geometry-matcher.md
```diff
@@ -0,0 +1,275 @@
...
+
+1. Loop through all images in your dataset;
+2. Loop through all labels;
+3. Get predictions and ground truths with the evaluating label;
+4. Sort predictions by descending confidence score;
+5. Check against all ground truths and find a ground truth that results in maximum IoU;
+6. Check for the following criteria for a valid match:
+	1. This ground truth is not matched yet AND
+	2. The IoU is greater than or equal to the IoU threshold;
+7. Repeat 5-6 on the next prediction;
```
gordonhart: Do you think this provides value, @y27choi? Would a reader of these docs care?

y27choi: For me personally, it helps me understand the logic a bit better with the pseudocode.  It was easier to highlight the difference when comparing matching algorithm variations.  I learned this matching algorithm so I can implement it so my view on this could be biased.

I take that you don't find it useful?  Is it more noise to you?  Will it make it better if I hide in in a collapsible?

gordonhart: Put in a collapsible block.

gordonhart:
Finished my editing pass. One thing I noted is that we shy away from mentioning TP/FP/FN in most places. Is this intentional? It would be clearer if we drew a stronger link between "matched" and TP, "unmatched inference" with FP, and "unmatched GT" with FN.

docs/metrics/geometry-matcher.md
y27choi: Good point. No specific reason.

gordonhart: I've updated this and other guides to use `inference` instead of `prediction`

y27choi:
> Finished my editing pass. One thing I noted is that we shy away from mentioning TP/FP/FN in most places. Is this intentional? It would be clearer if we drew a stronger link between "matched" and TP, "unmatched inference" with FP, and "unmatched GT" with FN.


TP/FP/FN comes after thresholding which is why I avoided using TP/FP/FN for bounding box matcher.  Any matched pair could be FP and FN depending on the confidence score threshold used, so I didn't want to cause any confusion.  But, I agree that it will make the doc clearer and easier to understand / scan, and I have seen other sources (e.g., open-source codes) referring them as TP/FP/FNs so using them in this context is not completely wrong.

gordonhart:
Makes sense. Do you think it would help to include a section about the relationship between matched/unmatched and TP/FP/FN?

docs/metrics/geometry-matcher.md
y27choi: Thanks. I also created a ticket to review all docs to update `prediction` to `inference`.  I think some images with the word `prediction` need to be updated as well.

y27choi:
> Makes sense. Do you think it would help to include a section about the relationship between matched/unmatched and TP/FP/FN?

I believe that's covered in [this section](http://localhost:8999/metrics/tp-fp-fn-tn/#single-class).  While taking another look at this page, I've noticed that threshold is not mentioned in this table:
image
 
 Will edit this.

y27choi:
Completed the review.  It looks good to me.  I also made an edit on the tp-fp-fn-tn page to differentiate the TP / FP / FN definition with matched/unmatched.

docs/assets/images/metrics-matcher-example1.png
gordonhart, line 1: Noting that these will need to be updated before the Metrics Glossary goes live.

gordonhart (approved):

PR #107, opened 2023-06-26
Single class OD workflow

marko-polo-cheno:
Stems from the multiclass OD workflow


gordonhart (approved):
Looks great, nice tests. There's some cleanup/reorg we should do before promoting this out of `kolena._experimental` but that shouldn't block this PR. Would appreciate if @y27choi could take a look before merging.

Are there results up with this single-class evaluation logic anywhere? As a part of the acceptance process for this pre-built workflow, let's plan to do a thorough review of the end result on the platform.

kolena/_experimental/object_detection/evaluator_single_class.py
```diff
@@ -0,0 +1,233 @@
...
+    Caches matchings per test case for test case metrics and test case plots.
+    """
+
+    def compute_image_metrics(
+        self,
+        ground_truth: GroundTruth,
+        inference: Inference,
+        configuration: ThresholdConfiguration,
+        test_case_name: str,
+        labels: Set[str],  # the label being tested in this test case
```
y27choi: I think we can omit this.

y27choi (commented):
Finished reviewing the code. Will take a look at the tests.

tests/integration/_experimental/object_detection/test_evaluator_single_class_test_case_metrics.py
```diff
@@ -0,0 +1,904 @@
...
+    ): TestCaseMetricsSingleClass(
+        Objects=1,
+        Inferences=1,
+        TP=1,
+        FN=0,
+        FP=0,
+        Precision=1.0,
+        Recall=1.0,
+        F1=1.0,
+        AP=0.07494252873563219,
```
y27choi: Is `AP` correct there?

y27choi: It's weird to me that `AP` changes as the threshold changes (see the next 3 tests)

tests/unit/_experimental/object_detection/test_plots.py
```diff
@@ -1019,41 +1019,13 @@ def test__curve__plots(
     [
         (
             "no confusion, one TP per label",
```
y27choi, line 1021: @marko-polo-cheno do you know what caused these changes in the unit test?

y27choi: nvm, I get it now.

tests/integration/_experimental/object_detection/test_evaluator_single_class_test_case_metrics.py
marko-polo-cheno: Agreed, originally the cache of matchings by test case name would follow through in an evaluator, but it makes more sense for each test to be completely independent.

tests/integration/_experimental/object_detection/test_evaluator_single_class_test_case_metrics.py
```diff
@@ -0,0 +1,910 @@
...
+    ): TestCaseMetricsSingleClass(
+        Objects=1,
+        Inferences=1,
+        TP=1,
+        FN=0,
+        FP=0,
+        Precision=1.0,
+        Recall=1.0,
+        F1=1.0,
+        AP=0.0,
```
y27choi, line 300: Does 0 `AP` make sense here? 🤔

marko-polo-cheno: AP should follow the PR curves if they exist. If a PR curve does not exist, maybe AP should be None instead of 0.0
Will think about this a bit more

marko-polo-cheno: cc @gordonhart
if a PR curve does not exist, would it be better to see an AP of 0, the precision value, or None?
I think None is the best choice here

y27choi: Obviously this is a very rare edge case that I don't expect customers to come across but I personally think that the PR curve should be (0.0, 1.0) (i.e. extending the first precision value all the way to the y-axis)  and (1.0, 1.0) which should also give you `AP` value of 1.  But given that this is a very rare case and unlikely that anyone will stumble upon, I'm ok not changing the logic for it. 

Just to confirm though, let's say there are more than 1 point on the PR curve but recall doesn't exist all the way to 0 (so say we have (0.3, 0.5) and (0.7, 0.4)) then what would be the AP?  Would it be (0.5 * 0.3) + (0.4 * 0.4) = 0.31?

marko-polo-cheno: It would be 0.29 by our sum of rectangular areas starting from (0,0), and precision=0 where recall is not defined (on the right)
```
(0.3 - 0) * 0.7 + (0.5 - 0.3) * 0.4 + (1 - 0.5) * 0
or
(0.4 - 0) * 0.5 + (0.7 - 0.4) * 0.3 + (1 - 0.7) * 0
```

PR #108, opened 2023-06-26
Prebuilt multiclass classification workflow

marko-polo-cheno:


gordonhart:
Does this include changes from #82?

PR #118, opened 2023-06-28
validate model, test case and test suite names are non empty

sandalns:
### Linked issue(s):
Fixes KOL-2451

Enforces that model, test case and test suite names are non empty. 


kolena/_utils/validators.py
```diff
@@ -20,3 +20,8 @@ class ValidatorConfig:
     arbitrary_types_allowed = True
     smart_union = True
     extra = Extra.allow  # do not fail when unrecognized values are provided
+
+
+def validate_not_blank(field: str):
+    if not bool(field and not field.isspace()):
+        raise ValueError("Names must be non empty")
```
gordonhart: The function name doesn't indicate that this is specific to `name` fields, but the exception assumes this is a `name` field. Which is correct?

tests/unit/utils/test_validators.py
```diff
@@ -0,0 +1,24 @@
...
+
+from kolena._utils.validators import validate_not_blank
+
+
+def test__validate__field_name__empty() -> None:
+    with pytest.raises(ValueError):
+        validate_not_blank("")
+
+    with pytest.raises(ValueError):
+        validate_not_blank("  ")
```
gordonhart: Minor, but I'd use a parametrize fixture here and test more whitespace variants

tests/integration/fr/test_test_case.py
```diff
@@ -344,3 +344,8 @@ def test__edit__reset(test_samples: List[TestCaseRecord]) -> None:
     assert test_case.image_count == 3
     assert test_case.pair_count_genuine == 0
     assert test_case.pair_count_imposter == 3
+
+
+def test__empty_test_case_name() -> None:
+    with pytest.raises(ValueError):
+        TestCase("")
```
gordonhart: Please follow the test naming convention — here, this might be `test__init__validate_name`

kolena/workflow/test_case.py
```diff
@@ -93,6 +94,7 @@ def __init__(
     ):
         if type(self) == TestCase:
             raise Exception("<TestCase> must be subclassed.")
+        validate_not_blank(name)
```
gordonhart: Does this prevent creating empty-name test cases via `TestCase.create`?

gordonhart (commented):
I'm tempted to say that this should be rolled into the Pydantic setup, rather than applied as a separate function call, but it seems there are still some rough edges with `@validate_arguments` that block that approach. In light of that, adding a separate function call seems fine.

ly-kolena:
> I'm tempted to say that this should be rolled into the Pydantic setup, rather than applied as a separate function call, but it seems there are still some rough edges with `@validate_arguments` that block that approach. In light of that, adding a separate function call seems fine.

Although nice to do this check also in client sdk as a pre-check, but shouldn't we do this check at backend and raise error from the endpoints?

kolena/_utils/validators.py
sandalns: Updated the error to not mentioned name as this could be used for other fields later on.

tests/integration/fr/test_test_case.py
sandalns: Updating across all tests that were added.

kolena/workflow/test_case.py
sandalns: Added this to TestCase.create as well and added matching tests

kolena/_utils/validators.py
gordonhart: If a user sees the new "Field must be non-empty" exception, how do they know what to fix?

sandalns:
> > I'm tempted to say that this should be rolled into the Pydantic setup, rather than applied as a separate function call, but it seems there are still some rough edges with `@validate_arguments` that block that approach. In light of that, adding a separate function call seems fine.
> 
> Although nice to do this check also in client sdk as a pre-check, but shouldn't we do this check at backend and raise error from the endpoints?

I opted to do this on the client sdk side since it is a very simple check makes sense to be done here. Given we also validate other inputs in the client sdk, it seemed like a good approach.

kolena/_utils/validators.py
sandalns: Updating this to have a clearer error message.

ly-kolena:
> > > I'm tempted to say that this should be rolled into the Pydantic setup, rather than applied as a separate function call, but it seems there are still some rough edges with `@validate_arguments` that block that approach. In light of that, adding a separate function call seems fine.
> > 
> > 
> > Although nice to do this check also in client sdk as a pre-check, but shouldn't we do this check at backend and raise error from the endpoints?
> 
> I opted to do this on the client sdk side since it is a very simple check makes sense to be done here. Given we also validate other inputs in the client sdk, it seemed like a good approach.

In principle, we should not rely on client sdk to protect backend. If this should not be valid in the platform, we should put in server, which in this case should be relatively straightforward. Non-blocking to this PR.

sandalns:
> > > > I'm tempted to say that this should be rolled into the Pydantic setup, rather than applied as a separate function call, but it seems there are still some rough edges with `@validate_arguments` that block that approach. In light of that, adding a separate function call seems fine.
> > > 
> > > 
> > > Although nice to do this check also in client sdk as a pre-check, but shouldn't we do this check at backend and raise error from the endpoints?
> > 
> > 
> > I opted to do this on the client sdk side since it is a very simple check makes sense to be done here. Given we also validate other inputs in the client sdk, it seemed like a good approach.
> 
> In principle, we should not rely on client sdk to protect backend. If this should not be valid in the platform, we should put in server, which in this case should be relatively straightforward. Non-blocking to this PR.

Thought about this more, I agree this should be put in backend regardless. This might still be useful on the client side as well to reduce calls that are invalid, but I can close this in favor of only doing this on the backend.

ly-kolena:
> Thought about this more, I agree this should be put in backend regardless. This might still be useful on the client side as well to reduce calls that are invalid, but I can close this in favor of only doing this on the backend.

Doing client-side check is fine.

kolena/_utils/validators.py
sandalns: Updated to the approach we discussed offline.

tests/unit/utils/test_validators.py
sandalns: Updated

kolena/_utils/consts.py
```diff
@@ -21,3 +21,9 @@ class BatchSize(int, Enum):
 
     LOAD_RECORDS = UPLOAD_RECORDS
     LOAD_SAMPLES = 1_000_000
+
+
+class FieldNamesError(str, Enum):
```
munkyshi: nit: This `Enum` isn't an `Error` -- why not just `FieldNames`?

kolena/_utils/validators.py
```diff
@@ -20,3 +22,9 @@ class ValidatorConfig:
     arbitrary_types_allowed = True
     smart_union = True
     extra = Extra.allow  # do not fail when unrecognized values are provided
+
+
+def validate_name(field: str, field_name: Optional[str] = None):
+    field_name_str = field_name if field_name else "field"
+    if not bool(field and not field.isspace()):
```
munkyshi: Why is this so complicated? Why not just:
```
if field.isspace():
    ...
```

If you're also hoping to validate that `field` is non-None, the `@validate_arguments` would probably work.

sandalns: field.isspace() is False for the empty string "". I simplified this to `if not field or field.isspace():`

kolena/_utils/consts.py
sandalns: Updated to FieldNames

tests/integration/detection/test_model.py
```diff
@@ -45,6 +45,11 @@ def test__init() -> None:
     assert model == Model(name)
 
 
+def test__init__validate_name() -> None:
```
ly-kolena: Nit: this can be in unit-test, same comment for other similar tests.

tests/integration/workflow/test_test_suite.py
```diff
@@ -262,3 +262,15 @@ def test__load_test_samples(
         (test_case_1, sorted([dummy_test_samples[0], dummy_test_samples[1]])),
         (test_case_2, sorted([dummy_test_samples[2], dummy_test_samples[1]])),
     ]
+
+
+def test__init__validate_name() -> None:
```
ly-kolena: Nit: we can move these to unit tests with slight modification by not including testcase.

sandalns: Will do for similar tests as well.

tests/integration/detection/test_model.py
sandalns: Will do for similar tests as well.

.mypy.ini
```diff
@@ -1,4 +1,5 @@
 [mypy]
+plugins = pydantic.mypy
```
gordonhart: The Pydantic plugin is already specified on line 11 of this file

examples/keypoint_detection/pyproject.toml
```diff
@@ -10,6 +10,8 @@ python = ">=3.8,<3.12"
 kolena = ">=0.69.0,<1"
 s3fs = "^2023.5.0"
 fsspec = "^2023.5.0"
+pydantic = ">=1.8,<2.0"
+
```
gordonhart, line 14: Super-nit: extra newline

kolena/detection/_internal/model.py
```diff
@@ -79,6 +81,7 @@ class BaseModel(ABC, Frozen, WithTelemetry):
 
     @validate_arguments(config=ValidatorConfig)
     def __init__(self, name: str, workflow: WorkflowType, metadata: Optional[Dict[str, Any]] = None):
+        validate_name(name, FieldNames.MODEL_NAME)
```
gordonhart: I don't think this works well in Python 3.11 (which we support):

```
$ poetry run python3
Python 3.11.1 (main, Feb 13 2023, 09:21:24) [Clang 13.1.6 (clang-1316.0.21.2.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from kolena.detection import Model
>>> Model("")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/gordonhart/Software/kolena/kolena/_utils/instrumentation.py", line 40, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gordonhart/Software/kolena/kolena/detection/model.py", line 56, in __init__
    super().__init__(name, WorkflowType.DETECTION, metadata)
  File "/Users/gordonhart/Software/kolena/kolena/_utils/instrumentation.py", line 40, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "pydantic/decorator.py", line 40, in pydantic.decorator.validate_arguments.validate.wrapper_function
  File "pydantic/decorator.py", line 134, in pydantic.decorator.ValidatedFunction.call
  File "pydantic/decorator.py", line 206, in pydantic.decorator.ValidatedFunction.execute
  File "/Users/gordonhart/Software/kolena/kolena/detection/_internal/model.py", line 84, in __init__
    validate_name(name, FieldNames.MODEL_NAME)
  File "/Users/gordonhart/Software/kolena/kolena/_utils/validators.py", line 30, in validate_name
    raise ValueError(f"{field_name_str} must be non empty")
ValueError: FieldNames.MODEL_NAME must be non empty
```

Probably want to pass `FieldNames.MODEL_NAME.value` here or have `validate_name` take an `Optional[FieldNames]` and access the value itself.

kolena/_utils/consts.py
```diff
@@ -21,3 +21,9 @@ class BatchSize(int, Enum):
 
     LOAD_RECORDS = UPLOAD_RECORDS
     LOAD_SAMPLES = 1_000_000
+
+
+class FieldNames(str, Enum):
```
gordonhart: Usually a name like this would be singular — see `BatchSize` above

tests/integration/detection/test_model.py
```diff
@@ -45,6 +45,11 @@ def test__init() -> None:
...
+    with pytest.raises(ValueError):
+        Model("")
+
```
gordonhart: For this and other definitions, we should be testing the `create` method as well

tests/integration/workflow/test_test_case.py
```diff
@@ -201,3 +201,8 @@ def test__edit__remove_only(
 
     assert test_case.version == 2
     assert sorted(test_case.load_test_samples()) == sorted(rest)
+
+
+def test__init__validate_name() -> None:
+    with pytest.raises(ValueError):
+        TestCase("")
```
gordonhart, line 208: Should test `create` as well

tests/integration/workflow/test_test_suite.py
```diff
@@ -262,3 +262,9 @@ def test__load_test_samples(
         (test_case_1, sorted([dummy_test_samples[0], dummy_test_samples[1]])),
         (test_case_2, sorted([dummy_test_samples[2], dummy_test_samples[1]])),
     ]
+
+
+def test__create__validate_name() -> None:
+    test_case_1 = TestCase(with_test_prefix(f"{__file__}::test__create__validate_name test case 1"))
+    with pytest.raises(ValueError):
+        TestSuite.create("", test_cases=[test_case_1])
```
gordonhart, line 270: Should test `__init__` as well

tests/integration/detection/test_model.py
gordonhart: Actually, should these integration tests be removed entirely in favor of the unit tests? Looks like the unit tests cover all usages.

tests/integration/workflow/test_test_case.py
sandalns: create is tested in tests/unit/workflow/test_test_case

tests/integration/workflow/test_test_suite.py
sandalns: init is tested in tests/unit/workflow/test_test_suite.

tests/integration/detection/test_model.py
sandalns: I thought I had removed this, will remove integration tests in favor of unit tests.

kolena/_utils/consts.py
sandalns: Will update

kolena/detection/_internal/model.py
sandalns: will update

sandalns: Will update and test with 3.11.

kolena/_utils/consts.py
sandalns: Will update

kolena/detection/_internal/model.py
sandalns: Updated to use `Optional[FieldNames]`.

This is the new error on python 3.11:

```
Python 3.11.3 (main, Jul  4 2023, 22:05:40) [Clang 13.1.6 (clang-1316.0.21.2.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from kolena.detection import Model
>>> Model("")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/niteshsandal/Workspace/kolena/kolena/_utils/instrumentation.py", line 40, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/niteshsandal/Workspace/kolena/kolena/detection/model.py", line 56, in __init__
    super().__init__(name, WorkflowType.DETECTION, metadata)
  File "/Users/niteshsandal/Workspace/kolena/kolena/_utils/instrumentation.py", line 40, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/niteshsandal/Library/Caches/pypoetry/virtualenvs/kolena-1YW62IuT-py3.11/lib/python3.11/site-packages/pydantic/decorator.py", line 40, in wrapper_function
    return vd.call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/niteshsandal/Library/Caches/pypoetry/virtualenvs/kolena-1YW62IuT-py3.11/lib/python3.11/site-packages/pydantic/decorator.py", line 134, in call
    return self.execute(m)
           ^^^^^^^^^^^^^^^
  File "/Users/niteshsandal/Library/Caches/pypoetry/virtualenvs/kolena-1YW62IuT-py3.11/lib/python3.11/site-packages/pydantic/decorator.py", line 201, in execute
    return self.raw_function(**d, **var_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/niteshsandal/Workspace/kolena/kolena/detection/_internal/model.py", line 84, in __init__
    validate_name(name, FieldNames.MODEL_NAME)
  File "/Users/niteshsandal/Workspace/kolena/kolena/_utils/validators.py", line 32, in validate_name
    raise ValueError(f"{field_name_str} must be non empty")
ValueError: Model name must be non empty

```

kolena/_utils/consts.py
sandalns: Done

examples/keypoint_detection/pyproject.toml
sandalns: Removed the new line but it is not being shown correctly in the diff

.mypy.ini
sandalns: Going to remove, this extra line fixes pre-commit issues with mypy for python 3.9 but causes other issues on 3.11. Will remove and find an alternative way that works across all supported python versions.

tests/integration/detection/test_model.py
sandalns: Done, some tests for workflow are kept in integration (test case init) as they require define_worfklow to be called (<TestCase> must be subclassed).

gordonhart:
The Pydantic pin landed in #125 — I'd rebase and revert `pyproject.toml` changes in this PR (the `examples` projects shouldn't need an explicit pin of the transient Pydantic dependency).

.mypy.ini
sandalns: Removed, this will be fixed by pinning in .pre-commit-config.yaml

sandalns:
> The Pydantic pin landed in #125 — I'd rebase and revert `pyproject.toml` changes in this PR (the `examples` projects shouldn't need an explicit pin of the transient Pydantic dependency).

I tried reverting the changes to 'pyproject.toml' for examples, however the tests were failing due to the installation of pydantic 2.0.1 as they use the published client. Adding them back temporarily for this PR.

ly-kolena (approved):

PR #125, opened 2023-07-05
Cap pydantic version to <2.0

marko-polo-cheno:


gordonhart (approved):
@sandalns also uncovered this one on #118 but I agree let's merge this in as a separate change to unblock other dev.

gordonhart:
For posterity — we try to be as unopinionated as possible with dependency version pins to maximize compatibility with other projects. In this particular case, [Pydantic just released a V2 with a few breaking changes](https://docs.pydantic.dev/2.0/migration/) that earns it a stricter pin in this project. We'll pick up the new version once it's had a bit more soak time.

PR #128, opened 2023-07-06
Add prebuilt OD multiclass tests

marko-polo-cheno:
Unit tests and integration tests on 4 configs (2 single, 2 multi)


tests/integration/_experimental/object_detection/test_evaluator_multiclass_test_sample_metrics.py
```diff
@@ -0,0 +1,5873 @@
...
+    ],
+}
+
+
+TEST_CONFIGURATIONS: Dict[str, ThresholdConfiguration] = {
+    "Threshold: Fixed(0.3), IoU: 0.3, confidence ≥ 0.0": ThresholdConfiguration(
+        threshold_strategy=ThresholdStrategy.FIXED_03,
+        iou_threshold=0.3,
+        min_confidence_score=0.0,
+        with_class_level_metrics=False,
```
y27choi: Are these `single_class` tests necessary for multiclass evaluator test?  Just curious how these differ from the one from [the previous PR](https://github.com/kolenaIO/kolena/pull/107)

marko-polo-cheno: I think it is still important to keep since a stratification could introduce a test case where only one class is present. Might be pretty rare, but in my opinion, not rare enough to be unnecessary.

y27choi: But wouldn't it still use `with_class_level_metrics=True` configuration on the single class test case?  This is configuration-based and we apply the configuration per test suite. I think a better way to test this out would be construct a test suite with multiple test case and have some of them be single class only and evaluate it with the `with_class_level_metrics=True` configuration.

tests/integration/_experimental/object_detection/test_evaluator_multiclass_test_case_metrics.py
```diff
@@ -0,0 +1,4535 @@
...
+                Threshold=0.3,
+                Objects=1,
+                Inferences=1,
+                TP=0,
+                FN=1,
+                FP=1,
+                Precision=0.0,
+                Recall=0.0,
+                F1=0.0,
+                AP=0.0,
```
y27choi: This still bothers me.  We know that this is incorrect, and having these all over our unit tests seems just wrong to me.  Not plotting a single point PR curve was our decision but I don't think that should affect the AP calculation. A single point precision and recall has a valid AP and I think we should reflect that in our AP calculation. 

At this point, I would rather prefer to plot a single point PR than to have incorrect AP value. What do you think @marko-polo-cheno?

y27choi:
Have skimmed through all the tests.  Left two comments on:
1. a lot of tests are duplicate of `single_class` tests so we should just omit them and
2. incorrect AP calculation for single value precision and recall.

tests/integration/_experimental/object_detection/test_evaluator_multiclass_test_sample_metrics.py
marko-polo-cheno: Good point with multiple test cases - will add tests for a test suite with multiple test cases.

A test case with multiple labels can still use `with_class_level_metrics=False` because some people might not use the nested test case metrics feature. I don't think there are identical dups of tests because the `TEST_DATA` is different.

tests/integration/_experimental/object_detection/test_evaluator_multiclass_test_case_metrics.py
marko-polo-cheno: We shouldn't plot a PR curve that has a single point because it is noise to the interface. The AP calculation at its core estimates AUC-PR, and it relies on a PR curve.
From internal discussions, we set our AP calculation to include (0, 0) on the PR curve. By this definition, we expect the area to usually be non-zero when there is only a single point, but the AP value can never be 1.0. If we want to see a PR curve with AP = 1 (this [discussion](https://github.com/kolenaIO/kolena/pull/107#discussion_r1253245198)) then we will have to include (recall=0, precision=1) and change our AP calculation util. I don't think that would be a good change.

I know this feels wrong so I suggest that whenever there is no valid PR curve, `AP <-- None` or `AP <-- precision value`. I prefer `None` because a random AP=0.4 is meaningless - lmk what you think!

tests/integration/_experimental/object_detection/test_evaluator_multiclass_test_sample_metrics.py
y27choi: Ok I understand now. I think what got me confused was the different interpretation of `single class` and `multiclass` used in tests vs. evaluator config.  Given that the tests are supposed to test any multiclass usages, these tests are good.

tests/integration/_experimental/object_detection/test_evaluator_multiclass_test_case_metrics.py
y27choi: > we will have to include (recall=0, precision=1) and change our AP calculation util. I don't think that would be a good change

Why don't you think it would be a good change?  As far as I understand, the AP calculation used in Pascal VOC extends the precision value to its left.

kolena/_experimental/object_detection/evaluator_multiclass.py
```diff
@@ -168,11 +167,7 @@ def compute_test_sample_metrics(
         assert configuration is not None, "must specify configuration"
         # compute thresholds to cache values for subsequent steps
         self.compute_and_cache_f1_optimal_thresholds(configuration, inferences)
-        labels = {gt.label for _, gts, _ in inferences for gt in gts.bboxes}
-        return [
-            (ts, self.compute_image_metrics(gt, inf, configuration, test_case.name, labels))
-            for ts, gt, inf in inferences
-        ]
+        return [(ts, self.compute_image_metrics(gt, inf, configuration, test_case.name)) for ts, gt, inf in inferences]
```
y27choi, line 186: What's the reason for dropping the `labels` here?

tests/unit/_experimental/object_detection/test_utils.py
```diff
@@ -13,17 +13,89 @@
...
+            0.5,
+            {"b", "c"},
+            [
+                ScoredLabeledBoundingBox((0, 0), (1, 1), "c", 0.9),
+                ScoredLabeledBoundingBox((0, 0), (1, 1), "b", 0.6),
+            ],
+        ),
+    ],
+)
+def test__filter__inferences(
```
gordonhart: Tiny nit — `test__filter_inferences` (drop an underscore)

tests/integration/_experimental/object_detection/test_evaluator_multiclass_test_case_metrics.py
```diff
@@ -0,0 +1,2086 @@
...
+    TEST_PARAMS,
+)
+def test__prebuilt__object__detection__multiclass__compute__test__case__metrics(
+    config_name: str,
+    test_name: str,
+) -> None:
+    config = TEST_CONFIGURATIONS[config_name]
+    eval = ObjectDetectionEvaluator(configurations=[config])
+    random_test_case_name = with_test_prefix("test_evaluator_multiclass") + str(uuid.uuid4())
+    test_case = TestCase(random_test_case_name, reset=True)
```
gordonhart: Does this need to create a real test case? This round-trip will slow down these tests significantly — recommend refactoring here to use e.g. a mocked test case, or figuring out a way to make these unit tests instead of integration tests.

gordonhart (commented):
Latest test refactor looks like a huge improvement!

One final request: these should really be unit tests, not integration tests. From what I can gather, the only reason they are integration tests currently is to create real `TestCase` objects — unless that class is used heavily during evaluation (doesn't seem like it is), recommend refactoring such that the bulk of the functionality does not rely on the `TestCase`, or mocking out the test case objects such that these tests can run as unit tests.

tests/integration/_experimental/object_detection/test_evaluator_multiclass_test_case_plots.py
```diff
@@ -0,0 +1,477 @@
...
+                                0.5377358490566038,
+                                0.4716981132075472,
+                                0.42452830188679247,
+                                0.37735849056603776,
+                                0.29245283018867924,
+                                0.27358490566037735,
+                                0.11320754716981132,
+                                0.09433962264150944,
+                                0.0660377358490566,
+                                0.0,
```
gordonhart: No changes requested here, but noting that these tests are going to be extremely brittle — usually we like to avoid using copy/pasting actual output as expected output, in favor of using hand-written, smaller tests that are easier to understand. What are these different parametrized cases testing? It isn't clear.

tests/integration/_experimental/object_detection/test_evaluator_multiclass_test_case_plots.py
```diff
@@ -0,0 +1,477 @@
...
+        configuration=config,
+    )
+
+    result = eval.compute_test_case_plots(
+        test_case=test_case,
+        inferences=[],
+        metrics=[],
+        configuration=config,
+    )
+    assert expected == result
```
gordonhart: This direct equality is likely vulnerable to floating point flakiness. I'm surprised that these tests run successfully across platforms

tests/integration/_experimental/object_detection/test_evaluator_multiclass_test_case_plots.py
```diff
@@ -0,0 +1,477 @@
...
+    eval.compute_test_sample_metrics(
+        test_case=test_case,
+        inferences=[item for data in TEST_DATA.values() for item in data],
+        configuration=config,
+    )
+
+    result = eval.compute_test_case_plots(
+        test_case=test_case,
+        inferences=[],
+        metrics=[],
```
gordonhart: Should these be populated?

tests/integration/_experimental/object_detection/test_evaluator_multiclass_test_case_plots.py
```diff
@@ -0,0 +1,477 @@
...
+                    labels=["a", "b", "c", "d", "e"],
+                    matrix=[[29, 2, 0, 0, 0], [0, 10, 0, 0, 0], [0, 0, 18, 0, 0], [0, 0, 0, 2, 0], [0, 1, 0, 0, 4]],
+                    x_label="Predicted",
+                    y_label="Actual",
+                ),
+            ],
+        ),
+    ],
+)
+def test__prebuilt__object__detection__multiclass__compute__test__case__plots__advanced(
```
gordonhart: The intention with the double underscores isn't to double-underscore everything, but to distinguish between different "parts" of the title, where within a "part," different words are separated by a single underscore. This name is already a bit of a mouthful, but I'd update it to something like:

```python
def test__object_detection__multiclass__compute_test_case_plots__advanced(...):
```

tests/integration/_experimental/object_detection/test_evaluator_multiclass_test_sample_metrics.py
```diff
@@ -0,0 +1,2672 @@
...
+        inferences=TEST_DATA[test_name],
+        configuration=config,
+    )
+    print(config_name, test_name)
+    print(result)
+    assert EXPECTED_COMPUTE_TEST_SAMPLE_METRICS[config_name, test_name] == result
+
+
+def test__prebuilt__object__detection__multiclass__compute__test__sample__metrics__all() -> None:
+    config_name = "Threshold: Fixed(0.5) by class, IoU: 0.5, confidence ≥ 0.0"
```
gordonhart: These names are going to be quite brittle. Is there a reason we're not just using the config object itself, rather than its stringified display name?

kolena/_experimental/object_detection/utils.py
```diff
@@ -79,18 +94,25 @@ def _compute_threshold_curve(
             zero_division=0,
         )
 
-        precisions.append(precision)
-        recalls.append(recall)
+        # avoid curves with one x-value and two 2-values
```
y27choi: Wouldn't this affect the AP calculation?

y27choi: nit: comment is unclear.  Should it be two y-values?

kolena/_experimental/object_detection/workflow.py
```diff
@@ -174,12 +174,13 @@ def display_name(self) -> str:
 class ThresholdConfiguration(EvaluatorConfiguration):
     threshold_strategy: ThresholdStrategy
     iou_threshold: float
-    min_confidence_score: float
+    min_confidence_score: Optional[float]
```
y27choi: `Optional[float] = 0`?

y27choi: Then we probably don't need the changes made in `display_name()` below.

tests/integration/_experimental/object_detection/test_evaluator_multiclass_test_case_metrics.py
```diff
@@ -0,0 +1,2086 @@
...
+        inferences=[ts_gt_inf for _, data in TEST_DATA.items() for ts_gt_inf in data],
+        metrics=[
+            metric
+            for config_test, metrics in EXPECTED_COMPUTE_TEST_SAMPLE_METRICS.items()
+            for _, metric in metrics
+            if config_test[0] == config_name
+        ],
+        configuration=config,
+    )
+    print(config_name)
```
y27choi: nit: remove prints

tests/integration/_experimental/object_detection/test_evaluator_multiclass_test_suite_metrics.py
```diff
@@ -0,0 +1,320 @@
...
+                        ],
+                        Objects=0,
+                        Inferences=0,
+                        TP=0,
+                        FN=0,
+                        FP=0,
+                        macro_Precision=0,
+                        macro_Recall=0,
+                        macro_F1=0,
+                        mean_AP=0.123,
```
y27choi: Hmm this mAP value doesn't make sense to me. I thought `mean_AP` is calculated by taking average of `AP` across all labels, so in this case shouldn't it be 1? Or did I miss something?

tests/integration/_experimental/object_detection/test_evaluator_multiclass_test_suite_metrics.py
```diff
@@ -0,0 +1,320 @@
...
+            TestCaseMetricsSingleClass(
+                Objects=0,
+                Inferences=0,
+                TP=0,
+                FN=0,
+                FP=0,
+                Precision=0.0,
+                Recall=0.0,
+                F1=0.0,
+                AP=0.3,
```
y27choi: Can you have non-zero AP value when there is 0 objects and 0 inferences? 🤔

y27choi: Or is this just a made up number to check for `TestSuiteMetric` computation?

tests/unit/_experimental/object_detection/test_plots.py
```diff
@@ -513,6 +511,7 @@ def test__none__curve__plot(
     from kolena._experimental.object_detection.utils import compute_pr_plot
     from kolena._experimental.object_detection.utils import compute_pr_curve
 
+    print(test_name)
```
y27choi: nit: remove print

y27choi (commented):
Finished reviewing the changes and left some comments.

tests/integration/_experimental/object_detection/test_evaluator_multiclass_test_case_metrics.py
marko-polo-cheno: Resolved via default average precision method and PR curve fix

kolena/_experimental/object_detection/evaluator_multiclass.py
marko-polo-cheno: Discussed internally - to ignore nothing from the inferences

tests/integration/_experimental/object_detection/test_evaluator_multiclass_test_case_plots.py
marko-polo-cheno: I hear you! My guess as to why tests do not flake are because at the end of the day, they are fractions `eg. 63/106, 57/106` with a hint of luck.

tests/integration/_experimental/object_detection/test_evaluator_multiclass_test_case_plots.py
marko-polo-cheno: Might be outdated, but no because the plots rely on caches and don't use those fields.

kolena/_experimental/object_detection/utils.py
marko-polo-cheno: Good point - I've changed the logic to only consider the largest precision for each recall. This is keep a consistent AP calculation.

tests/integration/_experimental/object_detection/test_evaluator_multiclass_test_suite_metrics.py
marko-polo-cheno: Outdated, but yes the `mean_AP` was a dummy number to just test the behaviour of the function - here it averages the APs across test cases.

.circleci/config.yml
```diff
@@ -20,6 +20,8 @@ jobs:
           mapping: |
             kolena/classification/.* classification true
             tests/integration/classification/.* classification true
+            kolena/_experimental/.* _experimental true
```
ly-kolena: Should we just use `workflow` as this is built on top of generic workflow codes, as a result, when generic workflow codes are updated, `_experimental` tests should also be run?

kolena/_experimental/object_detection/evaluator_multiclass.py
```diff
@@ -102,11 +92,13 @@ def compute_image_metrics(
         ]
         non_ignored_inferences = tp + fp
         scores = [inf.score for inf in non_ignored_inferences]
-        image_labels = {gt.label for gt in ground_truth.bboxes}
+        inference_labels = {inf.label for _, inf in bbox_matches.matched + bbox_matches.unmatched_gt if inf} | {
```
y27choi: We can omit `unmatched_gt` because ` unmatched_inf` includes all the confused matches as well.

kolena/_experimental/object_detection/evaluator_multiclass.py
```diff
@@ -195,52 +210,93 @@ def compute_aggregate_label_metrics(
...
+        self,
+        matchings: List[MulticlassInferenceMatches],
+        label: str,
+        thresholds: Dict[str, float],
+    ) -> ClassMetricsPerTestCase:
+        match_and_count: Tuple[MulticlassInferenceMatches, int] = self.bbox_matches_for_one_label(matchings, label)
+        class_matches: MulticlassInferenceMatches = match_and_count[0]
+        samples_count: int = match_and_count[1]
+
+        average_precision: Optional[float] = None
```
y27choi: Is there a reason for computing `average_precision` outside the `class_metrics_per_test_case`?

kolena/_experimental/object_detection/utils.py
```diff
@@ -79,18 +94,29 @@ def _compute_threshold_curve(
             zero_division=0,
         )
 
-        precisions.append(precision)
-        recalls.append(recall)
+        # avoid curves with one x-value and two y-values
+        if recall in recalls:
+            idx = recalls.index(recall)
+            precisions[idx] = max(precisions[idx], precision)
+            recalls[idx] = max(recalls[idx], recall)
```
y27choi: I think ` recalls[idx] = max(recalls[idx], recall)` can be omitted because `recalls[idx] == recall`

kolena/_experimental/object_detection/utils.py
```diff
@@ -79,18 +94,29 @@ def _compute_threshold_curve(
             zero_division=0,
         )
 
-        precisions.append(precision)
-        recalls.append(recall)
+        # avoid curves with one x-value and two y-values
```
y27choi, line 97: I'm curious to know why this is necessary.  Wouldn't the plot still look the same?

y27choi (commented):
Looked at all the non-test code changes. I left some minor comments. Will resume the review!

kolena/_experimental/object_detection/utils.py
y27choi: NVM I saw your comment above.

tests/integration/_experimental/object_detection/test_workflow.py
```diff
@@ -15,22 +15,24 @@
...
-from kolena._experimental.object_detection import TestCase
-from kolena._experimental.object_detection import TestSample
-from kolena._experimental.object_detection import TestSuite
 from kolena.workflow import test
 from kolena.workflow.annotation import LabeledBoundingBox
 from kolena.workflow.annotation import ScoredLabeledBoundingBox
 from tests.integration.helper import fake_locator
 from tests.integration.helper import with_test_prefix
 
+object_detection = pytest.importorskip("kolena._experimental.object_detection")
```
y27choi: In which case are we expecting these imports to fail?

tests/unit/_experimental/object_detection/test_plots.py
```diff
@@ -271,12 +296,12 @@
         MulticlassInferenceMatches(
             matched=[
                 (
-                    LabeledBoundingBox(top_left=(1.0, 1.0), bottom_right=(6.0, 6.0), label="cow"),
-                    ScoredLabeledBoundingBox((1.0, 1.0), (6.0, 6.0), "cow", 0.9),
+                    LabeledBoundingBox(top_left=(1, 1.0), bottom_right=(6, 6.0), label="cow"),
```
y27choi: why did the first element in the tuple change to int?

tests/unit/_experimental/object_detection/test_plots.py
```diff
@@ -1020,21 +1023,48 @@ def test__curve__plots(
...
+    test_name: str,
+    f1_curve: CurvePlot,
+    pr_curve: CurvePlot,
+) -> None:
+    from kolena._experimental.object_detection.utils import compute_f1_plot_multiclass
+    from kolena._experimental.object_detection.utils import compute_pr_plot_multiclass
+
+    f1: CurvePlot = compute_f1_plot_multiclass(all_matches=TEST_MATCHING[test_name])
+    pr: CurvePlot = compute_pr_plot_multiclass(all_matches=TEST_MATCHING[test_name])
+    assert f1 == f1_curve
```
y27choi, line 1052:
```
assert_curve_plot_equals_expected(f1, f1_curve)
assert_curve_plot_equals_expected(pr, pr_curve)
```

tests/unit/_experimental/object_detection/test_evaluators.py
```diff
@@ -0,0 +1,1420 @@
...
+    expected: TestSampleMetrics,
+) -> None:
+    from kolena._experimental.object_detection.evaluator_multiclass import MulticlassObjectDetectionEvaluator
+
+    od_multi = MulticlassObjectDetectionEvaluator()
+    result = od_multi.test_sample_metrics(
+        bbox_matches=bbox_matches,
+        thresholds=thresholds,
+    )
+    print("multi", test_name, result)
```
y27choi: remove delete

tests/unit/_experimental/object_detection/test_evaluators.py
```diff
@@ -0,0 +1,1420 @@
...
+    expected: TestSampleMetricsSingleClass,
+) -> None:
+    from kolena._experimental.object_detection.evaluator_single_class import SingleClassObjectDetectionEvaluator
+
+    od_single = SingleClassObjectDetectionEvaluator()
+    result = od_single.test_sample_metrics_single_class(
+        bbox_matches=bbox_matches,
+        thresholds=thresholds,
+    )
+    print("single", test_name, result)
```
y27choi: remove print

kolena/_experimental/object_detection/utils.py
marko-polo-cheno: PR calculations might produce a higher or lower `precision` for the same `recall`. Here, we take the max value to ensure that the highest `precision` of each `recall` is chosen, and not added.

kolena/_experimental/object_detection/evaluator_multiclass.py
marko-polo-cheno: Mainly for bare bones unit testing coverage and cleanliness (avoids including `TestCase` in unit tests, and the evaluator caches).

tests/integration/_experimental/object_detection/test_evaluator_single_class_f1_optimal.py
```diff
@@ -0,0 +1,291 @@
...
+            ),
+        ],
+        x_config=None,
+        y_config=None,
+    ),
+]
+
+
+@pytest.mark.metrics
+def test__object_detection__multiclass_evaluator__fixed() -> None:
```
y27choi: should the name be `__f1_optimal`?

tests/integration/_experimental/object_detection/test_evaluator_single_class_fixed.py
```diff
@@ -0,0 +1,506 @@
...
+
+TEST_DATA: List[Tuple[TestSample, GroundTruth, Inference]] = [
+    (
+        TestSample(locator=fake_locator(112, "OD")),
+        GroundTruth(
+            bboxes=[
+                LabeledBoundingBox((1, 1), (2, 2), "a"),
+                LabeledBoundingBox((3, 3), (4, 4), "a"),
+                LabeledBoundingBox((5, 5), (6, 6), "a"),
+                LabeledBoundingBox((7, 7), (8, 8), "d"),
```
y27choi: should this be class "a"?

tests/integration/_experimental/object_detection/test_evaluator_single_class_f1_optimal.py
```diff
@@ -0,0 +1,291 @@
...
+
+EXPECTED_COMPUTE_TEST_SAMPLE_METRICS: List[Tuple[TestSample, TestSampleMetricsSingleClass]] = [
+    (
+        TestSample(locator=fake_locator(112, "OD"), metadata={}),
+        TestSampleMetricsSingleClass(
+            TP=[
+                ScoredLabeledBoundingBox((1.0, 1.0), (2.0, 2.0), "a", 1.0),
+                ScoredLabeledBoundingBox((3.0, 3.0), (4.0, 4.0), "a", 0.9),
+                ScoredLabeledBoundingBox((5.0, 5.0), (6.0, 6.0), "a", 0.8),
+                ScoredLabeledBoundingBox((7.0, 7.0), (8.0, 8.0), "d", 0.7),
```
y27choi, line 51: should this be "a"?

tests/integration/_experimental/object_detection/test_evaluator_multiclass_fixed.py
y27choi, line 1: Should we add a test case where multiclass evaluated as single class (i.e., `with_class_level_metrics=False`)?

y27choi (commented):
Finished reviewing tests. Please see the comments I left.

kolena/_experimental/object_detection/utils.py
y27choi: Right. That makes sense for `precision` but does it make sense to take the max for `recall`? 

We are taking the index of `recall` so isn't `recalls[idx] == recall` always true?

tests/integration/_experimental/object_detection/test_workflow.py
marko-polo-cheno: This is dependant on the CI environment having the extra `kolena[metrics]`. If it doesn't have it, imported functions fail when seeing `sklearn_metrics.function`.

tests/unit/_experimental/object_detection/test_plots.py
marko-polo-cheno: I replaced all the `.0,` with `,` for cleaner numbers, and didn't do it for `.0)`.

tests/unit/_experimental/object_detection/test_plots.py
marko-polo-cheno: In this case I want `==` for the `None`s

tests/integration/_experimental/object_detection/test_evaluator_single_class_fixed.py
marko-polo-cheno: This test checks that more than one label is allowed. In the case where a user puts multiple labels into the single class workflow, it will not differentiate labels for matching.

tests/integration/_experimental/object_detection/test_evaluator_single_class_f1_optimal.py
marko-polo-cheno: above

tests/integration/_experimental/object_detection/test_evaluator_multiclass_fixed.py
marko-polo-cheno: above

tests/integration/_experimental/object_detection/test_evaluator_single_class_fixed.py
y27choi: Gotcha. Since the rest of the test cases had the same label, it wasn't too clear. Can we leave a comment mentioning the intention behind using different labels here?

tests/unit/_experimental/object_detection/test_plots.py
y27choi: Should we then just check for `f1 == None` and `pr == None` for clarity?

tests/integration/_experimental/object_detection/test_workflow.py
y27choi: Ah I see. Thanks for the explanation.

tests/integration/_experimental/object_detection/test_evaluator_single_class_fixed.py
marko-polo-cheno: Yeah, that would definitely be good if we come back to these tests in the future!

tests/unit/_experimental/object_detection/test_plots.py
marko-polo-cheno: One of the plots isnt `None`, but the numbers are very clean, so I'd prefer leaving it as is here

y27choi: Then let's rename the test function name.  It makes you think that the return types are all none.

tests/integration/_experimental/object_detection/test_workflow.py
y27choi: If our CI test environment doesn't have `kolena[metrics]` installed, will it silently skip the test?
Curious if there is a better workaround that doesn't silently skip tests. CC @munkyshi

y27choi (approved):
It looks good to me. Just waiting for @munkyshi's comment on the `importorskip` but otherwise 👍

munkyshi:
> It looks good to me. Just waiting for @munkyshi's comment on the `importorskip` but otherwise 👍

Hmmm, it's a bit unclean / looks like it'll be hard to maintain. It looks like all of the tests are under `tests/*/_experimental/object_detection/`. Is there any way we can declare this once (or twice, maybe in a `conftest.py`?), rather than in every file?

I'm also not sure if it's so bad for these tests to fail if they don't have `[metrics]` installed -- these are only being run by developers and the CI tool. We should fail if the CI doesn't have the right dependencies installed, and devs can be explicit about which tests they want to run.

munkyshi:
What's the failure mode? I'd rather set things up to fail with a "you need `kolena[metrics]` installed to run these tests" type error, rather than silently passing.

ly-kolena:
> > It looks good to me. Just waiting for @munkyshi's comment on the `importorskip` but otherwise 👍
> 
> Hmmm, it's a bit unclean / looks like it'll be hard to maintain. It looks like all of the tests are under `tests/*/_experimental/object_detection/`. Is there any way we can declare this once (or twice, maybe in a `conftest.py`?), rather than in every file?
> 
> I'm also not sure if it's so bad for these tests to fail if they don't have `[metrics]` installed -- these are only being run by developers and the CI tool. We should fail if the CI doesn't have the right dependencies installed, and devs can be explicit about which tests they want to run.

They are intentionally skipped when `[metrics]` are not installed, because those are not the context the tests should be run. We only want to run the tests when metrics are installed, and we have a right job to tests them.

munkyshi:
> They are intentionally skipped when [metrics] are not installed, because those are not the context the tests should be run. We only want to run the tests when metrics are installed, and we have a right job to tests them.

If I accidentally change CI to not install `[metrics]`, does this mean that I will now silently skip tests that might have otherwise been relevant to my code change? My gut instinct is that people/code should explicitly opt into skipping tests, rather than magically opting out. Maybe I'm wrong -- does other python code that depends on extras test in the same way?

ly-kolena:
> > They are intentionally skipped when [metrics] are not installed, because those are not the context the tests should be run. We only want to run the tests when metrics are installed, and we have a right job to tests them.
> 
> If I accidentally change CI to not install `[metrics]`, does this mean that I will now silently skip tests that might have otherwise been relevant to my code change? My gut instinct is that people/code should explicitly opt into skipping tests, rather than magically opting out. Maybe I'm wrong -- does other python code that depends on extras test in the same way?

Any incorrect CI change could lead to tests being missed. There could be other ways to track which tests to run under which context, but you could always missed something. (Not discouraging discussing other approaches).

On another note, weights and bias does use `importorskip` in their [unit tests](https://github.com/wandb/wandb/blob/main/tests/pytest_tests/unit_tests/test_plots.py#L41).

tests/integration/_experimental/object_detection/test_evaluator_single_class_f1_optimal.py
```diff
@@ -0,0 +1,292 @@
...
+    assert_test_case_metrics_equals_expected,
+)
+from tests.integration._experimental.object_detection.test_evaluator_single_class_fixed import (
+    assert_test_case_plots_equals_expected,
+)
+from tests.integration._experimental.object_detection.test_evaluator_single_class_fixed import TEST_CASE
+from tests.integration._experimental.object_detection.test_evaluator_single_class_fixed import TEST_DATA
+from tests.integration.helper import fake_locator
+
+object_detection = pytest.importorskip("kolena._experimental.object_detection")
```
ly-kolena: On small thing we can add is put in `reason=xxx` so if developers are working locally and the tests are skipped, they will be more informed.

munkyshi:
@ly-kolena thoughts on something like:
```
if os.getenv("CI"):
    import kolena._experimental.object_detection as object_detection
else:
    object_detection = pytest.importorskip("kolena._experimental.object_detection")
```
based on this [thread](https://github.com/cta-observatory/ctapipe/issues/675#issuecomment-446263254)

marko-polo-cheno:
> ```
> if os.getenv("CI"):
>     import kolena._experimental.object_detection as object_detection
> else:
>     object_detection = pytest.importorskip("kolena._experimental.object_detection")
> ```

Why do we want to force all tests? `import kolena._experimental.object_detection as object_detection` will fail without `kolena[metrics]`.
Or am I misunderstanding? We want `import` when we have `metrics`, and `importorskip` otherwise?

munkyshi:
Synced with @marko-polo-cheno -- I'm still concerned with `pytest.importorskip` (feels like an anti-pattern for good testing hygeine), but I'd rather not introduce too much friction here (and it's unlikely for us to break CI imports in the way we described).

✅

kolena/_experimental/object_detection/evaluator_single_class.py
```diff
@@ -163,11 +167,8 @@ def compute_test_case_metrics(
         recall = compute_recall(tp_count, fn_count)
         f1_score = compute_f1_score(tp_count, fp_count, fn_count)
 
-        average_precision = 0.0
-        if precision > 0:
-            baseline_pr_curve = compute_pr_curve(all_bbox_matches)
-            if baseline_pr_curve is not None:
-                average_precision = compute_average_precision(baseline_pr_curve.y, baseline_pr_curve.x)
+        if average_precision is None:
+            average_precision = compute_average_precision([precision], [recall])
```
munkyshi: I'm not sure if I follow this change -- why are we attempting to compute AP with a single `([precision], [recall])` point? Is this a previously missed edge case?

munkyshi: I'm not sure if I follow this change -- why are we attempting to compute AP with a single `([precision], [recall])` point? Is this a previously missed edge case?

munkyshi: It's also a bit of smell to be attempting to compute it in two steps that are in different functional steps (`test_case_metrics_single_class` and `compute_test_case_metrics`) -- makes the computation very hard to follow

kolena/_experimental/object_detection/utils.py
```diff
@@ -79,18 +94,28 @@ def _compute_threshold_curve(
...
         return None
 
     if curve_type == "f1":
-        return Curve(x=thresholds, y=f1s, label=curve_label)
+        return Curve(x=thresholds, y=f1s, label=curve_label) if len(f1s) >= 2 else None
     else:
-        return Curve(x=recalls, y=precisions, label=curve_label)
+        # add a point to start the PR curve on the vertical axis if needed
+        if 0.0 not in recalls:
+            minpos = recalls.index(min(recalls))
```
munkyshi, line 115: This fails if `recalls` is `[]` -- is it guaranteed that there's at least one value?

y27choi: check for `recall` length on line 107

kolena/_experimental/object_detection/evaluator_single_class.py
y27choi: >I'm not sure if I follow this change -- why are we attempting to compute AP with a single ([precision], [recall]) point? Is this a previously missed edge case?

Removed this case.  With the recent change in `_compute_threshold_curves` handling a single datapoint case, `None` `average_precision` case check is unnecessary. Adjusted some unit tests to accommodate this change.

y27choi: >It's also a bit of smell to be attempting to compute it in two steps that are in different functional steps (test_case_metrics_single_class and compute_test_case_metrics) -- makes the computation very hard to follow

@marko-polo-cheno can take a look at this comment once he is back.

munkyshi (approved):

kolena/_experimental/object_detection/evaluator_single_class.py
munkyshi: Your latest change also gets rid of the second computation, which sorts out the code-smell 👍

marko-polo-cheno: I computed AP in 2 steps:
  1. from the PR curve, get the AP from the curve if it exists
  2. if the PR curve doesn't exist (e.g. only one point), we calculate AP with that single `([precision], [recall])` point
This complication only exists to handle the one-point PR curve case, which is hardly ever seen.

PR #129, opened 2023-07-06
FAQ for pydantic breakage

ly-kolena:

### DIFF
--- a/docs/faq/index.md
+++ b/docs/faq/index.md
@@ -94,10 +94,10 @@ If you don't see your question here, please reach out to us on Slack or at
     Please include any relevant stacktrace or platform URL when reporting an issue.
 
 
-## Troubleshooting errors
+## Troubleshooting Errors
 
-??? faq "I'm seeing error like AttributeError: type object XXX has no attribute '\_\_pydantic_model__'"
+??? faq "I'm seeing the error: AttributeError: type object XXX has no attribute '\_\_pydantic_model__'"
 
-    If you see error message lik this, please check that `pydantic < 2.0` is installed. Currently Kolena is not
-    compatible with pydantic V2. For more information, refer to
+    If you see error message like this, please check that `pydantic < 2.0` is installed. Kolena is not currently
+    compatible with `pydantic` V2. For more information, refer to
     [`kolena` Installation](../installing-kolena/#installation).
--- a/docs/faq/index.md
+++ b/docs/faq/index.md
@@ -92,3 +92,11 @@ If you don't see your question here, please reach out to us on Slack or at
     issues.
 
     Please include any relevant stacktrace or platform URL when reporting an issue.
+
+
+## Troubleshooting errors
+
+??? faq "I'm seeing error like AttributeError: type object 'SingleProcessResponse' has no attribute '__pydantic_model__'"
+
+    If you see error message lik this, please check that `pydantic < 2.0` is installed. Currently Kolena is not
+    compatible with pydantic V2. For more information, refer to [`kolena` Installation](../installing-kolena.md).
--- a/docs/faq/index.md
+++ b/docs/faq/index.md
@@ -96,7 +96,7 @@ If you don't see your question here, please reach out to us on Slack or at
 
 ## Troubleshooting errors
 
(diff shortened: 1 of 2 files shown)


docs/faq/index.md
```diff
@@ -92,3 +92,12 @@ If you don't see your question here, please reach out to us on Slack or at
     issues.
 
     Please include any relevant stacktrace or platform URL when reporting an issue.
+
+
+## Troubleshooting errors
+
+??? faq "I'm seeing error like AttributeError: type object XXX has no attribute '\_\_pydantic_model__'"
+
+    If you see error message lik this, please check that `pydantic < 2.0` is installed. Currently Kolena is not
```
munkyshi: sp: `like`

munkyshi: nit: `Kolena is not currently` flows better.

docs/faq/index.md
```diff
@@ -92,3 +92,12 @@ If you don't see your question here, please reach out to us on Slack or at
     issues.
 
     Please include any relevant stacktrace or platform URL when reporting an issue.
+
+
+## Troubleshooting errors
```
munkyshi: nit: Would uppercase both words in the title.

docs/faq/index.md
```diff
@@ -92,3 +92,12 @@ If you don't see your question here, please reach out to us on Slack or at
...
+
+??? faq "I'm seeing error like AttributeError: type object XXX has no attribute '\_\_pydantic_model__'"
```
munkyshi: nit: I'd explicitly state
```
"I'm seeing the error: AttributeError: type object XXX has no attribute '\_\_pydantic_model__'"
```
to explicitly call out that specific error.

docs/installing-kolena.md
```diff
@@ -29,6 +29,19 @@ or [Poetry](https://python-poetry.org/):
     poetry add kolena
     ```
 
+!!! note
+    Kolena uses pydantic for data validation. Kolena client only works with pydantic V1. Pydantic introduced
+    breaking changes in V2 which would cause problem. Before `0.76.0`, Kolena client did not set an
+    upper bound on compatible pydantic version, therefore a fresh installation of earlier versions could pull down
+    incompatible pydantic `2.0.x` alongside, and this would lead to error like
```
munkyshi: I'd structure the wording a bit differently:

```
Kolena uses `pydantic` for data validation, but is compatible only with pydantic V1. Prior to `v0.76.0`, the `kolena` client did not set an upper bound on compatible `pydantic` version, and as a result, fresh installations of older versions of `kolena` may install an incompatible version of `pydantic`, resulting in errors such as:
```

docs/installing-kolena.md
```diff
@@ -29,6 +29,19 @@ or [Poetry](https://python-poetry.org/):
...
+
+    ```
+    AttributeError: type object 'SingleProcessResponse' has no attribute '__pydantic_model__'
+    ```
+
+    If you see error above, please check pydantic version and make sure V1 is installed.
```
munkyshi:
```
If you see the error above, please check your `pydantic` version and make sure that V1 is installed.
```

munkyshi (commented):
Some suggestions for wording, but thanks for putting this together.

munkyshi (approved):

gordonhart:
Thanks for getting this up.

Should we update our strategy to pin major versions of packages? Annoying for installs to just break like this.

docs/installing-kolena.md
gordonhart: The real resolution here is to install the latest version of `kolena`. Would recommend updating this line to reflect that as it is a much more straightforward solution than "pin a transitive dependency"

ly-kolena:
> Thanks for getting this up.
> 
> Should we update our strategy to pin major versions of packages? Annoying for installs to just break like this.

Agree with that. For this specific issue, Mark's PR already set that in a different syntax. Do you have other concerns?

gordonhart:
> Do you have other concerns?

We are still not pinning major version on: `pandera`, `dacite`, `pyarrow`, `scikit-learn`. Wondering specifically if we should add upper bounds for those too

PR #138, opened 2023-07-12
Prebuilt Classification Utils

marko-polo-cheno:
To add here after init review: test case histograms/barplots/ROC/confusion matrix.

### DIFF
--- a/kolena/_experimental/classification/workflow.py
+++ b/kolena/_experimental/classification/workflow.py
@@ -134,7 +134,7 @@ class ClassMetricsPerTestCase(MetricsTestCase):
     n_incorrect: int
     """Number of incorrect predictions for this class within the test case."""
 
-    accuracy: float
+    Accuracy: float
     """Accuracy for predictions for this class within the test case."""
 
     Precision: float
@@ -188,7 +188,7 @@ class TestCaseMetrics(MetricsTestCase):
     n_incorrect: int
     """Total number of incorrect predictions within this test case."""
 
-    macro_accuracy: float
+    macro_Accuracy: float
     """Macro-averaged accuracy for predictions within this test case."""
 
     macro_Precision: float
--- a/kolena/_experimental/classification/workflow.py
+++ b/kolena/_experimental/classification/workflow.py
@@ -1,0 +1,247 @@
+# Copyright 2021-2023 Kolena Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


kolena/_experimental/classification/workflow.py
```diff
@@ -0,0 +1,247 @@
...
+
+    metadata: Metadata = dataclasses.field(default_factory=dict)
+    """The metadata associated with the test sample."""
+
+
+@dataclass(frozen=True)
+class GroundTruth(BaseGroundTruth):
+    """Ground truth type for the pre-built Classification workflow."""
+
+    classification: ClassificationLabel
```
y27choi: We should think about whether we want to support a multi-label multiclass classification.
And for a binary classification, will this `classification` have to be something like this, (e.g., "dog" and "not dog")?

kolena/_experimental/classification/workflow.py
```diff
@@ -0,0 +1,247 @@
...
+    label: str
+    """Label associated with this class."""
+
+    n_correct: int
+    """Number of correct predictions for this class within the test case."""
+
+    n_incorrect: int
+    """Number of incorrect predictions for this class within the test case."""
+
+    accuracy: float
```
y27choi: `Accuracy` to be consistent?

kolena/_experimental/classification/workflow.py
```diff
@@ -0,0 +1,247 @@
...
+    n_labels: int
+    """Total number of classes within this test case."""
+
+    n_correct: int
+    """Total number of correct predictions within this test case."""
+
+    n_incorrect: int
+    """Total number of incorrect predictions within this test case."""
+
+    macro_accuracy: float
```
y27choi: `macro_Accuracy`?

tests/unit/_experimental/classification/test_utils.py
```diff
@@ -0,0 +1,180 @@
...
+        ),
+    ],
+)
+def test__get__label__confidence(
+    test_name: str,
+    label: str,
+    inference_labels: List[ScoredClassificationLabel],
+    expected: float,
+) -> None:
+    from kolena._experimental.classification.utils import get_label_confidence
```
y27choi: Curious why you are importing within the scope of this function instead of importing at the top

y27choi (approved):
Left some minor comments but overall it looks good. Thanks for keeping the PR short.

kolena/_experimental/classification/workflow.py
marko-polo-cheno: `multi-label` should probably be a separate workflow.
For binary classification, users should be able to have the positive class be an optional `ClassificationLabel` which should be `None` for a negative class.

tests/unit/_experimental/classification/test_utils.py
marko-polo-cheno: `importorskip`, which I didn't know of before, is a workaround to the issue of module imports that rely on the sklearn extra. There's no way to `@pytest.mark.metrics` an import, so in the past, my workaround was to put it inside of each test.

y27choi:
Fixed in https://github.com/kolenaIO/kolena/pull/155

PR #141, opened 2023-07-13
Create 2D object detection example

marko-polo-cheno:
Used COCO transportation images.


y27choi:
Could you add a link to an example result page that you used to test?

marko-polo-cheno:
> Could you add a link to an example result page that you used to test?

sure! https://trunk.kolena.io/yoohee/testing?testSuiteId=1320

examples/object_detection_2d/object_detection_2d/constants.py
```diff
@@ -0,0 +1,124 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Dict
+
+DATASET = "coco-2014-val"
+WORKFLOW = "Object Detection"
+
+S3_IMAGE_LOCATION = "s3://kolena-public-datasets/coco-2014-val/imgs/"
+S3_BUCKET = "kolena-public-datasets"
+S3_ANNOTATION_FILE_PATH = "coco-2014-val/meta/instances_val2014.json"
```
y27choi: These constants seem to be inconsistent with the usage of `S3_BUCKET`.  Can we either drop `s3://kolena-public-datasets` from `S3_IMAGE_LOCATION` or add `s3://kolena-public-datasets` to `S3_ANNOTATION_FILE_PATH`?

examples/object_detection_2d/object_detection_2d/seed_test_run.py
```diff
@@ -0,0 +1,154 @@
...
+from kolena._experimental.object_detection import Model
+from kolena._experimental.object_detection import ObjectDetectionEvaluator
+from kolena._experimental.object_detection import TestSample
+from kolena._experimental.object_detection import TestSuite
+from kolena._experimental.object_detection import ThresholdConfiguration
+from kolena._experimental.object_detection import ThresholdStrategy
+from kolena.workflow.annotation import ScoredLabeledBoundingBox
+from kolena.workflow.test_run import test
+
+MODEL_INFERENCE_PREFIX = "s3://kolena-public-datasets/coco-2014-val/results/object_detection/coco_models/"
```
y27choi: use `S3_BUCKET` from `constants.py`?

examples/object_detection_2d/object_detection_2d/seed_test_run.py
```diff
@@ -0,0 +1,154 @@
...
+
+# transforms test samples into inferences using a dataframe
+def get_stored_inferences(
+    metadata_by_image: pd.DataFrame,
+) -> Callable[[TestSample], Inference]:
+    # a function that creates an inference from a test sample
+    def infer(sample: TestSample) -> Inference:
+        try:
+            image_loc = sample.locator
+            image_name = "imgs/" + image_loc.split("/")[-1]
```
y27choi: It will be nice to have the image locator as one of the columns in the results csv file so no extra processing is needed.

examples/object_detection_2d/object_detection_2d/seed_test_run.py
```diff
@@ -0,0 +1,154 @@
...
+                        top_left=(float(record.min_x), float(record.min_y)),
+                        bottom_right=(float(record.max_x), float(record.max_y)),
+                        label=str(record.label),
+                        score=float(record.confidence_score),
+                    )
+                    for record in image_inferences.itertuples()
+                ],
+            )
+        except Exception as e:
+            if e is None:
```
y27choi, line 78: should this be `if e is not None`?

examples/object_detection_2d/object_detection_2d/seed_test_run.py
gordonhart: Agreed here.

Also is there a particular reason to use CSV here instead of COCO format JSON? (not critical for now)

kolena/_experimental/object_detection/workflow.py
```diff
@@ -174,8 +174,8 @@ def display_name(self) -> str:
 class ThresholdConfiguration(EvaluatorConfiguration):
     threshold_strategy: ThresholdStrategy
     iou_threshold: float
-    min_confidence_score: float
     with_class_level_metrics: bool
+    min_confidence_score: Optional[float] = dataclasses.field(default=0.0)
```
gordonhart: For a scalar value like `float`, you can just provide the default directly:

```py
class ThresholdConfiguration(EvaluatorConfiguration):
    ...
    min_confidence_score: float = 0.0
```

examples/object_detection_2d/tests/test_object_detection_2d.py
gordonhart, line 1: Rather than committing a TODO file, I would omit the `tests` directory in this PR and add in a follow-up.

examples/object_detection_2d/pyproject.toml
```diff
@@ -0,0 +1,22 @@
...
+description = "Example Kolena integration for 2D object detection"
+authors = ["Kolena Engineering <eng@kolena.io>"]
+license = "Apache-2.0"
+
+[tool.poetry.dependencies]
+python = ">=3.8,<3.12"
+kolena = ">=0.76.0,<1"
+s3fs = "^2023.5.0"
+fsspec = "^2023.5.0"
+scikit-learn = "^1.1.2"
```
gordonhart: Should we install `kolena[metrics]` instead of adding `scikit-learn` manually?

gordonhart:
Thanks for getting this up! Looks almost there for a first pass. I would also request that you add an entry to `examples/README.md` with a single sentence for this new example workflow

y27choi:
Thanks to per-class metrics, we don't use `ignored_bboxes` as much as we used to, but it would still be nice to have another test suite that illustrates how `ignored_bboxes` are used (maybe stratification by the bbox size?).

examples/object_detection_2d/object_detection_2d/seed_test_run.py
marko-polo-cheno: wanted silent failures - had rare pandas errors before, will improve in the future for sure

examples/object_detection_2d/object_detection_2d/seed_test_run.py
marko-polo-cheno: I think most model predictions come out as csvs, or some sort of less structured format. JSONs would be a nice improvement to avoid grouping. Cacheing might also be a nice alternative because of how small the subset of data is.

y27choi (approved):

examples/object_detection_2d/pyproject.toml
gordonhart: Now that we're installing `kolena[metrics]` can we remove this explicit sklearn dependency?

kolena/_experimental/object_detection/workflow.py
```diff
@@ -174,8 +174,8 @@ def display_name(self) -> str:
 class ThresholdConfiguration(EvaluatorConfiguration):
     threshold_strategy: ThresholdStrategy
     iou_threshold: float
-    min_confidence_score: float
     with_class_level_metrics: bool
+    min_confidence_score: Optional[float] = 0.0
```
gordonhart: Is there a reason for this to be `Optional[float]`? It looks like a normal `float` with a default value provided. I can't think of a reason why we'd want to accommodate `min_confidence_score=None`, but I may be missing something.

examples/object_detection_2d/object_detection_2d/seed_test_suite.py
```diff
@@ -0,0 +1,156 @@
...
+
+    # create the test suite with the complete test case and new test cases
+    test_suite = TestSuite(
+        test_suite_name,
+        description=f"{TEST_SUITE_DESCRIPTION}, stratified by `light`, `normal`, and `dark` brightness",
+        test_cases=[complete_test_case, *test_cases],
+        reset=True,
+    )
+
+    print(f"created test suite {test_suite.name} v{test_suite.version}")
```
gordonhart: Nit: wrap `test_suite.name` in single quotes so it's easier to copy/paste to `seed_test_run` command

examples/object_detection_2d/object_detection_2d/seed_test_run.py
```diff
@@ -0,0 +1,160 @@
...
+            ),
+        ],
+    )
+
+    # runs the evaluation
+    test(model, test_suite, evaluator, reset=True)
+
+
+def main(args: Namespace) -> None:
+    print("loading csv of inferences...")
```
gordonhart: This took more than a minute for me to download, during which there was no feedback about what was happening. We should GZip these files to minimize data transfer — the end result will be transparent to the user, and a quick test on `yolo_r` shows the file compressing down to ~60MB from ~330MB.

gordonhart (approved):
A few small comments but I ran through the scripts on scratch and the output looks good. Nothing blocking.

There's a significant amount of cleanup we can do here — these scripts should be ~half as long and ~half as complex — but this is not critical in the short term.

gordonhart (approved):

y27choi (approved):

PR #146, opened 2023-07-17
pin pandera<0.16 in examples

sandalns:
### Linked issue(s):
Pin pandera<0.16 for examples.

### DIFF
--- a/examples/age_estimation/pyproject.toml
+++ b/examples/age_estimation/pyproject.toml
@@ -9,6 +9,7 @@ license = "Apache-2.0"
 python = ">=3.8,<3.12"
 kolena = ">=0.76.0,<1"
 s3fs = "^2023.5.0"
+pandera = ">=0.9.0,<0.16"
 
 [tool.poetry.group.dev.dependencies]
 pre-commit = "^2.17"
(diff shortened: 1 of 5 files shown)


gordonhart:
Maybe we should lock these examples such that installs are deterministic. This will prevent us from having to pin transitive dependencies like this. Thoughts @ly-kolena @munkyshi?

ly-kolena:
> Maybe we should lock these examples such that installs are deterministic. This will prevent us from having to pin transitive dependencies like this. Thoughts @ly-kolena @munkyshi?

One concern I have about that is maintenance burden, e.g. keeping track of dependencies of dependencies and version updates.

gordonhart:
The alternative is this maintenance burden, where we need to update every example to pin transitive dependencies in the time between fixing this on `trunk` and releasing the next `kolena` version. This has proven to be non-negligible

ly-kolena:
> The alternative is this maintenance burden, where we need to update every example to pin transitive dependencies in the time between fixing this on `trunk` and releasing the next `kolena` version. This has proven to be non-negligible

We're not shipping examples in `kolena` package. It would be comparatively easy to fix up the versions for later users. For those who already used them to clone their project likely can manage on their own.

I'm not tipped over to take the maintenance burden yet, but not very strongly against it.

ly-kolena (approved):

PR #150, opened 2023-07-17
update 3d od sample metrics

ly-kolena:

1. Updated ground_truth format - removed `n_easy`, `n_moderate`,  `n_hard`, and `n_unknown` because their computation was not a correct interpretation of those criteria. Similar updates to bounding_box's `difficulty` value.
2. Updated test sample metrics: include TP/FP/FN

preview in [studio](https://trunk.kolena.io/liuyuan/studio?pageSize=50&testSuiteId=28&testCaseId=342&workflow=KITTI+3D+Object+Detection&models=N4IglgJiBcAsA0IDGB7AdgMzAcwK4CcBDAFzHRlEhgEYBfWoA&models=N4IglgJiBcAsA0IDGB7AdgMzAcwK4CcBDAFzHRlEhgCYBfWoA&models=N4IglgJiBcAsA0IDGB7AdgMzAcwK4CcBDAFzHRlEhgGYBfWoA&sortByType=test_sample&sortBy=locator&overlays=N4Igtg9gJgpgNgZxALlAFhaADJkALCAFwGsYBPFEAQxABoQA3ASwSYCM4YVCAnAVxgBfegEZcBEuUps6jFu07d%2BQ%2BgCZxRUhWQgAxrOasOXZLwGCL9KgDtrRKoSYRrSVCEIwEhAPoIqYAAdFVGF3Tx8-QM5vMBhCKigHGhD6AHMeCD5rKG8zQjxMUKZrADMYHhhrXRNgUI8vX38gmBi4niZdV1qLIA)

### DIFF
--- a/examples/object_detection_3d/object_detection_3d/evaluator.py
+++ b/examples/object_detection_3d/object_detection_3d/evaluator.py
@@ -39,6 +39,7 @@ from kolena.workflow import MetricsTestCase as BaseMetricsTestCase
 from kolena.workflow import MetricsTestSample as BaseMetricsTestSample
 from kolena.workflow.annotation import LabeledBoundingBox
 from kolena.workflow.annotation import LabeledBoundingBox3D
+from kolena.workflow.annotation import ScoredClassificationLabel
 from kolena.workflow.annotation import ScoredLabeledBoundingBox
 from kolena.workflow.annotation import ScoredLabeledBoundingBox3D
 from kolena.workflow.plot import Curve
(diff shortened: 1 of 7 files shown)


examples/object_detection_3d/object_detection_3d/evaluator.py
```diff
@@ -79,9 +87,22 @@ def name(self):
 
 @dataclass(frozen=True)
 class MetricsTestSample(BaseMetricsTestSample):
-    nObjects: int
     nInferences: int
     nValidObjects: int
+    thresholdCar: float
+    thresholdCyclist: float
+    thresholdPedestrian: float
```
gordonhart: `thresholds: List[ScoredClassificationLabel]` is the strategy we use for prebuilt OD — not sure if we want to apply that here too, but it could be nice.

examples/object_detection_3d/object_detection_3d/evaluator.py
```diff
@@ -77,11 +86,53 @@ def name(self):
         return name
 
 
+@dataclass(frozen=True)
+class UnmatchedLabeledBoundingBox3D(LabeledBoundingBox3D):
+    max_overlap: float
```
ly-kolena, line 91: Added `max_overlap` for FP and FN because I wasted a day thinking metrics implementation may be wrong because some boxes look to have high overlap visually. Turned out visual is deceiving and overlap is indeed below threshold.

ly-kolena:
Will update pre-computed files once this PR is merged.

examples/object_detection_3d/object_detection_3d/evaluator.py
```diff
@@ -77,11 +86,53 @@ def name(self):
...
 @dataclass(frozen=True)
 class MetricsTestSample(BaseMetricsTestSample):
-    nObjects: int
     nInferences: int
     nValidObjects: int
+    thresholds: List[ScoredClassificationLabel]
+    nMatchedInferences: int
+    nMissedObjects: int
+    nMismatchedInferences: int
+    valid_2D: List[LabeledBoundingBox]
```
ly-kolena: This is the replacement for previous `n_easy`, `n_moderate`, and `n_hard`, where only ground_truths matching the target EvaluatorConfiguration (easy, moderate, or hard) are considered valid.

ly-kolena:
side note: even `numba.jit` (not cuda.jit) made a significant difference in computation time for this relatively small sample size (non-scientific measure). Something we can think about.

examples/object_detection_3d/object_detection_3d/evaluator.py
gordonhart: This makes for a noisier Studio experience than is ideal — it wasn't clear to me that this was a ground truth bbox reproduced in the metrics section. When navigating these results locally, I found that I had to turn off the `valid_*` sections to follow what was going on.

Can the valid annotations for a given eval config be inferred from the TP/FN annotations? This seems like the least confusing way to model this, taking into account the way the Studio behaves for visualization.

gordonhart (approved):

PR #162, opened 2023-07-28
Remove version upper bound from `importlib-metadata`

gordonhart:
Remove upper bound for `importlib-metadata` — this package is only used for Python<3.8 (after which it became a part of the standard library) and was originally restricted to `<5.0` for reasons that are no longer relevant.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,7 @@ pydantic = ">=1.8,<2"
 dacite = ">=1.6,<2"
 requests = ">=2.20,<2.30" # TODO: revert upper bound when urllib3 situation sorts out: https://github.com/psf/requests/issues/6432
 requests-toolbelt = "*"
-importlib-metadata = { version = "<5.0", python = "<3.8" }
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -39,7 +39,7 @@ pydantic = ">=1.8,<2"
 dacite = ">=1.6,<2"
 requests = ">=2.20,<2.30" # TODO: revert upper bound when urllib3 situation sorts out: https://github.com/psf/requests/issues/6432
 requests-toolbelt = "*"
-importlib-metadata = { version = "<5.0", python = "<3.8" }
+importlib-metadata = { version = "*", python = "<3.8" }
```
ly-kolena, line 42: nit: latest version `6.8.0` requires `python>=3.8`. `6.7.0` is good for `python >=3.7`

gordonhart: Pip should be smart enough to figure out the max version to install given the current Python version. As far as I am aware, we don't need to pin an upper bound here directly.

ly-kolena (approved):

PR #165, opened 2023-07-28
Add upload embeddings client function

munkyshi:
### Linked issue(s):
Client side fix: PRO-65

Add upload embeddings client function


kolena/_experimental/search/embeddings.py
```diff
@@ -0,0 +1,58 @@
...
+from kolena._experimental.search._internal.datatypes import LocatorEmbeddingsDataFrameSchema
+from kolena._utils import krequests
+from kolena._utils import log
+from kolena._utils.batched_load import init_upload
+from kolena._utils.batched_load import upload_data_frame
+from kolena._utils.consts import BatchSize
+from kolena._utils.dataframes.validators import validate_df_schema
+
+
+def upload_embeddings(embeddings: List[Tuple[str, np.typing.ArrayLike]]) -> None:
```
munkyshi: The more I look at this, the more I think we should include `model_locator` as part of the data bundle (instead of inferring from the galleon table)

ly-kolena: Do we control the model, i.e. can we handle user-provided data here?

kolena/_experimental/search/_internal/datatypes.py
```diff
@@ -0,0 +1,75 @@
...
+        return cell.dtype in __ALLOWED_NUMERIC_DTYPES
+
+    return _validate_ndarray(series) and series.dropna().apply(validate_cell).all()
+
+
+class LocatorEmbeddingsDataFrameSchema(pa.SchemaModel):
+    locator: Series[pa.typing.String] = pa.Field(coerce=True, _validate_locator=())
+    """External locator pointing to a sample in bucket."""
+
+    embedding: Series[EmbeddingVector] = pa.Field(coerce=True, _validate_embedding_vector=())
```
ly-kolena: This looks unit-testable

kolena/_experimental/search/embeddings.py
gordonhart: Would users call this in the same script that they call `extract_embeddings` from?

munkyshi: > Do we control the model, i.e. can we handle user-provided data here?

We currently bake the model into the bundle that we provide to our customers.

> Would users call this in the same script that they call extract_embeddings from?

Hmm we could update the `extract_embeddings` script to include an `upload_embeddings(images: List[PIL.Image.Image])` entrypoint that automatically handles upload. Taking a look.

munkyshi: Okay, updated this approach -- including `key` to act as a user friendly `model_locator` (since it actually just acts as a key).

Proposing [this](https://github.com/kolenaIO/kolena-studio-client-api/pull/13/files#diff-34b4a4dec40fd50e2de6307f12d8c3a2578fdc9deef4290ccdb66c040bdb5dbaR21) change for how users end up calling into this endpoint.

kolena/_experimental/search/_internal/datatypes.py
munkyshi: Updated so that we don't validate the embedding vector in the pandera schema anymore.

munkyshi:
Related server side PR: https://github.com/kolenaIO/shipyard/pull/1537

kolena/_experimental/search/embeddings.py
```diff
@@ -0,0 +1,68 @@
...
+    request = API.UploadEmbeddingsRequest(
+        uuid=init_response.uuid,
+    )
+    res = krequests.post(
+        endpoint_path=API.Path.EMBEDDINGS.value,
+        data=json.dumps(dataclasses.asdict(request)),
+    )
+    krequests.raise_for_status(res)
+    data = from_dict(data_class=API.UploadEmbeddingsResponse, data=res.json())
+    log.success(f"uploaded embeddings for {data.n_samples} samples")
```
gordonhart: Nit: pair with corresponding `log.info` at the beginning of the operation

kolena/_experimental/search/embeddings.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def upload_embeddings(key: str, embeddings: List[Tuple[str, np_typing.ArrayLike]]) -> None:
+    """
+    Upload a list of search embeddings corresponding to sample locators.
+
+    :param key: String value uniquely corresponding to the model used to extract the embedding vectors.
+        This is typically a locator.
+    :param embeddings: List of locator-embedding pairs, as tuples. Locators should be string values, while embeddings
+        should be an `numpy.typing.ArrayLike` of numeric values.
```
gordonhart, line 43: Why not `np.ndarray`?

gordonhart:
What is the expected usage here? Do you have a script you can share of the new end-to-end flow?

gordonhart (approved):
Approving. Noted a few nits but overall looks good. Curious to see how it all plugs together. From there, we can decide on adding docs and promoting out of experimental status.

munkyshi:
> What is the expected usage here? Do you have a script you can share of the new end-to-end flow?

https://github.com/kolenaIO/kolena-studio-client-api/pull/13/files

kolena/_experimental/search/embeddings.py
munkyshi: Mostly just trying to be as permissive as possible for user facing client code -- you could theoretically upload a `List[float]` and still have that be a valid embedding.

gordonhart: For what it's worth, we use `np.ndarray` in signatures throughout the client already, and `np.typing.ArrayLike` was introduced in 1.20 which may not be installed given our current version pin for `numpy`.

munkyshi: Ack, I'll just use `ndarray`

PR #166, opened 2023-07-29
Example of extending 2D Object Detection Pre-built Workflow

y27choi:
### Linked issue(s):
Fixes KOL-2728

[example on Kolena](https://trunk.kolena.io/kolena/results?modelIds=1568&workflow=Object+Detection+Extend&testSuiteId=3214)


examples/object_detection_2d_extend/object_detection_2d_extend/workflow.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+@dataclass(frozen=True)
+class Inference(BaseInference):
+    ...
+
+
+@dataclass(frozen=True)
+class TestSample(BaseTestSample):
+    ...
```
gordonhart: These don't need to be explicitly extended like this if there aren't any new fields being added — can just use the types imported from the prebuilt OD workflow directly.

examples/object_detection_2d_extend/object_detection_2d_extend/constants.py
gordonhart, line 1: This is a lot of copypasta to demo a simple extension. Can we refactor to have this as maybe a subdirectory in the existing `object_detection_2d` example? We should be able to remove all but the `workflow.py` changes and a new test suite seed script.

examples/object_detection_2d/README.md
```diff
@@ -47,3 +49,32 @@ optional arguments:
...
+
+@dataclass(frozen=True)
+class GroundTruth(BaseGroundTruth):
+    bboxes: List[ExtendedBoundingBox]
+```
+
+And define a new workflow with the extended `GroundTruth` class:
+
+```python
+_workflow, TestCase, TestSuite, Model = define_workflow(
```
gordonhart, line 71: Nit: I'd omit the `workflow` in favor of just `_`. At some point (1.0?) we will likely remove workflow from this signature as it isn't particularly useful.

examples/object_detection_2d/object_detection_2d/extended/seed_test_suite.py
```diff
@@ -0,0 +1,208 @@
...
+    complete_test_case = TestCase(
+        f"complete transportation :: {DATASET} [{WORKFLOW}]",
+        test_samples=test_samples_and_ground_truths,
+        reset=True,
+    )
+
+    return complete_test_case
+
+
+def seed_test_suite_by_brightness(test_suite_name: str, complete_test_case: TestCase) -> None:
```
gordonhart, line 109: Since this example is just meant to illustrate the extension of the workflow, we can just make a single test case test suite and pare this `seed_test_suite` script down significantly. Your call

gordonhart (approved):

examples/object_detection_2d/object_detection_2d/extended/seed_test_suite.py
y27choi: I agree that it will pare it down quite a bit but I prefer to leave it as is because `seed_test_run` has a list of `TestSuite` names specified by default so you can run the script without passing in the test suite name as an argument but the default argument won't work if I make this change.

PR #169, opened 2023-08-02
Binary Classification Evaluator

y27choi:
### Linked issue(s):
Fixes KOL-2704

* Refactors multiclass classification evaluator
* Modified the workflow schema 

[dogs-vs-cats binary classification (threshold=0.5)](https://trunk.kolena.io/yoohee/results?modelIds=684&workflow=Image+Classification&testSuiteId=1370&evaluators=N4IgjAzAbAHCBcpZ3mADATgCwF8A04EA7GgkjFguhlAbAKxWa0ixRM045A)
[dogs-vs-cats multiclass classification](https://trunk.kolena.io/yoohee/results?modelIds=686&workflow=Image+Classification&testSuiteId=1370&evaluators=N4IgjAzAbAHCBcpZ3mADATgCwF8A04EA7GgkjFguhlAbAKxWa0ixRM045A)

Based on the # of prediction labels seeded as part of `Inference` determines which evaluator to run.  If only positive label's prediction is used, then it will be treated as a binary classification.  If both labels' predictions are added, then it will be 2 class multiclass evaluation. 

How are multiclass and binary evaluator different?:
* Per-class evaluation for multiclass (for both "dog" and "cat" class metrics are available) (e.g., TP for class "cat" is for correctly predicted "cat" class) — binary shows the same information but in different wordings (e.g., TP is for positive label which in this case is "dog" and TN is for negative label which in this case is "cat") 
* ROC curve on positive label only for binary vs on both labels for multiclass
* Score distribution (for binary, only the positive prediction scores are used so it ranges from 0 to 1; whereas, for multiclass, maximum confidence score label is used so it will range from 0.5 - 1.
* Classification label annotation: for binary, it will only show the prediction on the positive label and for multiclass, the label which has the "maximum" confidence score will be shown. 

Screen Shot 2023-08-03 at 9 58 53 AM Screen Shot 2023-08-03 at 9 59 02 AM

* TestSampleMetrics: for binary, there are more fields like `is_TP`, `is_FN` and etc., whereas for multiclass, only `is_correct` and predicted label is shown.

Screen Shot 2023-08-03 at 10 01 08 AM
Screen Shot 2023-08-03 at 10 01 20 AM

For binary classification cases where performance on positive label can be clearly defined (e.g., "is_live" or "is_dog") then using binary classification evaluator is a better experience.  For cases where both label are equally important (e.g., "dog" vs "cat") where a positive label cannot be specified, using multiclass evaluator is preferred.

Proposal on UX improvement:
Once we make this into a pre-built, we should consider changing the color of `TestSampleMetrics.classification` label to be green (if `is_correct=True`) or red (otherwise).

### DIFF
--- a/examples/classification/classification/evaluator_binary.py
+++ b/examples/classification/classification/evaluator_binary.py
@@ -1,0 +1,124 @@
+# Copyright 2021-2023 Kolena Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import List
+from typing import Optional
+from typing import Tuple
+
+from classification.evaluator_base import BaseClassificationEvaluator
+from classification.workflow import GroundTruth
+from classification.workflow import Inference
+from classification.workflow import TestCaseMetricsSingleClass
+from classification.workflow import TestSampleMetricsSingleClass
+from classification.workflow import ThresholdConfiguration
+
+from kolena._experimental.classification.utils import compute_confusion_matrix
+from kolena._experimental.classification.utils import compute_roc_curves
+from kolena._utils import log
+from kolena.workflow import Plot
+from kolena.workflow.metrics import accuracy as compute_accuracy
+from kolena.workflow.metrics import f1_score as compute_f1_score
+from kolena.workflow.metrics import precision as compute_precision
+from kolena.workflow.metrics import recall as compute_recall
+
+
+class BinaryClassificationEvaluator(BaseClassificationEvaluator):
+    def compute_test_sample_metrics(
+        self,
+        ground_truth: GroundTruth,
+        inference: Inference,
+        threshold_configuration: ThresholdConfiguration,
+    ) -> TestSampleMetricsSingleClass:
+        if len(inference.inferences) != 1:
+            log.warn(
+                f"detected more than one ({len(inference.inferences)}) predicted labels for a binary "
+                "classification test suite — aborting test",
+            )
+            raise ValueError
+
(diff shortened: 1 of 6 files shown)


y27choi:
will have a separate PR for seeding scripts and doc updates

examples/classification/classification/workflow.py
```diff
@@ -65,24 +78,83 @@ class Inference(BaseInference):
...
+    """
+
+    missing_inference: bool = False
+
+    def __post_init__(self):
+        object.__setattr__(self, "missing_inference", self.classification is None)
+
+
 @dataclass(frozen=True)
 class TestSampleMetrics(MetricsTestSample):
```
ly-kolena, line 136: Took me some time to figure out what's the difference between `TestSampleMetricsSingleClass` and `TestSampleMetrics` (hopefully I'm not wrong). They have the same class description. Is it intentional to have them the same? Would it be helpful to describe which class to use for which problem?

ly-kolena (approved):

examples/classification/classification/workflow.py
y27choi: Thanks for catching it.  No they are not supposed to be the same. I will update it to be more descriptive.

examples/classification/classification/evaluator_binary.py
```diff
@@ -0,0 +1,124 @@
...
+        plots.append(
+            compute_confusion_matrix(
+                [gt.classification.label for gt in ground_truths],
+                [positive_label if metric.is_TP or metric.is_FP else negative_label for metric in metrics],
+                labels=[positive_label, negative_label],
+            ),
+        )
+        plots = list(filter(lambda plot: plot is not None, plots))
+
+        return plots
```
gordonhart, line 124: Should we add an Accuracy vs. T plot? On the [linked binary classification example](https://trunk.kolena.io/yoohee/results?modelIds=684&workflow=Image+Classification&testSuiteId=1370&evaluators=N4IgjAzAbAHCBcpZ3mADATgCwF8A04EA7GgkjFguhlAbAKxWa0ixRM045A), I find myself wanting to know the optimal threshold, which isn't quite conveyed in any of the existing plots.

gordonhart (approved):

examples/classification/classification/evaluator_binary.py
y27choi: That's a great idea. I will add a threshold plot (with accuracy, precision, recall, f1-score vs. T).

PR #172, opened 2023-08-03
Add derived values to kolena.workflow.BoundingBox{,3D}

gordonhart:
Automatically hydrate `BoundingBox` and `BoundingBox3D` with derived fields such as `area` and `volume`. This approach is transparent to the end user and does not impact serde or extension of these annotation objects.

### DIFF
--- a/kolena/workflow/annotation.py
+++ b/kolena/workflow/annotation.py
@@ -89,11 +89,10 @@ class BoundingBox(Annotation):
         return _AnnotationType.BOUNDING_BOX
 
     def __post_init__(self) -> None:
-        object.__setattr__(self, "width", getattr(self, "width", self.bottom_right[0] - self.top_left[0]))
-        object.__setattr__(self, "height", getattr(self, "height", self.bottom_right[1] - self.top_left[1]))
-        object.__setattr__(self, "area", getattr(self, "area", self.width * self.height))
-        aspect_ratio = getattr(self, "aspect_ratio", self.width / self.height if self.height != 0 else 0)
-        object.__setattr__(self, "aspect_ratio", aspect_ratio)
+        object.__setattr__(self, "width", self.bottom_right[0] - self.top_left[0])
+        object.__setattr__(self, "height", self.bottom_right[1] - self.top_left[1])
+        object.__setattr__(self, "area", self.width * self.height)
+        object.__setattr__(self, "aspect_ratio", self.width / self.height if self.height != 0 else 0)
 
 
 @dataclass(frozen=True, config=ValidatorConfig)
@@ -230,7 +229,7 @@ class BoundingBox3D(Annotation):
         return _AnnotationType.BOUNDING_BOX_3D
 
     def __post_init__(self) -> None:
-        object.__setattr__(self, "volume", getattr(self, "volume", reduce(lambda a, b: a * b, self.dimensions)))
+        object.__setattr__(self, "volume", reduce(lambda a, b: a * b, self.dimensions))
 
 
 @dataclass(frozen=True, config=ValidatorConfig)
--- a/kolena/workflow/annotation.py
+++ b/kolena/workflow/annotation.py
@@ -31,6 +31,7 @@ For example, when viewing images in the Studio, any annotations (such as lists o
 [`Inference`][kolena.workflow.Inference], or [`MetricsTestSample`][kolena.workflow.MetricsTestSample] objects are
 rendered on top of the image.
 """
+import dataclasses
 from abc import ABCMeta
 from typing import Dict
 from typing import List
@@ -73,9 +74,20 @@ class BoundingBox(Annotation):
     bottom_right: Tuple[float, float]
     """The bottom right vertex (in `(x, y)` image coordinates) of this bounding box."""
 
+    height: float = dataclasses.field(init=False)
+    width: float = dataclasses.field(init=False)
+    area: float = dataclasses.field(init=False)
+    aspect_ratio: float = dataclasses.field(init=False)
+
     @staticmethod
     def _data_type() -> _AnnotationType:
         return _AnnotationType.BOUNDING_BOX
+
+    def __post_init__(self) -> None:
+        object.__setattr__(self, "width", self.bottom_right[0] - self.top_left[0])
+        object.__setattr__(self, "height", self.bottom_right[1] - self.top_left[1])
+        object.__setattr__(self, "area", self.width * self.height)
+        object.__setattr__(self, "aspect_ratio", self.width / self.height)
 
 
 @dataclass(frozen=True, config=ValidatorConfig)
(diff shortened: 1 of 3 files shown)


kolena/workflow/annotation.py
```diff
@@ -73,10 +74,21 @@ class BoundingBox(Annotation):
...
+    height: float = dataclasses.field(init=False)
+    width: float = dataclasses.field(init=False)
+    area: float = dataclasses.field(init=False)
+    aspect_ratio: float = dataclasses.field(init=False)
+
     @staticmethod
     def _data_type() -> _AnnotationType:
         return _AnnotationType.BOUNDING_BOX
 
+    def __post_init__(self) -> None:
```
ly-kolena, line 92: looks good to me.

munkyshi (approved):

y27choi:
This change can come in handy for stratification effort in `seed_test_suite.py` for all 2D OD workflows! 
Should we consider adding these derived fields to the API doc?

gordonhart:
> Should we consider adding these derived fields to the API doc?

I'm on the fence about this — it would add a fair bit of noise to have each of these derived fields in the API docs. Thoughts on including a sentence in the main object docstring saying something like:

> The fields `width`, `height`, `area`, and `aspect` ratio are automatically derived based on the provided coordinates.

tests/unit/workflow/test_annotation.py
```diff
@@ -75,9 +78,41 @@ class Tester(DataObject):
...
+
+@pytest.mark.parametrize(
+    "top_left,bottom_right,expected",
+    [
+        ((0, 0), (0, 0), dict(width=0, height=0, area=0, aspect_ratio=0)),
+        ((10, 10), (10, 10), dict(width=0, height=0, area=0, aspect_ratio=0)),
+        ((10, 10), (20, 30), dict(width=10, height=20, area=200, aspect_ratio=0.5)),
+    ],
+)
+def test__bounding_box__derived(
```
ly-kolena, line 147: Maybe a test to check deserialization also works?

gordonhart: Added another test to explicitly check this.

tests/unit/workflow/test_annotation.py
```diff
@@ -32,11 +35,29 @@
...
+    assert obj_dict == {
+        "top_left": [0, 0],
+        "bottom_right": [0, 0],
+        "width": 0,
+        "height": 0,
+        "area": 0,
+        "aspect_ratio": 0,
         DATA_TYPE_FIELD: f"{_AnnotationType._data_category()}/{_AnnotationType.BOUNDING_BOX.value}",
     }
+    assert BoundingBox._from_dict(obj_dict) == obj
```
ly-kolena, line 65: Nit: verified the scenario below works, but slightly different from the test here:

```
def test__deserialize__derived() -> None:
    obj_dict = {
        "top_left": [10, 10],
        "bottom_right": [20, 20],
        DATA_TYPE_FIELD: f"{_AnnotationType._data_category()}/{_AnnotationType.BOUNDING_BOX.value}",
    }
    assert BoundingBox._from_dict(obj_dict) == BoundingBox(top_left=(10, 10), bottom_right=(20, 20))
```

ly-kolena (approved):

tests/unit/workflow/test_annotation.py
gordonhart: Added check for deserialization from dict without derived fields 👍

gordonhart:
As a follow-up once this is released, we can simplify the `object_detection_2d` example to remove the additional `_area` function in favor of the derived value.

tests/unit/workflow/test_annotation.py
```diff
@@ -75,9 +128,41 @@ class Tester(DataObject):
...
     obj_dict = obj._to_dict()
 
-    assert obj_dict["b"] == {
-        "top_left": [0, 0],
-        "bottom_right": [1, 1],
-        DATA_TYPE_FIELD: f"{_AnnotationType._data_category()}/{_AnnotationType.BOUNDING_BOX.value}",
+    assert obj_dict["e"] == {
+        "points": [[0, 0], [1, 1], [2, 2], [0, 0]],
+        "label": "e",
+        DATA_TYPE_FIELD: f"{_AnnotationType._data_category()}/{_AnnotationType.POLYGON.value}",
```
munkyshi, line 134: Why is this deserializing as a `POLYGON` now?

kolena/workflow/annotation.py
```diff
@@ -65,18 +67,34 @@ class Annotation(TypedDataObject[_AnnotationType], metaclass=ABCMeta):
...
+        object.__setattr__(self, "width", getattr(self, "width", self.bottom_right[0] - self.top_left[0]))
+        object.__setattr__(self, "height", getattr(self, "height", self.bottom_right[1] - self.top_left[1]))
+        object.__setattr__(self, "area", getattr(self, "area", self.width * self.height))
+        aspect_ratio = getattr(self, "aspect_ratio", self.width / self.height if self.height != 0 else 0)
+        object.__setattr__(self, "aspect_ratio", aspect_ratio)
```
gordonhart: I'd appreciate another look here @ly-kolena @munkyshi. Updated the pattern to allow overriding the values of these derived fields. Overriding with different types also works, provided the extension class uses `@pydantic.dataclasses.dataclass`.

ly-kolena: May need to read a bit more how to properly use this, but I updated a unit test to use code like

```
obj = BoundingBox(top_left=(5, 5), bottom_right=(10, 10), width=25, height=15, area=255, aspect_ratio=21)
```
and got error.

gordonhart: The derived fields are intentionally left out of the init constructor for the base objects, but this `getattr` approach allows users to define their own `BoundingBox`-extending class to explicitly specify or override these values like this:

```python
@dataclass(frozen=True)
class MyCustomBoundingBox(BoundingBox):
    area: int
    supercategory: str
    ...
```

Before these changes, any `area` value provided to `MyCustomBoundingBox` would be overridden by the derived value in `__post_init__`.

ly-kolena: > provided the extension class uses @pydantic.dataclasses.dataclass

Weighing between the value of having this feature, and potential footnotes a user has to be aware of to use of it, not sure if it is worth putting in. No strong objection as this doesn't impede with original use cases.

As a side note, is there a place for a quick example or footnote if a user does want to override these derived fields?

ly-kolena (approved):
looks good

kolena/workflow/annotation.py
gordonhart: Good call. I've removed the override functionality and marked these fields has "reserved" in the relevant docstrings.

PR #173, opened 2023-08-07
Seeding scripts for binary classification example

y27choi:
### Linked issue(s):
Fixes KOL-2704

Seeding scripts for binary classification example (dogs-vs-cats)


examples/README.md
```diff
@@ -16,3 +16,5 @@ arbitrary ML problem.
   [CNN-DailyMail](https://paperswithcode.com/dataset/cnn-daily-mail-1) dataset
 - [Mutliclass Classification](https://github.com/kolenaIO/kolena/tree/trunk/examples/classification#multiclass-classification-on-cifar-10): Multiclass Classification
   on image data from [CIFAR-10](https://www.cs.toronto.edu/~kriz/cifar.html) dataset
+- [Binary Classification](https://github.com/kolenaIO/kolena/tree/trunk/examples/classification#binary-classification-on-dogs-vs-cats):
+  Binary Classification on image data from [Dogs vs. Cats](https://www.kaggle.com/c/dogs-vs-cats) dataset
```
gordonhart, line 20: Also add to [quickstart](https://docs.kolena.io/quickstart/#clone-the-examples) guide?

examples/classification/scripts/binary/seed_test_suite.py
```diff
@@ -0,0 +1,98 @@
...
+        f"complete {DATASET}",
+        description=f"All images in {DATASET} dataset",
+        test_samples=test_samples_and_ground_truths,
+        reset=True,
+    )
+
+    # Metadata Test Cases
+    stratification_logic_map = {
+        "small": lambda area: 0 <= area < 80000,
+        "large": lambda area: area >= 160000,
```
gordonhart: Is there a missing `medium` size here?

examples/README.md
y27choi: Yup, I got another branch started for the doc update.

y27choi:
Found an edge case (test case with only positive samples or only negative samples) so I had to introduce a new logic to infer the negative label's name. 

If the test suite only has positive samples, then the negative label will be: "NOT {positive_sample_label}" which is only used for confusion matrix.  An alternative approach is to use "TP", "FP", "FN", and "TN" instead of using the actual label name for binary classification's confusion matrix.  This will involve changing the logic in `kolena._experimental.classification.utils.compute_confusion_matrix` so I prefer to leave it as is considering that this is be a rare case.

gordonhart:
> Found an edge case (test case with only positive samples or only negative samples) so I had to introduce a new logic to infer the negative label's name.

This complication is not ideal, but otherwise 👍

gordonhart (approved):

PR #174, opened 2023-08-07
Use derived bbox area in 2D OD example seed scripts

gordonhart:
`area` was introduced in #172 as a derived value of the provided coordinates. No need to implement the `_area` function in the 2D OD example seed scripts.

### DIFF
--- a/examples/object_detection_2d/object_detection_2d/extended/seed_test_suite.py
+++ b/examples/object_detection_2d/object_detection_2d/extended/seed_test_suite.py
@@ -36,7 +36,6 @@ from object_detection_2d.extended.workflow import TestSample
 from object_detection_2d.extended.workflow import TestSuite
 
 import kolena
-from kolena.workflow.annotation import BoundingBox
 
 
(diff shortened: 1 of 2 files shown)


examples/object_detection_2d/object_detection_2d/extended/seed_test_suite.py
```diff
@@ -56,7 +55,7 @@ def load_transportation_data() -> Dict[str, List[ExtendedBoundingBox]]:
     label_map: Dict[int, str] = {int(category["id"]): category["name"] for category in coco_data["categories"]}
 
     # cache bounding boxes per image
-    image_to_boxes: Dict[int, List[ExtendedBoundingBox]] = defaultdict(lambda: [])
+    image_to_boxes: Dict[str, List[ExtendedBoundingBox]] = defaultdict(lambda: [])
```
munkyshi, line 58: I'm surprised mypy didn't previously catch this.

munkyshi (approved):

PR #176, opened 2023-08-08
Remove superfluous 'labels' fields from prebuilt OD objects

gordonhart:
No longer necessary with new list filtering functionality.

Fixes KOL-2799

### DIFF
--- a/kolena/_experimental/object_detection/workflow.py
+++ b/kolena/_experimental/object_detection/workflow.py
@@ -55,11 +55,9 @@ class GroundTruth(BaseGroundTruth):
     associated with an image.
     """
 
-    labels: List[str] = dataclasses.field(default_factory=list)
     n_bboxes: int = dataclasses.field(default_factory=lambda: 0)
 
     def __post_init__(self):
-        object.__setattr__(self, "labels", sorted({box.label for box in self.bboxes}))
         object.__setattr__(self, "n_bboxes", len(self.bboxes))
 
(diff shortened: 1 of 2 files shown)


gordonhart:
@marko-polo-cheno do you agree that this is no longer necessary?

marko-polo-cheno:
> @marko-polo-cheno do you agree that this is no longer necessary?

Agree - The new list filtering is really nice!

kolena/_experimental/object_detection/workflow.py
```diff
@@ -122,13 +120,9 @@ class TestCaseMetricsSingleClass(MetricsTestCase):
 
 @dataclass(frozen=True)
 class TestSampleMetrics(MetricsTestSample):
-    TP_labels: List[str]
```
ly-kolena, line 125: Do we lose visualization capability with this change? I suppose this is a quick way to see all labels

ly-kolena (approved):

kolena/_experimental/object_detection/workflow.py
gordonhart: You would have to visually scan the list of boxes, yeah. I think this is an acceptable tradeoff for reduced noise. If we wanted to preserve this functionality, we could extend the "X Bounding Boxes" summary to include the unique set of labels — this may be worth doing anyway to increase the signal in this collapsible header.

PR #178, opened 2023-08-08
Threshold curves for binary classification example

y27choi:
### Linked issue(s):
Fixes KOL-2793

* Adds util function for computing threshold curves for a single label
* Adds threshold curves (precision, recall and f1 vs. thresholds) for binary classification evaluator
* Intentionally left out accuracy as precision, recall and f1 scores are more meaningful metrics.

[example plot](https://trunk.kolena.io/yoohee/results?modelIds=684&testSuiteId=1371&workflow=Image+Classification&evaluators=N4IgjAzA7GIFygGwA4As8wAYCciC%2BeQA)

### DIFF
--- a/kolena/_experimental/classification/utils.py
+++ b/kolena/_experimental/classification/utils.py
@@ -207,7 +207,7 @@ def compute_roc_curves(
 
     :param ground_truths: The list of ground truth
         [`ClassificationLabel`][kolena.workflow.annotation.ClassificationLabel]. For binary classification, the negative
-        class is `None`.
+        class can be `None`.
     :param inferences: The list of inference
         [`ScoredClassificationLabel`][kolena.workflow.annotation.ScoredClassificationLabel]. For `N`-class problems,
         each inference is expected to contain `N` entries, one for each class and its associated confidence score.
@@ -261,6 +261,23 @@ def compute_threshold_curves(
     inferences: List[ScoredClassificationLabel],
     thresholds: Optional[List[float]] = None,
 ) -> Optional[List[CurvePlot]]:
+    """
+    Computes scores (i.e. [`Precision`][kolena.workflow.metrics.precision], [`Recall`](kolena.workflow.metrics.recall)
+    and [`F1-score`](kolena.workflow.metrics.f1_score)) vs. threshold curves for a **single** class presented in
+    `inferences`.
+
+    :param ground_truths: The list of ground truth
+        [`ClassificationLabel`][kolena.workflow.annotation.ClassificationLabel]. For binary classification, the negative
+        class can be `None`.
+    :param inferences: The list of inference
+        [`ScoredClassificationLabel`][kolena.workflow.annotation.ScoredClassificationLabel]. The list should only
+         include inferences of a specific class to plot the threshold curves for.
+    :param thresholds: The list of thresholds to plot with. If not specified, all the unique confidence scores are used
+        as thresholds, including evenly spaced thresholds from 0 to 1 with 0.1 step.
+    :return: A list of [`CurvePlot`][kolena.workflow.plot.CurvePlot]s if there is any valid `Curve` computed;
+        otherwise, `None`.
+
+    """
     if len(ground_truths) != len(inferences):
         log.warn(f"ground_truths ({len(ground_truths)}) and inferences ({len(inferences)}) differ in length")
         return None
--- a/kolena/_experimental/classification/utils.py
+++ b/kolena/_experimental/classification/utils.py
@@ -22,6 +22,9 @@ import numpy as np
 from kolena._utils import log
 from kolena.workflow.annotation import ClassificationLabel
 from kolena.workflow.annotation import ScoredClassificationLabel
+from kolena.workflow.metrics import f1_score
+from kolena.workflow.metrics import precision
+from kolena.workflow.metrics import recall
 from kolena.workflow.plot import ConfusionMatrix
 from kolena.workflow.plot import Curve
 from kolena.workflow.plot import CurvePlot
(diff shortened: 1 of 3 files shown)


tests/unit/_experimental/classification/test_utils.py
```diff
@@ -435,3 +436,152 @@ def test__compute_roc_curves__multiclass() -> None:
...
+    "gts, infs",
+    [
+        ([], []),
+        ([ClassificationLabel(1)], []),
+        (
+            [ClassificationLabel(0), ClassificationLabel(1)],
+            [ScoredClassificationLabel(label=1, score=0.0)],
+        ),
+        (
+            [ClassificationLabel(0), ClassificationLabel(1)],
```
ly-kolena, line 451: Should this have a single class?

ly-kolena: Misunderstood, resolved

y27choi:
I realized I can't use the util function in binary classification evaluator in this PR as the util function is not available in the version of kolena client that it's importing in the example, so reverted that part.  Will put up a separate PR for the example change once we release this tonight.

kolena/_experimental/classification/utils.py
```diff
@@ -251,3 +254,75 @@ class is `None`.
...
+    """
+    Computes scores (i.e. [`Precision`][kolena.workflow.metrics.precision], [`Recall`](kolena.workflow.metrics.recall)
+    and [`F1-score`](kolena.workflow.metrics.f1_score)) vs. threshold curves for a **single** class presented in
+    `inferences`.
+
+    :param ground_truths: The list of ground truth
+        [`ClassificationLabel`][kolena.workflow.annotation.ClassificationLabel]. For binary classification, the negative
+        class can be `None`.
+    :param inferences: The list of inference
+        [`ScoredClassificationLabel`][kolena.workflow.annotation.ScoredClassificationLabel]. The list should only
```
munkyshi, line 273:
```suggestion
        [`ScoredClassificationLabel`][kolena.workflow.annotation.ScoredClassificationLabel]s. The list should only
```

kolena/_experimental/classification/utils.py
```diff
@@ -251,3 +254,75 @@ class is `None`.
...
+    inferences: List[ScoredClassificationLabel],
+    thresholds: Optional[List[float]] = None,
+) -> Optional[List[CurvePlot]]:
+    """
+    Computes scores (i.e. [`Precision`][kolena.workflow.metrics.precision], [`Recall`](kolena.workflow.metrics.recall)
+    and [`F1-score`](kolena.workflow.metrics.f1_score)) vs. threshold curves for a **single** class presented in
+    `inferences`.
+
+    :param ground_truths: The list of ground truth
+        [`ClassificationLabel`][kolena.workflow.annotation.ClassificationLabel]. For binary classification, the negative
```
munkyshi, line 270:
```suggestion
        [`ClassificationLabel`][kolena.workflow.annotation.ClassificationLabel]s. For binary classification, the negative
```

kolena/_experimental/classification/utils.py
```diff
@@ -251,3 +254,75 @@ class is `None`.
...
+        class can be `None`.
+    :param inferences: The list of inference
+        [`ScoredClassificationLabel`][kolena.workflow.annotation.ScoredClassificationLabel]. The list should only
+         include inferences of a specific class to plot the threshold curves for.
+    :param thresholds: The list of thresholds to plot with. If not specified, all the unique confidence scores are used
```
ly-kolena, line 275: Is there order assumption for this param?

kolena/_experimental/classification/utils.py
```diff
@@ -251,3 +254,75 @@ class is `None`.
...
+
+    inference_label = {inf.label for inf in inferences}
+    if len(inference_label) > 1:
+        log.warn(
+            f"more than one class passed in as inferences: {inference_label} — expecting inferences belonging to "
+            "a single class",
+        )
+        return None
+
+    inference_label = next(iter(inference_label))
```
munkyshi, line 300: nit: not a fan of this type change for the variable -- makes things confusing to read

munkyshi (approved):
A few nits, nothing blocking.

kolena/_experimental/classification/utils.py
```diff
@@ -251,3 +254,75 @@ class is `None`.
             curves=curves,
         )
     return None
+
+
+def compute_threshold_curves(
+    ground_truths: List[Optional[ClassificationLabel]],
+    inferences: List[ScoredClassificationLabel],
```
munkyshi, line 261: Is it expected that ground truths and inferences correspond to the same sample (for the same given index)? If so, should probably be included in the documentation.

kolena/_experimental/classification/utils.py
```diff
@@ -251,3 +254,75 @@ class is `None`.
...
+
+    inference_label = next(iter(inference_label))
+    y_true = [gt.label == inference_label if gt else False for gt in ground_truths]
+
+    if thresholds is None:
+        unique_scores = list({inf.score for inf in inferences})
+        initial_thresholds = list(np.linspace(0, 1, 11))
+        thresholds = [0.0]
+        for threshold in sorted(set(initial_thresholds + unique_scores)):
+            if abs(threshold - thresholds[-1]) >= 1e-2:
```
ly-kolena, line 308: Nit: does this somehow contradict

> If not specified, all the unique confidence scores are used

or is it commonly understood practice?

kolena/_experimental/classification/utils.py
```diff
@@ -251,3 +254,75 @@ class is `None`.
...
+                thresholds.append(threshold)
+
+    precisions = []
+    recalls = []
+    f1s = []
+    for threshold in thresholds:
+        y_pred = [inf.score >= threshold for inf in inferences]
+        tp = len([True for t, p in zip(y_true, y_pred) if t and p])
```
ly-kolena, line 316: Nit: what does `t` and `p` mean?

kolena/_experimental/classification/utils.py
y27choi: Thank you for flaggin this.  I will sort it in the function and add a unit test for this.

kolena/_experimental/classification/utils.py
y27choi: This is because of the float precision — not guaranteed to filter out "duplicate" floats in the set (e.g., 0.2 vs. 0.200000001)

kolena/_experimental/classification/utils.py
y27choi: y_**t**rue and y_**p**red but I will update them to something better than these.

kolena/_experimental/classification/utils.py
ly-kolena: would this unintentionally drop thresholds smaller than `0.01` diff? I would think `0.01` is still valid difference if intentional?

y27choi: It is intentional to keep the diff as small as `0.01` because we don't want to plot too many points.  This logic is only applied to the default thresholds.  If users want to use specific set of thresholds with `< 0.01` diff then that can be done using the `thresholds` parameter.

PR #180, opened 2023-08-08
Keypoints workflow improvements

munkyshi:
### Linked issue(s):
N/A

Improve keypoints example implementation to better match the provided data.

- Update orderings to recognize that the 0th point in the 300-W sample dataset is the nose
- Use a log axis for AFR vs Threshold plot
- Use a NMSE calculation that actually normalizes distance with the normalization factor
- Rename fields that referenced `ear` to `eye`

### DIFF
--- a/examples/keypoint_detection/keypoint_detection/evaluator.py
+++ b/examples/keypoint_detection/keypoint_detection/evaluator.py
@@ -55,12 +55,12 @@ class KeypointsEvaluator(Evaluator):
             inference.face.points[0],
             normalization_factor,
         )
-        Δ_left_ear, norm_Δ_left_ear = compute_distances(
+        Δ_left_eye, norm_Δ_left_eye = compute_distances(
             ground_truth.face.points[1],
             inference.face.points[1],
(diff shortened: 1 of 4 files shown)


examples/keypoint_detection/keypoint_detection/evaluator.py
```diff
@@ -49,17 +50,17 @@ def compute_test_sample_metrics_single(
             return TestSampleMetrics(match_type="failure_to_detect")
 
         normalization_factor = ground_truth.normalization_factor
-        Δ_left_ear, norm_Δ_left_ear = compute_distances(
+        Δ_nose, norm_Δ_nose = compute_distances(
```
ly-kolena, line 53: Was the metrics incorrect before, i.e. value and name/meaning mismatch?

munkyshi: I believe so. We don't have an example of this up on any of our main tenants, so I wasn't able to verify this.

ly-kolena (approved):

PR #193, opened 2023-08-17
Change text summarization example with faster remote evaluator

marko-polo-cheno:
Remote evaluator is slow due to recomputing test sample metrics. Rather than recomputing, the script was changed to pull test sample metrics from CSVs.

- Changed `pyproject.toml` since torch download was failing
- Refactored helper functions into `utils`
- Made `evaluator_fast` to feed into the remote evaluator
- Cleaned up `workflow.py`

Remote evaluator works: [link](https://trunk.kolena.io/mark/testing?testSuiteId=24&testCaseId=106)

### DIFF
--- a/examples/text_summarization/text_summarization/workflow.py
+++ b/examples/text_summarization/text_summarization/workflow.py
@@ -42,13 +42,11 @@ class GroundTruth(BaseGroundTruth):
 @dataclass(frozen=True)
 class Inference(BaseInference):
     summary: str
+    source: str
     word_count: int
     is_failure: bool = False
     inference_time: Optional[float] = None
-    tokens_input_text: Optional[int] = None
-    tokens_ref_summary: Optional[int] = None
-    tokens_pred_summary: Optional[int] = None
-    tokens_prompt: Optional[int] = None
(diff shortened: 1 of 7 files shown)


examples/text_summarization/text_summarization/workflow.py
```diff
@@ -42,13 +42,11 @@ class GroundTruth(BaseGroundTruth):
 @dataclass(frozen=True)
 class Inference(BaseInference):
     summary: str
+    source: str
     word_count: int
     is_failure: bool = False
     inference_time: Optional[float] = None
-    tokens_input_text: Optional[int] = None
```
ly-kolena, line 48: what's the reason for the removal?

marko-polo-cheno: For demo/example purposes, one token count provides more clarity than 4 different counts. More importantly, the CSVs containing the precomputed metrics do not have the other token counts.

gordonhart:
This doesn't put our best foot forward as a publicly-visible example — hesitant to merge because it significantly impacts the readability of this example.

PR #194, opened 2023-08-18
Add configuration as key to matching cache for prebuilt OD

marko-polo-cheno:
Having multiple configurations in an evaluation leads to dup/triple/n- copies of cached matchings, leading to 2x/3x/Nx values on the results view.

By adding the configuration name as a key to the cache, there is no longer an issue with having multiple configs.


tests/integration/_experimental/object_detection/test_evaluator_single_class_f1_optimal.py
```diff
@@ -272,7 +272,8 @@ def test__object_detection__multiclass_evaluator__f1_optimal() -> None:
     assert config.display_name() in eval.evaluator.threshold_cache
     assert eval.evaluator.threshold_cache[config.display_name()] == 0.1
     assert len(eval.evaluator.matchings_by_test_case) != 0
-    assert len(eval.evaluator.matchings_by_test_case[TEST_CASE.name]) == len(TEST_DATA)
+    assert len(eval.evaluator.matchings_by_test_case[config.display_name()]) != 0
```
ly-kolena, line 275: Nitpicking: is it easy to actual use 2 configurations which would lead to different caching results and verify correctness?

ly-kolena (approved):

sandalns (approved):

PR #196, opened 2023-08-21
chore: poc of data ingestion

yifanwu-kolena:


kolena/_experimental/data_ingestion/_data_format/coco.py
```diff
@@ -0,0 +1,126 @@
...
+        for annotation in raw_data["annotations"]:
+            image_id = annotation["image_id"]
+            label = label_map[annotation["category_id"]]
+
+            # check that the box is in a valid image, not a crowd box
+            if image_id in ids and not annotation["iscrowd"]:
+                bbox = annotation["bbox"]
+                top_left = (bbox[0], bbox[1])
+                bottom_right = (bbox[0] + bbox[2], bbox[1] + bbox[3])
+                bounding_box = LabeledBoundingBox(top_left, bottom_right, label)
```
gordonhart, line 102: COCO JSON also includes additional attributes inside bboxes (and can be extended with more attributes) — these should make their way into Kolena.

kolena/_experimental/data_ingestion/_data_format/coco.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def _get_bbox_map(self, raw_data, image_map, label_map):
+        ids = set(image_map.keys())
+        image_to_boxes: Dict[int, List[LabeledBoundingBox]] = defaultdict(lambda: [])
+        for annotation in raw_data["annotations"]:
+            image_id = annotation["image_id"]
+            label = label_map[annotation["category_id"]]
+
+            # check that the box is in a valid image, not a crowd box
+            if image_id in ids and not annotation["iscrowd"]:
```
gordonhart, line 98: Not sure we should just silently drop `iscrowd` bounding boxes.

yifanwu-kolena: yeah, we definitely can handle `iscrowd` bbox. i just copied a lot of logic from https://github.com/kolenaIO/kolena/tree/trunk/examples/object_detection_2d for the poc purpose

kolena/_experimental/data_ingestion/_data_format/coco.py
```diff
@@ -0,0 +1,255 @@
...
+    def _get_infer(
+        self,
+        inf_map: Dict[int, List[ScoredLabeledBoundingBox]],
+    ) -> Callable[[TestSample], Inference]:
+        # a function that creates an inference from a test sample
+        def infer(sample: TestSample) -> Inference:
+            try:
+                locator = sample.locator
+                # TODO: maybe we should save the id for test samples
+                image_id = self._get_image_id(locator)
```
yifanwu-kolena: this is fragile, the locator based on `file_name` might not match with image id that is used in the inference coco json, we should add id field in the pre-build test samples, but it might be a breaking change

munkyshi:
@yifanwu-kolena what's a good example usage for this script -- I'd like to try it out locally. Do I need to download any files?

kolena/_experimental/data_ingestion/_data_format/coco.py
```diff
@@ -0,0 +1,255 @@
...
+        model = Model(model_name, infer=self._get_infer(self.inf_map))
+
+        # customizable configurations for the evaluator
+        evaluator = self._setup_evaluator()
+
+        # runs the evaluation
+        test_suite = TestSuite(test_suite_name)
+        test(model, test_suite, evaluator, reset=reset)
+
+    def _get_image_id(self, locator: str) -> int:
```
munkyshi: This seems very fragile -- I can't imagine expecting customer data (set up as COCO JSON data) to adhere to this type of structure.

kolena/_experimental/data_ingestion/_data_format/coco.py
munkyshi: Definitely very fragile -- I'd be a fan of saving lookup information in the sample itself, to help with resolution. Can we not declare an extended version of `TestSample` that includes `coco_image_id` as a field? I thought we were able to get extensions of the base class to work (with customer data).

munkyshi: An alternative is ingesting test suite and results at the same time, to ensure that we're dealing with the same internal ids in the pair of files.

munkyshi (commented):
Thanks for the example -- this gives a good sense of what a COCO ingester would look like.

What are next steps here?
- Handle crowd
- Better ID resolution logic
- Include additional bbox attributes

Are we expecting to merge this PR, or was this meant to be more of an example?

kolena/_experimental/data_ingestion/_data_format/coco.py
yifanwu-kolena: thanks for the ideas, seems like [extending the class](https://github.com/kolenaIO/kolena/pull/196/commits/b66528bbe172d35929f19b8c539cffd3105d300d) can fix this issue

yifanwu-kolena:
@munkyshi based on our syncing this morning, we will incubate the coco ingestion all in [galleon side](https://github.com/kolenaIO/shipyard/pull/1557), which will speed up the iteration without waiting release of kolena or worrying about any potential interface changes.

the next step will be solidify the backend endpoint and connect with FE. we will work on handing `crowd` and adding additional bbox attribute later after the whole flow is setup.

and will move all of the reusable logic back to this repo when the data ingestion pattern is stable.

gordonhart:
Can this be closed?

PR #201, opened 2023-08-23
Dataset APIs - data ingestion

ly-kolena:
### Linked issue(s):

Example codes converted from object_detection_2d example script

```
def load_transportation_data() -> Dict[str, List[LabeledBoundingBox]]:
    # same codes from example
    ...


def complete_datapoints(args: Namespace) -> pd.DataFrame:
    print("loading S3 files...")
    image_to_boxes = load_transportation_data()
    data = pd.read_csv(args.metadata)
    dataset = data[["width", "height", "brightness"]]
    dataset["locator"] = S3_IMAGE_LOCATION + data["name"]
    dataset["bboxes"] = data["name"].apply(lambda x: image_to_boxes[x])

    return dataset


def main(args: Namespace) -> None:
    kolena.initialize(os.environ["KOLENA_TOKEN"], verbose=True)

    register_dataset(DATASET, complete_datapoints(args))
    df_loaded = fetch_dataset(DATASET)
    print(df_loaded.shape)


if __name__ == "__main__":
    ap = ArgumentParser()
    ap.add_argument(
        "--metadata",
        type=str,
        default="s3://kolena-public-datasets/coco-2014-val/meta/metadata_attribution2.0_brightness.csv",
        help="The path to the image metadata stored in a csv.",
    )

    main(ap.parse_args())
```


ly-kolena:
cc @gordonhart @munkyshi for early review

kolena/_experimental/dataset/dataset.py
```diff
@@ -0,0 +1,134 @@
+# Copyright 2021-2023 Kolena Inc.
```
ly-kolena: Any suggestion better than `dataset.dataset`?

kolena/workflow/_datatypes.py
```diff
@@ -231,6 +232,78 @@ def deserialize_field(
                     items[key] = _try_deserialize_typed_dataobject(val)
         return cls(**items)
 
+    def toDict(self) -> Dict:
```
ly-kolena: Should be gone after https://github.com/kolenaIO/kolena/pull/239 is merged

tests/unit/workflow/test_data_object.py
```diff
@@ -234,3 +240,52 @@ class Tester(DataObject):
     assert deserialized.a.extra == "foo"
     assert deserialized.b == [data_object, data_object]
     assert deserialized.b[1].extra == "foo"
+
```
ly-kolena: Should be gone after #239 is merged

y27choi:
Do we want to return the dataset object from `register_dataset(DATASET, complete_datapoints(args))`?

kolena/_experimental/dataset/dataset.py
y27choi: can we add `register_dataset()` and `fetch_dataset()` to `__init__.py` so we can import them by `from kolena._experimental.dataset import register_dataset, fetch_dataset`?

ly-kolena:
> Do we want to return the dataset object from `register_dataset(DATASET, complete_datapoints(args))`?

Since we tried to avoid introducing a new `Dataset` object, and went with operating on `DataFrame`, the server response was not passed back to users in this implementation. Users currently cannot do anything with this return. Do you think it'd be useful if a plain `Dict` is returned back from `register_dataset`?

kolena/_experimental/dataset/dataset.py
ly-kolena: updated as suggested

y27choi:
> Do you think it'd be useful if a plain Dict is returned back from register_dataset? 

Probably not. I was thinking that I'd need the dataset object to run tests but I just need the dataset name to kick off a run, right? In that case, I don't think I would actually need the dataset object for the happy path. In which scenarios do you think I would ever need the object?

ly-kolena:
> Probably not. I was thinking that I'd need the dataset object to run tests but I just need the dataset name to kick off a run, right? In that case, I don't think I would actually need the dataset object for the happy path. In which scenarios do you think I would ever need the object?

Current thinking is (hopefully) users can do all operations with minimal info needed, so dataset name (and model name) would be all they need in the most common cases to send data to kolena.

y27choi:
And things like tags, description are something users would need to add from the frontend?

ly-kolena:
> And things like tags, description are something users would need to add from the frontend?

That is the expectation thus far.

y27choi:
The interface for dataset looks very simple which I like.  I will play around with different annotation types / cell types later but for now it's 👍 from me.

kolena/_experimental/dataset/_dataset.py
```diff
@@ -0,0 +1,142 @@
...
+    upload_data_frame(df=df_serialized, batch_size=BatchSize.UPLOAD_RECORDS.value, load_uuid=load_uuid)
+    request = RegisterRequest(name=name, uuid=load_uuid)
+    response = krequests.post(Path.REGISTER, json=asdict(request))
+    krequests.raise_for_status(response)
+
+
+def iter_dataset(
+    name: str,
+    batch_size: int = BatchSize.LOAD_SAMPLES.value,
+) -> Iterator[pd.DataFrame]:
```
munkyshi, line 129: I would expect `iter_dataset` to iterate over datapoints, not over data frames -- IMO it's an awkward interface with data frames.

munkyshi:
> > And things like tags, description are something users would need to add from the frontend?
> 
> That is the expectation thus far.

I would expect programmatic tags to be useful, especially if we're using tags in place of workflows (to indicate which remote evaluators are compatible)

munkyshi (approved):
I'd like to reach an agreement on https://github.com/kolenaIO/kolena/pull/201/files/5b4b94915192dd7732504b69d19e1647ba43a504#diff-2b897e1ce619a688e8b9627675e7d0f5509b44518717c29b2cb7755a9b7993f8 before merging, but everything else looks good. Approving to not block on merge after this gets sorted out.

munkyshi (approved):
Synced with @ly-kolena offline -- I'm convinced to move forward here as is. We can work through programmatic tags separately in the future, depending on validation on this approach from @y27choi's prototyping.

kolena/_experimental/dataset/_dataset.py
```diff
@@ -0,0 +1,142 @@
...
+COL_DATAPOINT = "datapoint"
+TEST_SAMPLE_TYPE = "TEST_SAMPLE"
+FIELD_LOCATOR = "locator"
+FIELD_TEXT = "text"
+
+_DATAPOINT_TYPE_MAP = {
+    "image": f"{TEST_SAMPLE_TYPE}/IMAGE",
+    "application/pdf": f"{TEST_SAMPLE_TYPE}/DOCUMENT",
+    "text": f"{TEST_SAMPLE_TYPE}/DOCUMENT",
+    "video": f"{TEST_SAMPLE_TYPE}/VIDEO",
```
dylangrandmont: nit: It would make sense to declare the enum of possible types directly and reference that here for the map values. We do something similar in `marina` (with typescript):

```ts
export enum TestSampleType {
  IMAGE = 'TEST_SAMPLE/IMAGE',
  COMPOSITE = 'TEST_SAMPLE/COMPOSITE',
  TEXT = 'TEST_SAMPLE/TEXT',
  VIDEO = 'TEST_SAMPLE/VIDEO',
  DOCUMENT = 'TEST_SAMPLE/DOCUMENT',
  POINT_CLOUD = 'TEST_SAMPLE/POINT_CLOUD',
}
```

dylangrandmont: Likewise below, for instance [here](https://github.com/kolenaIO/kolena/pull/201/files#diff-2b897e1ce619a688e8b9627675e7d0f5509b44518717c29b2cb7755a9b7993f8R67).

tests/unit/_experimental/dataset/test_dataset.py
```diff
@@ -0,0 +1,175 @@
...
+from kolena._experimental.dataset._dataset import TEST_SAMPLE_TYPE
+from kolena.workflow.annotation import BoundingBox
+from kolena.workflow.annotation import LabeledBoundingBox
+
+CUSTOM = f"{TEST_SAMPLE_TYPE}/CUSTOM"
+IMAGE = f"{TEST_SAMPLE_TYPE}/IMAGE"
+VIDEO = f"{TEST_SAMPLE_TYPE}/VIDEO"
+DOCUMENT = f"{TEST_SAMPLE_TYPE}/DOCUMENT"
+TEXT = f"{TEST_SAMPLE_TYPE}/TEXT"
+POINT_CLOUD = f"{TEST_SAMPLE_TYPE}/POINT_CLOUD"
```
dylangrandmont: [Declaring an enum](https://github.com/kolenaIO/kolena/pull/201/files#r1337314296) would also remove the need for this to be declared in a test file

kolena/_experimental/dataset/_dataset.py
```diff
@@ -0,0 +1,142 @@
...
+    return _DATAPOINT_TYPE_MAP.get(mimetype_str, None) or _DATAPOINT_TYPE_MAP.get(main_type, None)
+
+
+def _infer_datatype_value(x: str) -> str:
+    mtype, _ = mimetypes.guess_type(x)
+    if mtype:
+        datatype = _get_datapoint_type(mtype)
+        if datatype is not None:
+            return datatype
+    elif x.endswith(".pcd"):
```
dylangrandmont, line 77: Is this not covered by `mimetypes.guess_type`?

kolena/_experimental/dataset/_dataset.py
```diff
@@ -0,0 +1,142 @@
...
+
+COL_DATAPOINT = "datapoint"
+TEST_SAMPLE_TYPE = "TEST_SAMPLE"
+FIELD_LOCATOR = "locator"
+FIELD_TEXT = "text"
+
+_DATAPOINT_TYPE_MAP = {
+    "image": f"{TEST_SAMPLE_TYPE}/IMAGE",
+    "application/pdf": f"{TEST_SAMPLE_TYPE}/DOCUMENT",
+    "text": f"{TEST_SAMPLE_TYPE}/DOCUMENT",
```
dylangrandmont: Potentially this should map to `TEST_SAMPLE/TEXT`, which `marina` already supports. Which mime types will this include?

ly-kolena: How does marina handle a test sample with `locator: https://example.com/test.xml`? The `text` type here is for this case, not a test sample using `text: "my qa question"`.

kolena/_experimental/dataset/_dataset.py
ly-kolena: It does not. For reference, [mimetype](https://docs.python.org/3/library/mimetypes.html#mimetypes.guess_type) package references [this list ](https://www.iana.org/assignments/media-types/media-types.xhtml)

kolena/_experimental/dataset/_dataset.py
ly-kolena: updated as suggested

yifanwu-kolena (approved):

kolena/_experimental/dataset/_dataset.py
```diff
@@ -0,0 +1,153 @@
...
+from kolena.workflow._datatypes import DATA_TYPE_FIELD
+from kolena.workflow._datatypes import TypedDataObject
+
+COL_DATAPOINT = "datapoint"
+TEST_SAMPLE_TYPE = "TEST_SAMPLE"
+FIELD_LOCATOR = "locator"
+FIELD_TEXT = "text"
+
+
+class TestSampleType(str, Enum):
```
dylangrandmont, line 44: We should avoid all use of the terminology "test sample" and instead be using "datapoint"

PR #204, opened 2023-08-23
Link numpy definitions in API docs

gordonhart:
Add `numpy/objects.inv` to extra imports for docs rendering such that any `np` definitions in signatures are linked to the respective official documentation:

| [Before](https://docs.kolena.io/reference/workflow/visualization/#kolena.workflow.visualization.encode_png) | After |
| ---- | --- |
| Screen Shot 2023-08-23 at 3 36 54 PM | Screen Shot 2023-08-23 at 3 36 43 PM |

### DIFF
--- a/kolena/workflow/test_run.py
+++ b/kolena/workflow/test_run.py
@@ -89,7 +89,7 @@ class TestRun(Frozen, WithTelemetry, metaclass=ABCMeta):
 
     model: Model
     test_suite: TestSuite
-    evaluator: Union[Evaluator, BasicEvaluatorFunction, None]
+    evaluator: Optional[Union[Evaluator, BasicEvaluatorFunction]]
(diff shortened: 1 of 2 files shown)


kolena/workflow/test_run.py
```diff
@@ -89,15 +89,15 @@ class TestRun(Frozen, WithTelemetry, metaclass=ABCMeta):
 
     model: Model
     test_suite: TestSuite
-    evaluator: Union[Evaluator, BasicEvaluatorFunction, None]
+    evaluator: Optional[Union[Evaluator, BasicEvaluatorFunction]]
```
gordonhart, line 92: @munkyshi was there a reason this was previously a `Union[..., None]` instead of an `Optional[Union[...]]`? If so I will revert these drive-bys.

ly-kolena (approved):

PR #206, opened 2023-08-24
Remove old pre-built multiclass classification workflow

gordonhart:
Replaced with [example](https://github.com/kolenaIO/kolena/tree/trunk/examples/classification) and [pre-built utilities](https://docs.kolena.io/reference/pre-built/classification/) in the near-term.


munkyshi (approved):
Just confirmed that there isn't any external usage of this outside of the single example we built.

ly-kolena (approved):

PR #210, opened 2023-08-25
Bump to new cache

ly-kolena:

### DIFF
--- a/kolena/_api/v1/core.py
+++ b/kolena/_api/v1/core.py
@@ -192,3 +192,7 @@ class TestRun:
     @dataclass(frozen=True)
     class MarkCrashedRequest:
         test_run_id: int
+
+
+class Dummy:
+    ...
--- a/kolena/_api/v1/core.py
+++ b/kolena/_api/v1/core.py
@@ -192,7 +192,3 @@ class TestRun:
     @dataclass(frozen=True)
     class MarkCrashedRequest:
         test_run_id: int
-
-
-class Dummy:
-    ...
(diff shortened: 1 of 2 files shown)


gordonhart:
🤔 How did this fix the cache issue?

ly-kolena:
> 🤔 How did this fix the cache issue?

```
      - restore_cache:
          key: &ci-base-cache ci-cache-<< parameters.python-version >>-<< parameters.extras >>-{{ checksum "pyproject.toml" }}
```
checksum changed, so different cache (re-generated)

(This bypass troubleshooting how the previous cache bot corrupted)

gordonhart:
I'm not a huge fan of landing a code change to fix the CircleCI cache.

Is CI caching still necessary? I know that updating our Poetry version to the new parallel installer a few months ago sped setup up by 5x — can we get away with a less complex CI setup?

munkyshi:
A warm cache seems to save ~3 min on total runtime, and ~30 min of machine runtime. 9 min total CI time isn't too bad (though 6 is much nicer). 30 min machine runtime could add up in terms of cost on the CI side of things, but might not be worth it if we have bad caches without a clean way of invalidation.

Do we know why the cache was bad here?

yifanwu-kolena:
[tried to display the env info in CI](https://github.com/kolenaIO/kolena/pull/213) and results are interesting, some of them are valid, some of them are not, even for the same restored cache...

gordonhart:
While the underlying issue of flaky and immutable cache is still there, it's safe to close this PR.

PR #216, opened 2023-08-29
allow use of system preference for dark mode

dylangrandmont:
### Linked issue(s):

mkdocs allows the user rely on their system preference for dark/light mode (insider feature).

https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/#automatic-light-dark-mode

Fixes https://linear.app/kolena/issue/KOL-3341/automatically-change-lightdark-mode-on-the-doc-based-on-the-kolena (with the caveat that Kolena theme and this theme are not strictly synchronized, but at least they are _more likely_ to be synchronized)

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -82,19 +82,22 @@ theme:
   font:
     text: Inter
   palette:
+    # Palette toggle for automatic mode
+    - media: "(prefers-color-scheme)"
+      toggle:
+        icon: material/brightness-auto
+
+    # Palette toggle for light mode
     - media: "(prefers-color-scheme: light)"
       scheme: default
-      accent: cyan
       toggle:
-        icon: kolena/light-mode-20
-        name: "Switch to dark mode"
+        icon: material/brightness-7
+
+    # Palette toggle for dark mode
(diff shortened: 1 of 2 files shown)


mkdocs.yml
```diff
@@ -82,19 +82,22 @@ theme:
...
+      toggle:
+        icon: material/brightness-auto
+
+    # Palette toggle for light mode
     - media: "(prefers-color-scheme: light)"
       scheme: default
-      accent: cyan
       toggle:
-        icon: kolena/light-mode-20
-        name: "Switch to dark mode"
```
dylangrandmont, line 90: I have removed these `name`s here. mkdocs appears to have a bug where the wrong label is being rendered for this toggle when there are three items

y27choi:
Should Kolena platform use the system preference as well?

y27choi (approved):
Tested it out locally and it seems to work as expected 👍

dylangrandmont:
> Should Kolena platform use the system preference as well?

We do (if the user chooses "Auto" for their theme)!

PR #220, opened 2023-08-30
Create CoQA Evaluator

marko-polo-cheno:
Fixes KOL-2936


examples/question_answering/pyproject.toml
```diff
@@ -6,10 +6,18 @@ authors = ["Kolena Engineering <eng@kolena.io>"]
 license = "Apache-2.0"
 
 [tool.poetry.dependencies]
-python = ">=3.8,<3.12"
+python = ">=3.9,<3.10"
```
munkyshi: Is this because of the `torch` handling?

examples/question_answering/pyproject.toml
```diff
@@ -6,10 +6,18 @@ authors = ["Kolena Engineering <eng@kolena.io>"]
...
 kolena = ">=0.82.0,<1"
 s3fs = "^2022.7.1"
+torch = [
+  {markers = "sys_platform == 'darwin' and platform_machine == 'arm64'", url = "https://download.pytorch.org/whl/cpu/torch-2.0.1-cp39-none-macosx_11_0_arm64.whl"},
+  {markers = "sys_platform == 'darwin' and platform_machine == 'x86_64'", url = "https://download.pytorch.org/whl/cpu/torch-2.0.1-cp39-none-macosx_10_9_x86_64.whl"},
+  {markers = "sys_platform == 'linux' ", url = "https://download.pytorch.org/whl/cpu/torch-2.0.1%2Bcpu-cp39-cp39-linux_x86_64.whl"},
+  {markers = "sys_platform == 'win32' ", url = "https://download.pytorch.org/whl/cpu/torch-2.0.1%2Bcpu-cp39-cp39-win_amd64.whl"},
+]
```
munkyshi: What's the behavior if you don't have this block? `poetry add torch` seems to work on my mac, is this failing on other system?

examples/question_answering/pyproject.toml
marko-polo-cheno: Yes this was for torch handling

examples/question_answering/pyproject.toml
marko-polo-cheno: I'll give that a try. I believe we want to force it to only use the CPU version of torch though.
Without that block, on an M1 mac, it seems like I don't need to include torch, but CI will complain about it.

examples/question_answering/question_answering/evaluator.py
```diff
@@ -0,0 +1,202 @@
...
+    def populate_cache(self, file_path: str) -> None:
+        df = pd.read_csv(file_path)
+        metrics = ["BERT_prec", "BERT_rec", "BERT_f1", "ROUGE_1", "ROUGE_2", "ROUGE_L"]
+
+        for _, row in df.iterrows():
+            data_id = row["data_id"]
+            turn = row["turn"]
+            values_dict = {col: round(row[col], 6) for col in metrics}
+
+            self.precomputed_metrics.setdefault(data_id, {})[turn] = values_dict
```
gordonhart: We shouldn't use precomputed metrics for these example integrations.

munkyshi (approved):

examples/question_answering/question_answering/evaluator.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+class QuestionAnswerEvaluator(Evaluator):
+    """
+    The `QuestionAnswerEvaluator` transforms inferences into metrics for the question answer workflow.
+
+    For additional functionality, see the associated [base class documentation][kolena.workflow.evaluator.Evaluator].
+    """
+
+    test_samples_by_test_case: Dict[str, Dict[str, List[int]]] = {}  # test_case -> data_id -> turns
```
y27choi: Do you think we can avoid this cacheing if we use evaluate function instead?

examples/question_answering/question_answering/seed_test_run.py
```diff
@@ -0,0 +1,85 @@
...
+    model = Model(args.model, infer=infer)
+    test_suite = TestSuite.load(args.test_suite)
+    evaluator = QuestionAnswerEvaluator(configurations)
+    test(model, test_suite, evaluator, reset=True)
+    return 0
+
+
+if __name__ == "__main__":
+    ap = ArgumentParser()
+    ap.add_argument("test_suite", help="Name of the test suite to test.")
```
y27choi: Should we have a default value so simply running `poetry run python seed_test_run.py` can give you a test suite / test run?

examples/question_answering/question_answering/evaluator.py
marko-polo-cheno: It could be avoided, but I will leave it as is

y27choi: Aside from avoiding caching, I think we can benefit for using evaluate function on not repeating computing test sample metrics. It's ok if you can't get to this today. I can clean it up and wrap it up.

y27choi:
Wrapping up Mark's opened PR and here are the updates I made:
 * Changed the evaluator class to evaluator function
 * Added a tqdm progress bar
 * Reduced the default evaluator config from 2 to 1 for a better default experience
 * Added the example to the example README.md and quickstart guide.

@munkyshi would appreciate another round of review 🙏

y27choi (approved):

munkyshi (approved):

PR #219, opened 2023-08-29
Patch integration tests

marko-polo-cheno:
Patch #208

PR #208, opened 2023-08-24
Change Prebuilt OD Configuration and Test Case Metrics

marko-polo-cheno:
Fixes KOL-2846

PR #221, opened 2023-08-30
Fixed prebuilt OD tests with new config and metrics

marko-polo-cheno:
Restore #208 with #219 patch - no other changes outside of `.toml`


tests/integration/_experimental/object_detection/test_evaluator_multiclass_f1_optimal.py
```diff
@@ -487,47 +467,47 @@
...
+                    "Precision": [1, 1, 1, 1, 1, 1, 1, 1],
+                    "Recall": [10 / 13, 8 / 13, 7 / 13, 6 / 13, 5 / 13, 4 / 13, 3 / 13, 1 / 13],
+                },
             ),
-            Curve(x=[0.1, 0.6, 0.7, 0.8], y=[8 / 15, 3 / 7, 4 / 13, 1 / 6], label="e"),
-        ],
-        x_config=None,
-        y_config=None,
-    ),
-    CurvePlot(
```
ly-kolena, line 509: Can you point me to the change that would lead to different plots being created in this PR? Looks to me it used to have 2 CurvePlots but now only have 1.

tests/integration/_experimental/object_detection/test_evaluator_multiclass_fixed.py
```diff
@@ -860,36 +813,41 @@ def assert_curves(
...
     assert len(plots) == len(other_plots)
-    # check curve plots
-    for plot, expected in zip(plots[:2], other_plots[:2]):
-        assert plot.title == expected.title
-        assert plot.x_label == expected.x_label
-        assert plot.y_label == expected.y_label
-        assert_curves(plot.curves, expected.curves)
-        assert plot.x_config == expected.x_config
-        assert plot.y_config == expected.y_config
+    # check one plot (curve.extra contains other plot)
```
ly-kolena: Any reason we don't verify plots[1], even if `curve.extra` for first plot contain similar info?

tests/integration/_experimental/object_detection/test_evaluator_multiclass_f1_optimal.py
marko-polo-cheno: `assert_test_case_plots_equals_expected` was changed to ignore the second plot since covered by unit tests. I replaced the 2nd Curveplot with a placeholder.

tests/integration/_experimental/object_detection/test_evaluator_multiclass_fixed.py
marko-polo-cheno: Since the first plot contains the 2nd plot's points, and there's no need to go into detail like a unit test, I've left it out for simplicity.

tests/integration/_experimental/object_detection/test_evaluator_multiclass_f1_optimal.py
ly-kolena: I would prefer still verifying it or update verification logic so that we don't need a placeholder of unused fake data. It is a very confusing way to construct test data and to understand.

ly-kolena (approved):

y27choi (approved):
It took me a while to understand the CurvePlot change.  Changes look good to me 👍

PR #229, opened 2023-09-06
KOL-3377 Add FR 1:1 example integration to kolenaIO/kolena

ghost:
### Linked issue(s):
Linear ID: KOL-3377
https://linear.app/kolena/issue/KOL-3377/add-fr-11-example-integration-to-kolenaiokolena

- Add an example integration implementing the [NIST FRVT 1:1](https://pages.nist.gov/frvt/html/frvt11.html) evaluation standard as a pre-built workflow.
  - New directory examples/face_recognition_11 inside the kolena repository
  - Integration follows the structure of the other examples in this repo
  - Defines a workflow and implements an evaluator that has feature parity with legacy [kolena.fr](http://kolena.fr/) ([trunk/kolena](https://trunk.kolena.io/kolena/results/fr?modelIds=730&testSuiteId=1259) example)
  - Uses an open-source dataset (e.g. Labeled Faces in the Wild) such that any user can get up and running with this problem


ghost:
Good for review 😭

y27choi:
Do you have any example result to link?

ghost:
> Do you have any example result to link?

https://trunk.kolena.io/andy/results?modelIds=2&workflow=Face+Recognition+1%3A1&testSuiteId=8&evaluators=N4IgHCBcoExQ7GAvkoA

examples/face_recognition_11/README.md
```diff
@@ -0,0 +1,52 @@
...
+The data for this example integration lives in the publicly accessible S3 bucket `s3://kolena-public-datasets`.
+
+First, ensure that the `KOLENA_TOKEN` environment variable is populated in your environment. See our
+[initialization documentation](https://docs.kolena.io/installing-kolena/#initialization) for details.
+
+This project defines two scripts that perform the following operations:
+
+1. [`seed_test_suite.py`](face_recognition_11/seed_test_suite.py) creates a test case with all test samples in a test suite
+
+2. [`seed_test_run.py`](face_recognition_11/seed_test_run.py) tests a random keypoint generator against the test suite above
```
y27choi: `tests a random keypoint generator against the test suite above` needs to be updated

examples/face_recognition_11/README.md
```diff
@@ -0,0 +1,52 @@
...
+  -h, --help  show this help message and exit
+```
+
+```shell
+$ poetry run python3 face_recognition_11/seed_test_run.py --help
+usage: seed_test_run.py [-h] model_name test_suite
+
+positional arguments:
+  model_name  Name of the model to test.
+  test_suites  Name of the test suite(s) to run.
```
y27choi, line 53: `test_suites` doesn't match the argument name in the usage on line 44.

examples/face_recognition_11/face_recognition_11/workflow.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+@dataclass(frozen=True)
+class ImageWithMetadata(Image):
+    """
+    An image belonging to an Image Pair containing the locator and metadata
+    for the Face Recognition 1:1 workflow.
+    """
+
+    metadata: Metadata = dataclasses.field(default_factory=dict)
```
y27choi: I checked the example result you shared and test samples seem to be missing metadata.  Is there any metadata that we can use?

examples/face_recognition_11/face_recognition_11/workflow.py
```diff
@@ -0,0 +1,165 @@
...
+    """The metadata associated with an image in the image pair."""
+
+
+@dataclass(frozen=True)
+class TestSample(ImagePair):
+    """Test sample type for Face Recognition 1:1 workflow."""
+
+    a: ImageWithMetadata
```
y27choi: Should be use better names for these images?  What are the standard names for each image in an image pair for FR?

examples/face_recognition_11/face_recognition_11/workflow.py
```diff
@@ -0,0 +1,165 @@
...
+
+    is_same: bool
+    """Whether to treat this image pair as a a genuine pair (True) or an imposter pair (False)."""
+
+
+@dataclass(frozen=True)
+class Inference(BaseInference):
+    """Inference type for Face Recognition 1:1 workflow."""
+
+    left_bbox: Optional[BoundingBox] = None
```
y27choi: It's not clear that left image is image a and right image is image b. Should use consistent names here.

examples/face_recognition_11/face_recognition_11/workflow.py
```diff
@@ -0,0 +1,165 @@
...
+workflow, TestCase, TestSuite, Model = define_workflow(
+    "Face Recognition 1:1",
+    TestSample,
+    GroundTruth,
+    Inference,
+)
+
+
+@dataclass(frozen=True)
+class TestSampleMetrics(MetricsTestSample):  # TODO: Include failure to enroll?
```
y27choi: remove TODO?

examples/face_recognition_11/face_recognition_11/workflow.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+    Image-pair-level metrics for Face Recognition 1:1 workflow.
+    A test sample is can only be true for one of the following: match, false match (FM), or false non-match (FNM).
+    If all categories are false then the sample is a true non-match.
```
y27choi: why not add `is_true_non_match: bool`?

examples/face_recognition_11/face_recognition_11/workflow.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+
+    is_match: bool
+    """
+    An indication of whether the model correct classified an imposter pair as an imposter pair or
```
y27choi: Is this true? Does that mean a true non-match will have `is_match = True`?

examples/face_recognition_11/face_recognition_11/workflow.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+
+    is_false_match: bool
+    """An indication of whether the model incorrectly classified an imposter pair as a genuine pair."""
+
+    is_false_non_match: bool
+    """An indication of whether the model incorrectly classified an genuine pair as a imposter pair."""
+
+    failure_to_enroll: bool
+    """An indication of whether the model failed to detect a face."""
```
y27choi: It could also be that the quality of the detected face image isn't good enough or that the keypoint detection failed.  Instead we could phrase is as "An indication of whether the model failed to infer."

examples/face_recognition_11/face_recognition_11/workflow.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+@dataclass(frozen=True)
+class TestCaseMetrics(MetricsTestCase):
```
y27choi: In comparison to the legacy FR workflow, it's missing some metrics like FTE count, FTE rate, failure count, failure rate, and delta FNMR.

examples/face_recognition_11/face_recognition_11/workflow.py
```diff
@@ -0,0 +1,165 @@
...
+    The percentage of genuine pairs that are incorrectly classified as imposter pairs
+    (i.e. similarity is below threshold) within this test case.
+    """
+
+
+@dataclass(frozen=True)
+class FMRConfiguration(EvaluatorConfiguration):
+    """
+    False Match Rate (FMR) configuration for Face Recognition 1:1 workflow.
+    Specify a minimum
```
y27choi: Should this comment be combined with line 159?

examples/face_recognition_11/pyproject.toml
```diff
@@ -0,0 +1,28 @@
...
+[tool.poetry.group.dev.dependencies]
+pre-commit = "^2.17"
+pytest = "^7"
+pytest-depends = "^1.0.1"
+
+[build-system]
+requires = ["poetry-core>=1.0.0"]
+build-backend = "poetry.core.masonry.api"
+
+[tool.pytest.ini_options]
```
y27choi: I think we talked about this briefly but why is this needed again?

examples/face_recognition_11/face_recognition_11/seed_test_suite.py
```diff
@@ -0,0 +1,103 @@
...
+
+    complete_test_case = TestCase(
+        f"fr 1:1 complete :: {DATASET}",
+        description=f"All images in {DATASET} dataset",
+        test_samples=test_samples_and_ground_truths,
+        reset=True,
+    )
+
+    section_size = 2500
+    splits = [
```
y27choi: What are these splits for?  Ideally, we would still want to have some meaningful test cases as an example.

examples/face_recognition_11/face_recognition_11/workflow.py
```diff
@@ -0,0 +1,165 @@
...
+
+    false_match_rate: Optional[float] = None
+    """
+    FMR to apply for predictions.
+    """
+
+    def display_name(self) -> str:
+        if self.false_match_rate is not None:
+            return f"False Match Rate: {self.false_match_rate}"
+        return "Max False Match Rate"
```
y27choi: What exactly is "Max False Match Rate" here? I didn't see handling of this case in evaluator.py

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,172 @@
...
+from kolena.workflow import Evaluator
+from kolena.workflow import Plot
+
+
+class FaceRecognition11Evaluator(Evaluator):
+    @staticmethod
+    def compute_threshold(inferences: List[Inference], fmr: float) -> float:
+        """
+        The threhsold is calculated as Threshold = Quantile_v(1 - FMR) where the quantiles are of the observed
+        IMPOSTER scores, v.
```
y27choi: What does `v` here represent?

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,172 @@
...
+                is_false_non_match = True
+
+        return TestSampleMetrics(
+            is_match=is_match,
+            is_false_match=is_false_match,
+            is_false_non_match=is_false_non_match,
+            failure_to_enroll=False,
+        )
+
+    def compute_test_sample_metrics(
```
y27choi: consider using [evaluator function](https://docs.kolena.io/reference/workflow/evaluator/#kolena.workflow.evaluator_function.BasicEvaluatorFunction) instead of evaluator class as it computes the test sample metrics once for each configuration.

y27choi: Another benefit of using evaluator function is that we can be sure that we are computing threshold with the complete set of test samples. 

With the class implementation, `compute_test_sample_metrics` is called for each and every test case.  So, what is happening here is that a threshold is computed for each test case.

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,172 @@
...
+    ) -> TestCaseMetrics:
+        n_genuine_pairs = np.sum([gt.is_same for ts, gt, inf in inferences])
+        n_imposter_pairs = np.sum([not gt.is_same for ts, gt, inf in inferences])
+
+        # do not count FTE as belonging to FM or FNM
+        n_fm = np.sum([metric.is_false_match and not metric.failure_to_enroll for metric in metrics])
+        n_fnm = np.sum([metric.is_false_non_match and not metric.failure_to_enroll for metric in metrics])
+
+        return TestCaseMetrics(
+            n_images=len(metrics) * 2,
```
y27choi: Is # of **unique** images really 2 x the number of pairs?

examples/face_recognition_11/face_recognition_11/evaluator.py
ghost: Yeah, I tried with the evaluator function...but the problem is that I need to compute a new threshold across every new test sample since I'm using the similarity scores that come only from a test sample.

examples/face_recognition_11/face_recognition_11/workflow.py
ghost: Okay, I'll stick with A and B for everything then. It's just because the ImagePair class had A and B that I thought it would be good to be consistent with that.

examples/face_recognition_11/README.md
ghost: Oh my bad. Had the commit just forgot to push it.

examples/face_recognition_11/face_recognition_11/evaluator.py
y27choi: > I need to compute a new threshold across every new test sample

Isn't threshold computed against the complete test case? Shouldn't there be only one threshold per configuration for a test suite?

examples/face_recognition_11/face_recognition_11/workflow.py
ghost: The metadata is within the ImageWithMetadata object and so it doesn't seem like the studio can read from the `ImagePair.a.metdata`

examples/face_recognition_11/face_recognition_11/evaluator.py
ghost: Sorry. I meant that it's computed across each test case not test sample since I'm thinking the complete test case and every other test case would have their own threshold.

examples/face_recognition_11/face_recognition_11/workflow.py
ghost: I was following the previous example in confidence thresholds. But yeah that wouldn't really make sense for FMR

examples/face_recognition_11/face_recognition_11/evaluator.py
y27choi: Shouldn't the threshold be computed with the complete test case only and the same threshold should be used on evaluating all test cases?

ghost:
Results page: https://trunk.kolena.io/andy/results?modelIds=2&testSuiteId=20&workflow=Face+Recognition+1%3A1&evaluators=N4IgTADCBcpjBGALAXxUA

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,182 @@
...
+    test_sample_metrics = [compute_per_sample(gt, inf, threshold) for gt, inf in zip(ground_truths, inferences)]
+
+    # compute aggregate metrics across all test cases using `test_cases.iter(...)`
+    all_test_case_metrics: List[Tuple[TestCase, TestCaseMetrics]] = []
+    all_test_case_plots: List[Tuple[TestCase, List[Plot]]] = []
+    for test_case, ts, gt, inf, tsm in test_cases.iter(test_samples, ground_truths, inferences, test_sample_metrics):
+        all_test_case_metrics.append((test_case, compute_test_case_metrics(ts, gt, tsm)))
+        all_test_case_plots.append((test_case, compute_test_case_plots(gt, inf)))
+
+    # TODO: ADD test suite metrics
```
marko-polo-cheno: I'd delete all unused code

examples/face_recognition_11/face_recognition_11/seed_test_run.py
```diff
@@ -0,0 +1,118 @@
...
+            similarity=sample_results["similarity"],
+        )
+
+    model = Model(f"{model_name} [{DATASET}]", infer=infer)
+    print(f"Model: {model}")
+
+    configurations = [
+        FMRConfiguration(false_match_rate=1e-1),
+        FMRConfiguration(false_match_rate=1e-2),
+        FMRConfiguration(false_match_rate=1e-3),
```
marko-polo-cheno: I think 2 configs is sufficient for an example

examples/face_recognition_11/face_recognition_11/seed_test_suite.py
```diff
@@ -0,0 +1,125 @@
...
+    # Metadata Test Cases
+    demographic_subsets = dict(
+        race=["asian", "black", "indian", "middle eastern", "latino hispanic", "white"],  # ignore "unknown"
+        gender=["man", "woman"],  # ignore "unknown"
+    )
+
+    test_cases: List[TestCase] = []
+    for category, tags in demographic_subsets.items():
+        for tag in tags:
+            test_case = create_test_case_for_tag(test_samples_and_ground_truths, category, tag)
```
marko-polo-cheno, line 134: If this is slow, you could use [init_many](https://docs.kolena.io/reference/workflow/test-case/#kolena.workflow.test_case.TestCase.init_many)

examples/face_recognition_11/face_recognition_11/evaluator.py
ghost: This is for the next PR within enhancements

examples/face_recognition_11/face_recognition_11/seed_test_suite.py
ghost: Doesn't seem slow atm

examples/face_recognition_11/face_recognition_11/seed_test_run.py
ghost: The legacy fr has these 6 different configs so I'm trying to follow that

examples/face_recognition_11/README.md
y27choi: bump

examples/face_recognition_11/README.md
y27choi: bump

examples/face_recognition_11/face_recognition_11/workflow.py
y27choi: Are we still keeping `ImageWithMetadata`?  Should we just add metadata to the `TestSample` instead?

examples/face_recognition_11/pyproject.toml
ghost: For the pytests since poetry couldn't seem to find them. I can remove this though since I don't think it'll be required for our users.

y27choi:
> Results page: https://trunk.kolena.io/andy/results?modelIds=2&testSuiteId=20&workflow=Face+Recognition+1%3A1&evaluators=N4IgTADCBcpjBGALAXxUA

Is this the latest result?

ghost:
Results: https://trunk.kolena.io/andy/results?modelIds=20&workflow=Face+Recognition+1%3A1&testSuiteId=21&evaluators=N4IgTAjCBcpgDDMAOArAX3UA

Legacy: https://trunk.kolena.io/andy/results/fr?modelIds=19&testSuiteId=18&fmr=0.01

@y27choi

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,234 @@
...
+    n_fm = np.sum([metric.is_false_match for metric in metrics])
+    n_fnm = np.sum([metric.is_false_non_match for metric in metrics])
+    n_fte = np.sum([metric.failure_to_enroll for metric in metrics])
+
+    unique_images = set()
+    for ts in test_samples:
+        unique_images.add(ts.a.locator)
+        unique_images.add(ts.b.locator)
+
+    Δ_fnmr = 0.0
```
y27choi: Ha I'm surprised your lint checker let you do this. This is cool.

y27choi: What do you think of capitalizing the metric names: `nImages`, `nGenuinePairs`, `FM`, `FNM`, `FMR`, `FNMR`, `FTE`. I think getting rid of `_` can reduce their lengths so it's more presentable on the results table.

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,234 @@
...
+        n_images=len(unique_images),
+        n_genuine_pairs=n_genuine_pairs,
+        n_imposter_pairs=n_imposter_pairs,
+        n_fm=n_fm,
+        fmr=(n_fm / n_imposter_pairs) * 100,
+        n_fnm=n_fnm,
+        fnmr=(n_fnm / n_genuine_pairs) * 100,
+        Δ_fnmr=Δ_fnmr,
+        n_fte=n_fte,
+        fter=(n_fte / (n_genuine_pairs + n_imposter_pairs)) * 100,
```
y27choi: Shouldn't `FTER` be `n_fte / n_images` since FTE is at the image level not the pair level?

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,234 @@
...
+    ]
+
+    min_data = min(min(genuine_values), min(imposter_values))
+    max_data = max(max(genuine_values), max(imposter_values))
+
+    number_of_bins = 50
+    bin_size = (max_data - min_data) / number_of_bins
+    bin_edges = [min_data + i * bin_size for i in range(number_of_bins + 1)]
+
+    hist1_adjusted = list(np.histogram(genuine_values, bins=bin_edges, density=True)[0])
```
y27choi: re. `hist1` and `hist2`, we should be more descriptive on the variable names.

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,234 @@
...
+        fmr=(n_fm / n_imposter_pairs) * 100,
+        n_fnm=n_fnm,
+        fnmr=(n_fnm / n_genuine_pairs) * 100,
+        Δ_fnmr=Δ_fnmr,
+        n_fte=n_fte,
+        fter=(n_fte / (n_genuine_pairs + n_imposter_pairs)) * 100,
+    )
+
+
+def compute_test_case_plots(ground_truths: List[GroundTruth], inferences: List[Inference]) -> Optional[List[Plot]]:
```
y27choi: This function is getting long.  I think we can break this down to multiple functions.

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,234 @@
...
+    )
+
+    return [similarity_dist, curve_test_case_fnmr, curve_test_case_fmr]
+
+
+def compute_test_suite_metrics(
+    test_case_metrics: List[Tuple[TestCase, TestCaseMetrics]],
+    threshold: float,
+) -> TestSuiteMetrics:
+    baseline: TestCaseMetrics = test_case_metrics[0][1]
```
y27choi: How do we know the first test case is the baseline? I would rather not make this assumption

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,234 @@
...
+
+    # compute aggregate metrics across all test cases using `test_cases.iter(...)`
+    all_test_case_metrics: List[Tuple[TestCase, TestCaseMetrics]] = []
+    all_test_case_plots: List[Tuple[TestCase, List[Plot]]] = []
+    baseline_fnmr = 0.0
+    for test_case, ts, gt, inf, tsm in test_cases.iter(test_samples, ground_truths, inferences, test_sample_metrics):
+        all_test_case_metrics.append((test_case, compute_test_case_metrics(ts, gt, tsm, baseline_fnmr)))
+        all_test_case_plots.append((test_case, compute_test_case_plots(gt, inf)))
+
+        # first processed test case is baseline
```
y27choi: Is this assumption necessary?

examples/face_recognition_11/face_recognition_11/seed_test_run.py
y27choi: Given the size of the dataset, does it make sense to evaluate it at 1e-4, 1e-5, and 1e-6 levels?

examples/face_recognition_11/face_recognition_11/workflow.py
```diff
@@ -0,0 +1,190 @@
...
+from kolena.workflow import Metadata
+from kolena.workflow import MetricsTestCase
+from kolena.workflow import MetricsTestSample
+from kolena.workflow import MetricsTestSuite
+from kolena.workflow.annotation import BoundingBox
+from kolena.workflow.annotation import Keypoints
+
+
+@dataclass(frozen=True)
+class ImageWithMetadata(Image):
```
y27choi: Are we still going with `ImageWithMetadata`?  I thought it wasn't showing the seeded metadata with this design.

examples/face_recognition_11/pyproject.toml
```diff
@@ -0,0 +1,22 @@
...
+description = "Example Kolena integration for face recognition 1:1"
+authors = ["Kolena Engineering <eng@kolena.io>"]
+license = "Apache-2.0"
+
+[tool.poetry.dependencies]
+python = ">=3.8,<3.12"
+kolena = ">=0.81.0,<1"
+s3fs = "^2023.5.0"
+pandera = ">=0.9.0,<0.16"
+jupyter = "^1.0.0"
```
y27choi: Do we need this?

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,317 @@
...
+    metrics: List[TestSampleMetrics],
+    baseline_fnmr: float,
+    configuration: ThresholdConfiguration,
+) -> TestCaseMetrics:
+    n_genuine_pairs = np.sum([np.sum(gt.matches) for gt in ground_truths]) / 2
+    n_imposter_pairs = np.sum([np.sum(np.invert(gt.matches)) for gt in ground_truths]) / 2
+
+    n_fm, n_fnm, n_pair_failures = 0, 0, 0
+    for tsm in metrics:
+        n_fm += np.sum([ps.is_false_match for ps in tsm.pair_samples])
```
y27choi: Should we divide all these metrics by 2 since there are duplicate pairs?

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,317 @@
...
+    fmr_y = list()
+
+    for threshold in thresholds:
+        tsm_for_one_threshold = [
+            compute_per_sample(gt, inf, threshold, configuration) for gt, inf in zip(ground_truths, inferences)
+        ]
+
+        n_fm, n_fnm = 0, 0
+        for metric in tsm_for_one_threshold:
+            n_fm += np.sum([ps.is_false_match for ps in metric.pair_samples])
```
y27choi: and same here: divide by 2

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,317 @@
...
+from kolena.workflow import Curve
+from kolena.workflow import CurvePlot
+from kolena.workflow import EvaluationResults
+from kolena.workflow import Plot
+from kolena.workflow import TestCases
+from kolena.workflow.annotation import ScoredClassificationLabel
+from kolena.workflow.metrics import precision, recall, f1_score, iou
+
+
+def compute_threshold(ground_truths, inferences, fmr: float, eps: float = 1e-9) -> float:
```
y27choi: There will be duplicate pairs here as well.

examples/face_recognition_11/face_recognition_11/seed_test_run.py
```diff
@@ -0,0 +1,119 @@
...
+        return Inference(similarities=similarities, bbox=bbox, keypoints=keypoints)
+
+    model = Model(f"{model_name} [FR]", infer=infer)
+    print(f"Model: {model}")
+
+    configurations = [
+        ThresholdConfiguration(false_match_rate=1e-1, iou_threshold=0.5, nmse_threshold=0.5),
+        # ThresholdConfiguration(false_match_rate=1e-2),
+        # ThresholdConfiguration(false_match_rate=1e-3),
+        # ThresholdConfiguration(false_match_rate=1e-4),
```
gordonhart: Remove commented code

examples/face_recognition_11/pyproject.toml
```diff
@@ -0,0 +1,21 @@
...
+version = "0.1.0"
+description = "Example Kolena integration for face recognition 1:1"
+authors = ["Kolena Engineering <eng@kolena.io>"]
+license = "Apache-2.0"
+
+[tool.poetry.dependencies]
+python = ">=3.8,<3.12"
+kolena = ">=0.81.0,<1"
+s3fs = "^2023.5.0"
+pandera = ">=0.9.0,<0.16"
```
gordonhart, line 11: Shouldn't need to pin this transient dependency of `kolena`.

examples/face_recognition_11/tests/test_face_recognition_11.py
gordonhart, line 1: Thanks for adding these! Will need to update the CircleCI config to get these running in CI. See [diff in a previous PR](https://github.com/kolenaIO/kolena/pull/220/files#diff-71e295330550f17d011d6354c7e2c231e9f4a43cdbf92e511ffa0f15cea7ae60) for guidance.

examples/face_recognition_11/tests/test_face_recognition_11.py
```diff
@@ -0,0 +1,32 @@
...
+    args = Namespace(
+        dataset_csv="3://kolena-public-datasets/labeled-faces-in-the-wild/meta/pairs.sample.csv",
+        metadata_csv="s3://kolena-public-datasets/labeled-faces-in-the-wild/meta/metadata.csv",
+    )
+    seed_test_suite_main(args)
+
+
+@pytest.mark.depends(on=["test__seed_test_suite__smoke"])
+def test__seed_test_run__smoke() -> None:
+    args = Namespace(models=["deepface"], test_suites=["fr 1:1 :: labeled-faces-in-the-wild"])
```
y27choi: This might need to be updated

examples/face_recognition_11/README.md
```diff
@@ -0,0 +1,52 @@
+# Example Integration: Face Recognition (1:1)
+
+This example integration uses the [Labeled Faces in the Wild (LFW)](http://vis-www.cs.umass.edu/lfw/) dataset and Face Recognition (FR) workflow to demonstrate how to test and evaluate FR (1:1) systems on Kolena.
```
y27choi: Should we mention how this includes the end-to-end FR model pipeline here?

examples/face_recognition_11/README.md
```diff
@@ -0,0 +1,52 @@
...
+## Usage
+
+The data for this example integration lives in the publicly accessible S3 bucket `s3://kolena-public-datasets`.
+
+First, ensure that the `KOLENA_TOKEN` environment variable is populated in your environment. See our
+[initialization documentation](https://docs.kolena.io/installing-kolena/#initialization) for details.
+
+This project defines two scripts that perform the following operations:
+
+1. [`seed_test_suite.py`](face_recognition_11/seed_test_suite.py) creates a test case with all test samples in a test suite
```
y27choi: Let's include the types of stratification that we use in `seed_test_suite.py` (see other example as a reference)

examples/face_recognition_11/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+2. [`seed_test_run.py`](face_recognition_11/seed_test_run.py) tests a face recognition model against the test suite above
```
y27choi: The evaluation includes multiple models in the pipeline.

examples/face_recognition_11/README.md
```diff
@@ -0,0 +1,52 @@
...
+positional arguments:
+  dataset_csv  CSV file containing image pairs to be tested.
+  metadata_csv CSV file containing the metadata of each image.
+
+optional arguments:
+  -h, --help  show this help message and exit
+```
+
+```shell
+$ poetry run python3 face_recognition_11/seed_test_run.py --help
```
y27choi, line 46: Needs an update

examples/face_recognition_11/face_recognition_11/seed_test_suite.py
```diff
@@ -0,0 +1,173 @@
...
+    locator_normalization_factor = {}
+    non_metadata_fields = {"locator", "normalization_factor"}
+    for record in df_metadata.itertuples(index=False):
+        fields = set(record._fields)
+        metadata_by_locator[record.locator] = {f: getattr(record, f) for f in fields - non_metadata_fields}
+        locator_normalization_factor[record.locator] = record.normalization_factor
+
+    images = defaultdict(list)
+    bbox_keypoints = {}
+    for record in df.itertuples(index=False):
```
y27choi, line 92: For a large scale pair dataset, we should consider separating pair locators and detection GT into two CSV so we don't include duplicate detection for images that are included in multiple pairs.

examples/face_recognition_11/README.md
y27choi: Can we also include how we define the face bbox / keypoint gt as they are not originally from LFW

examples/face_recognition_11/face_recognition_11/seed_test_suite.py
```diff
@@ -0,0 +1,173 @@
...
+            bbox=bbox,
+            keypoints=keypoints,
+            normalization_factor=locator_normalization_factor[locator],
+            count_genuine_pair=np.sum(matches),
+            count_imposter_pair=np.sum(np.invert(matches)),
+        )
+        test_samples_and_ground_truths.append((ts, gt))
+
+    complete_test_case = TestCase(
+        name=f"fr 1:1 holistic complete :: {DATASET}",
```
y27choi: `fr` -> `FR`

examples/face_recognition_11/README.md
y27choi: And how we pared the pairs down to 10k.

examples/face_recognition_11/face_recognition_11/seed_test_run.py
```diff
@@ -0,0 +1,123 @@
...
+            )
+            if right_eye_x
+            else None
+        )
+
+        return Inference(similarities=similarities, bbox=bbox, keypoints=keypoints)
+
+    metadata = dict(detector=detector)
+    model = Model(f"{model_name} [FR]", infer=infer, metadata=metadata)
+    print(f"Model: {model}")
```
y27choi: Are these prints necessary?  When you run with `verbose=True` it should print out the details about `Model` and `TestSuite` when you create / load them.

examples/face_recognition_11/face_recognition_11/seed_test_run.py
```diff
@@ -0,0 +1,123 @@
...
+        for img in test_sample.pairs:
+            # assume no image pair with itself
+            query = df["locator_a"].isin([test_sample.locator, img.locator]) & df["locator_b"].isin(
+                [test_sample.locator, img.locator]
+            )
+            sample_results = df[query]
+            similarity = sample_results["similarity"].values[0] if not sample_results["failure"].values[0] else None
+            similarities.append(similarity)
+
+        if not df[df["locator_a"] == test_sample.locator].empty:
```
y27choi, line 51: I think we can clean this up by separating the detection data to a separate csv file where you can look it up simply by an image locator.

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,405 @@
...
+from kolena.workflow import CurvePlot
+from kolena.workflow import EvaluationResults
+from kolena.workflow import Plot
+from kolena.workflow import TestCases
+from kolena.workflow.annotation import ScoredClassificationLabel, ClassificationLabel
+from kolena.workflow.metrics import precision, recall, f1_score, iou
+
+
+def compute_threshold(ground_truths, inferences, fmr: float, eps: float = 1e-9) -> float:
+    total_matches = []
```
y27choi: `total_matches = [match for match in gt.matches for gt in ground_truths]`

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,405 @@
...
+        )
+        pair_samples.append(pair_sample)
+        count_fnm += is_false_non_match
+        count_fm += is_false_match
+        count_tm += is_match
+        count_tnm += not is_false_non_match and not is_false_match and not is_match
+
+    return TestSampleMetrics(
+        pair_samples=pair_samples,
+        count_FNM=count_fnm,
```
y27choi: I think instead of incrementing in the loop if we can just compute them from there, it will be cleaner. `count_FNM = np.sum([pair.is_false_non_match for pair in pair_samples])`

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,405 @@
...
+
+
+def compute_per_keypoint_case(ground_truths: List[GroundTruth], metrics: List[TestSampleMetrics]) -> PerKeypointMetrics:
+    def process_metric(metric):
+        return metric if metric is not None else 0.0
+
+    return PerKeypointMetrics(
+        Label="Keypoints Detection",
+        Total=np.sum([gt.keypoints is not None for gt in ground_truths]),
+        FTE=np.sum([tsm.keypoint_failure_to_align for tsm in metrics]),
```
y27choi: We should be consistent with the acronym here. Are we going to use FTA or FTE?

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,405 @@
...
+    )
+
+
+def compute_test_case_metrics(
+    test_samples: List[TestSample],
+    ground_truths: List[GroundTruth],
+    metrics: List[TestSampleMetrics],
+    baseline_fnmr: float,
+) -> TestCaseMetrics:
+    n_genuine_pairs = np.sum([np.sum(gt.matches) for gt in ground_truths]) / 2
```
y27choi: Should we use `len(...)` instead of `np.sum(...)` here?

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,405 @@
...
+        return total / len(nmses) if len(nmses) > 0 else 0
+
+    nmses = [mts.keypoint_NMSE for mts in metrics]
+    x = np.linspace(0, 0.5, 251).tolist()
+    y = [compute_failure_rate(x_value, nmses) for x_value in x]
+    plots.append(
+        CurvePlot(
+            title="Keypoint Detection: Alignment Failure Rate vs. NMSE Threshold",
+            x_label="NMSE Threshold",
+            # x_config=AxisConfig(type="log"),
```
y27choi: remove

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,400 @@
...
+
+
+def evaluate_face_recognition_11(
+    test_samples: List[TestSample],
+    ground_truths: List[GroundTruth],
+    inferences: List[Inference],
+    test_cases: TestCases,
+    configuration: ThresholdConfiguration,
+) -> EvaluationResults:
+    pairs = set([(ts.locator, pair.locator) for ts in test_samples for pair in ts.pairs])
```
y27choi: create a util function for this

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,400 @@
...
+    # compute per-sample metrics for each test sample
+    test_sample_metrics = [
+        compute_per_sample(gt, inf, threshold, configuration, test_samples[i])
+        for i, (gt, inf) in enumerate(zip(ground_truths, inferences))
+    ]
+
+    # compute aggregate metrics across all test cases using `test_cases.iter(...)`
+    all_test_case_metrics: List[Tuple[TestCase, TestCaseMetrics]] = []
+    all_test_case_plots: List[Tuple[TestCase, List[Plot]]] = []
+    baseline = compute_test_case_metrics(test_samples, ground_truths, test_sample_metrics, 0, unique_pairs)
```
y27choi: It looks like there are many places where `compute_test_case_metrics` is called just to calculate FMR and FNMR.  This will results in multiple computation of not only the FR metrics but also the detection metrics.  You can make the FR metrics computation a separate function and it can be called within `compute_test_case_metrics` and in this case, you don't perform unnecessary computation.

y27choi: We can also move `unique_pairs` computation inside this new `compute_pair_metrics` function.

examples/face_recognition_11/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+    - `labeled-faces-in-the-wild :: gender [FR]`, stratified by `gender` using image metadata
+    - `labeled-faces-in-the-wild :: race [FR]`, stratified by `race` using image metadata
+
+    Run this command to seed the default test suite:
+    ```shell
+    poetry run python3 face_recognition_11/seed_test_suite.py
+    ```
+
+    > **_NOTE:_**  Face bounding box and keypoint ground truths are not originally from LFW. They are collected from RetinaFace. Also, a sample size of 10K from LFW is used as the dataset instead.
```
y27choi:
```suggestion
    > **NOTE:**  Face bounding box and keypoint ground truths are inferred from RetinaFace as they are not provided in LFW dataset. Also, for the demo purpose, we have subsampled 10,000 pairs, made up of 9,983 images, from LFW.
```

It would also be nice to add a link to RetinaFace model

examples/face_recognition_11/README.md
```diff
@@ -0,0 +1,59 @@
...
+    ```shell
+    poetry run python3 face_recognition_11/seed_test_run.py
+    ```
+
+Command line arguments are defined within each script to specify what model to use and what test suite to seed/evaluate.
+Run a script using the `--help` flag for more information:
+
+```shell
+$ poetry run python3 face_recognition_11/seed_test_run.py --help
+usage: seed_test_run.py [-h] [--models MODELS [MODELS ...]] [--test_suites TEST_SUITES [TEST_SUITES ...]]
```
y27choi: I believe there is another argument to this script. Please check both scripts and copy the exact results from the `--help` output.

y27choi: I don't see any changes made here.

examples/face_recognition_11/face_recognition_11/seed_test_run.py
```diff
@@ -0,0 +1,123 @@
...
+            query = df["locator_a"].isin([test_sample.locator, img.locator]) & df["locator_b"].isin(
+                [test_sample.locator, img.locator],
+            )
+            sample_results = df[query]
+            similarity = sample_results["similarity"].values[0] if not sample_results["failure"].values[0] else None
+            similarities.append(similarity)
+
+        if not df[df["locator_a"] == test_sample.locator].empty:
+            r = next(df[df["locator_a"] == test_sample.locator].itertuples(index=False))
+            min_x, min_y, max_x, max_y = r.a_min_x, r.a_min_y, r.a_max_x, r.a_max_y
```
y27choi: Can we clean this up?  It's essentially the same operation for `a` and `b`.  If we use `iterrows()` then we can do something like
```
bbox = BoundingBox((r[f"{side}_min_x"], r[f"{side}_min_y"]), (r[f"{side}_min_x"], r[f"{side}_max_y"])) if r[f"{side}_min_x"] else None
```

examples/face_recognition_11/face_recognition_11/utils.py
```diff
@@ -0,0 +1,188 @@
...
+    pairs = {(ts.locator, pair.locator) for ts in test_samples for pair in ts.pairs}
+    unique_pairs = {(a, b) if a <= b else (b, a) for a, b in pairs}
+    return unique_pairs
+
+
+def compute_pair_metrics(
+    test_samples: List[TestSample],
+    ground_truths: List[GroundTruth],
+    metrics: List[TestSampleMetrics],
+) -> Tuple[list, list, list, list, list, list]:
```
y27choi: Is it returning lists?

examples/face_recognition_11/face_recognition_11/seed_test_run.py
```diff
@@ -0,0 +1,123 @@
...
+        )
+
+        return Inference(similarities=similarities, bbox=bbox, keypoints=keypoints)
+
+    metadata = dict(detector=detector)
+    model = Model(f"{model_name} [FR]", infer=infer, metadata=metadata)
+
+    configurations = [
+        ThresholdConfiguration(false_match_rate=1e-1, iou_threshold=0.5, nmse_threshold=0.5),
+        ThresholdConfiguration(false_match_rate=1e-4, iou_threshold=0.5, nmse_threshold=0.5),
```
y27choi: Let's remove this threshold configuration

examples/face_recognition_11/face_recognition_11/utils.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+def compute_threshold(
+    test_samples: List[TestSample],
+    ground_truths: List[GroundTruth],
+    inferences: List[Inference],
+    fmr: float,
+    eps: float = 1e-9,
+) -> float:
+    # address duplicates
```
y27choi, line 46: Isn't threshold computed only for the complete test case? If so, can't we just use the ` / 2` trick here?

y27choi: If we want to use this logic, `seen` should be a dict not a list.  Checking through a list should be O(n) and it becomes too expensive.

examples/face_recognition_11/face_recognition_11/utils.py
```diff
@@ -0,0 +1,188 @@
...
+            )
+
+    n_genuine_pairs = np.sum([genuine_pairs[pair] for pair in unique_pairs])
+    n_imposter_pairs = np.sum([imposter_pairs[pair] for pair in unique_pairs])
+    n_fm = np.sum([fm[pair] for pair in unique_pairs])
+    n_fnm = np.sum([fnm[pair] for pair in unique_pairs])
+    n_pair_failures = np.sum([pair_failures[pair] for pair in unique_pairs])
+    n_fte = np.sum([fte[pair] for pair in unique_pairs])
+
+    return (n_genuine_pairs, n_imposter_pairs, n_fm, n_fnm, n_pair_failures, n_fte)
```
y27choi: FYI when there are many returned values like here it's better to create an object.

y27choi: And then you can include things like FNMR, FMR and FTER as well.

examples/face_recognition_11/face_recognition_11/seed_test_run.py
```diff
@@ -0,0 +1,118 @@
...
+            Keypoints(
+                [
+                    (r[f"{pair}_right_eye_x"], r[f"{pair}_right_eye_y"]),
+                    (r[f"{pair}_left_eye_x"], r[f"{pair}_left_eye_y"]),
+                    (r[f"{pair}_nose_x"], r[f"{pair}_nose_y"]),
+                    (r[f"{pair}_mouth_right_x"], r[f"{pair}_mouth_right_y"]),
+                    (r[f"{pair}_mouth_left_x"], r[f"{pair}_mouth_left_y"]),
+                ]
+            )
+            if r[f"{pair}right_eye_x"]
```
y27choi: r[f"{pair}_right_eye_x"]

examples/face_recognition_11/tests/test_face_recognition_11.py
y27choi: bump. 
Also, we need to update the [quickstart.md](https://github.com/kolenaIO/kolena/blob/trunk/docs/quickstart.md) and [README.md](https://github.com/kolenaIO/kolena/blob/trunk/examples/README.md).

examples/face_recognition_11/tests/test_face_recognition_11.py
```diff
@@ -0,0 +1,39 @@
...
+    )
+    seed_test_suite_main(args)
+
+
+@pytest.mark.depends(on=["test__seed_test_suite__smoke"])
+def test__seed_test_run__smoke() -> None:
+    args = Namespace(
+        models=["VGG-Face", "Facenet512"],
+        test_suites=[f"{DATASET} :: gender [FR]", f"{DATASET} :: race [FR]"],
+    )
```
y27choi, line 40: missing `detectors`

examples/face_recognition_11/face_recognition_11/evaluator.py
y27choi: bump

examples/face_recognition_11/face_recognition_11/workflow.py
```diff
@@ -0,0 +1,201 @@
...
+class TestSuiteMetrics(MetricsTestSuite):
+    """Test-suite-level metrics for Face Recognition workflow."""
+
+    Threshold: float
+    FM: int
+    FNM: int
+    FNMR: float
+    TotalFTE: int
+    TotalBBoxFTE: int
+    TotalKeypointFTE: int
```
y27choi: Should this be FTA as well?

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,402 @@
...
+
+        n_genuine, n_imposter, n_fm, n_fnm, _, _ = compute_pair_metrics(
+            test_samples,
+            ground_truths,
+            tsm_for_one_threshold,
+        )
+        fnmr_y.append(n_fnm / n_genuine)
+        fmr_y.append(n_fm / n_imposter)
+
+    print(fnmr_y)
```
y27choi: print

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,402 @@
...
+
+    plots.append(
+        CurvePlot(
+            title="Recognition: Test Case FNMR vs. Baseline FMR",
+            x_label="Baseline False Match Rate",
+            y_label="Test Case False Non-Match Rate (%)",
```
y27choi: Should we drop (%) now that it's purely a rate?

y27choi (commented):
Finished the review.  It's a lot of metrics. Thanks for putting this together.  It's a lot of hard work trying to get all the numbers right.  Once all the comments are addressed, I will do a final review and should be good to merge.  I will also run a smoke test.

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,402 @@
...
+        pair_samples=pair_samples,
+        count_FNM=count_fnm,
+        count_FM=count_fm,
+        count_TM=count_tm,
+        count_TNM=count_tnm,
+        similarity_threshold=threshold,
+        bbox_IoU=iou_value if iou_value is not None else 0.0,
+        bbox_TP=[inference.bbox] if tp and inference.bbox is not None else [],
+        bbox_FP=[inference.bbox] if fp and inference.bbox is not None else [],
+        bbox_FN=[inference.bbox] if (not tp and not fp) and inference.bbox is not None else [],
```
gordonhart: How can an inference be a false negative? Usually, in object detection false negatives are ground truth bounding boxes that are not matched with a corresponding prediction.

gordonhart: It almost feels like we want `bbox_success` and `bbox_failure` here instead of TP/FP/FN. I'll defer to @y27choi here though

y27choi: Thanks for flagging.  I think FN should be GT bbox that are not matched, but `bbox_success` and `bbox_failure` make sense here. Will these names work with the default coloring logic?

gordonhart: Yep, default coloration logic will work as expected using those names.

I've been on the fence about making this suggestion because TP/FP/FN are so standard for OD, but it really does feel like success/failure is more relevant here when there's always a single box to predict in the image. If we see this integration extending to use cases with 0-n ground truth bboxes or 0-n prediction bboxes, then let's keep TP/FP/FN, and if so let's update the FN logic to be unmatched GTs.

y27choi: > let's keep TP/FP/FN, and if so let's update the FN logic to be unmatched GTs

I'm leaning towards just updating FN logic here. @a75w-kolena  and I talked about potentially adding multiple faces inference in the future and keeping it as TP/FP/FN makes sense in such case.

docs/assets/images/fr.png
y27choi, line 1: I would love to use an image from our website but I would prefer to see a pair image view for FR.  What do you think of adding an image like this?:
image

examples/face_recognition_11/face_recognition_11/seed_test_run.py
```diff
@@ -0,0 +1,117 @@
...
+    ]
+
+    for test_suite_name in test_suite_names:
+        test_suite = TestSuite.load(test_suite_name)
+        print(f"Test Suite: {test_suite}")
+        test(model, test_suite, evaluate_face_recognition_11, configurations)
+
+
+def main(args: Namespace) -> int:
+    kolena.initialize(os.environ["KOLENA_TOKEN"], verbose=True)
```
y27choi: `os.environ["KOLENA_TOKEN"]` doesn't need to be included anymore.

examples/face_recognition_11/tests/test_face_recognition_11.py
```diff
@@ -0,0 +1,40 @@
...
+        metadata_csv=f"s3://{BUCKET}/{DATASET}/meta/metadata.csv",
+    )
+    seed_test_suite_main(args)
+
+
+@pytest.mark.depends(on=["test__seed_test_suite__smoke"])
+def test__seed_test_run__smoke() -> None:
+    args = Namespace(
+        models=["VGG-Face", "Facenet512"],
+        default=["MTCNN", "yolov8n-face"],
```
y27choi: Is this the right name for `detectors` argument?

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,392 @@
...
+        pair_samples=pair_samples,
+        count_FNM=np.sum([pair.is_false_non_match for pair in pair_samples]),
+        count_FM=np.sum([pair.is_false_match for pair in pair_samples]),
+        count_TM=np.sum([pair.is_match for pair in pair_samples]),
+        count_TNM=np.sum([not is_false_non_match and not is_false_match and not is_match for pair in pair_samples]),
+        similarity_threshold=threshold,
+        bbox_IoU=iou_value if iou_value is not None else 0.0,
+        bbox_TP=[inference.bbox] if tp and not bbox_fte else [],
+        bbox_FP=[inference.bbox] if fp and not bbox_fte else [],
+        bbox_FN=[inference.bbox] if fn and not bbox_fte else [],
```
y27choi: should be `ground_truth.bbox`

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,392 @@
...
+    # Stage 1: Detection
+    if inference.bbox is None:
+        bbox_fte = True
+
+    iou_value = (
+        iou(ground_truth.bbox, inference.bbox)
+        if (ground_truth.bbox is not None and inference.bbox is not None)
+        else None
+    )
+    bbox_fte, tp, fp, fn = False, False, False, False
```
y27choi: Why is `bbox_fte` getting overwritten to `False` here?

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,392 @@
...
+
+    if iou_value is not None:
```
y27choi: If `inference.box` is None, `iou_value` becomes None therefore this block of the code doesn't get run. In which case, we are not counting FN properly.

examples/face_recognition_11/face_recognition_11/evaluator.py
```diff
@@ -0,0 +1,392 @@
...
+
+    return TestSampleMetrics(
+        pair_samples=pair_samples,
+        count_FNM=np.sum([pair.is_false_non_match for pair in pair_samples]),
+        count_FM=np.sum([pair.is_false_match for pair in pair_samples]),
+        count_TM=np.sum([pair.is_match for pair in pair_samples]),
+        count_TNM=np.sum([not is_false_non_match and not is_false_match and not is_match for pair in pair_samples]),
+        similarity_threshold=threshold,
+        bbox_IoU=iou_value if iou_value is not None else 0.0,
+        bbox_TP=[ground_truth.bbox] if tp and not bbox_fte else [],
```
y27choi: Why are TP/FP updated to `ground_truth`?

docs/assets/images/fr.png
gordonhart, line 1: Do we use "match pairs" and "mismatch pairs" terminology in this integration? I will provide an alternate image to use.

y27choi (approved):
With this [latest result](https://trunk.kolena.io/andy/results?modelIds=67&workflow=Face+Recognition+1%3A1&testSuiteId=149&evaluators=N4IgjALAnCBcoDYFzABjQGhAg7C9qAvoUA), I'm approving this PR.

y27choi:
@a75w-kolena can we re-seed this workflow on app/demo when you can?

ghost:
[Latest Result](https://trunk.kolena.io/andy/results?modelIds=67&testSuiteId=149&workflow=Face+Recognition+1%3A1)

PR #230, opened 2023-09-06
Generic workflow grabbag datatype with pydantic

ly-kolena:
### Linked issue(s):

Leverage `pydantic` dataclass's [`extra` config](https://docs.pydantic.dev/1.10/usage/dataclasses/#dataclass-config) to implement grabbag-like behavior for `DataObject`.

Limitation:
1. requires users to use pydantic dataclass, not stdlib dataclass, if they extend built-in `DataObject`
2. requires users to use `config=dict(extra="allow")` (or equivalent syntax) if they extend built-in `DataObject`
3. grabbag-ed fields could have mismatched datatypes between serialization and deserialization for non-python-native data objects. For example, kolena `ClassificationLabel` as an extra field, can only be deserialized back to `dict` as there's no type information at deserialization time.
4. other limitation of pydantic's `extra` capability, see docs linked above.
5. raises minimal pydantic version from 1.8 to 1.10

Error and behavior from existing implementation would remain otherwise.

### DIFF
--- a/kolena/workflow/_datatypes.py
+++ b/kolena/workflow/_datatypes.py
@@ -119,7 +119,7 @@ class DataObject(metaclass=ABCMeta):
                 for f in fields + extras
             ],
         )
-        return f"""{self.__class__.__qualname__}({value_str})"""
+        return f"{self.__class__.__qualname__}({value_str})"
 
     def _to_dict(self) -> Dict[str, Any]:
         def serialize_value(value: Any) -> Any:
--- a/kolena/workflow/_datatypes.py
+++ b/kolena/workflow/_datatypes.py
@@ -59,7 +59,7 @@ def _allow_extra(cls: Type[T]) -> bool:
 _DATA_TYPE_MAP = {}
 
 
-def get_full_type(obj: "TypedDataObject") -> str:
+def _get_full_type(obj: "TypedDataObject") -> str:
     data_type = obj._data_type()
     return f"{data_type._data_category()}/{data_type.value}"
 
@@ -70,7 +70,7 @@ def _get_data_type(name: str) -> Optional[Type["TypedDataObject"]]:
 
 # used for TypedBaseDataObject to register themselves to be used in dataclass extra fields deserialization
 def _register_data_type(cls) -> None:
-    full_name = get_full_type(cls)
+    full_name = _get_full_type(cls)
     # leverage class inheritance order, only keep base classes of a datatype
     if full_name not in _DATA_TYPE_MAP:
         _DATA_TYPE_MAP[full_name] = cls
--- a/kolena/workflow/_datatypes.py
+++ b/kolena/workflow/_datatypes.py
@@ -49,10 +49,51 @@ def _double_under(input: str) -> bool:
     return input.startswith("__") and input.endswith("__")
 
 
-def allow_extra(cls: Type[T]) -> bool:
+def _allow_extra(cls: Type[T]) -> bool:
     # `pydantic.dataclasses.is_built_in_dataclass` would have false-positive when a stdlib-dataclass decorated
     # class extends a pydantic dataclass
     return "__pydantic_model__" in vars(cls) and cls.__pydantic_model__.Config.extra == Extra.allow
+
+
+# used to track data_type string -> TypedDataObject
+_DATA_TYPE_MAP = {}
+
+
+def get_full_type(obj: "TypedDataObject") -> str:
+    data_type = obj._data_type()
+    return f"{data_type._data_category()}/{data_type.value}"
+
+
+def _get_data_type(name: str) -> Optional[Type["TypedDataObject"]]:
+    return _DATA_TYPE_MAP.get(name, None)
+
+
+# used for TypedBaseDataObject to register themselves to be used in dataclass extra fields deserialization
+def _register_data_type(cls) -> None:
+    full_name = get_full_type(cls)
+    # leverage class inheritance order, only keep base classes of a datatype
+    if full_name not in _DATA_TYPE_MAP:
+        _DATA_TYPE_MAP[full_name] = cls
+
+
+def _deserialize_typed_dataobject(value: Dict[Any, Any]) -> Any:
+    data_type = _get_data_type(value[DATA_TYPE_FIELD])
+    if data_type is None:
+        return value
+
+    return data_type._from_dict(value)
+
+
+def _try_deserialize_typed_dataobject(value: Any) -> Any:
+    if isinstance(value, list):
+        # only attempt deserialization when it is likely this is a list of typed data objects
+        if value and isinstance(value[0], dict) and DATA_TYPE_FIELD in value[0]:
+            return [_try_deserialize_typed_dataobject(val) for val in value]
+    elif isinstance(value, dict):
+        if DATA_TYPE_FIELD in value:
+            return _deserialize_typed_dataobject(value)
+
+    return value
 
 
 @dataclass(frozen=True, config=ValidatorConfig)
@@ -77,7 +118,7 @@ class DataObject(metaclass=ABCMeta):
 
         items = [(field.name, getattr(self, field.name)) for field in dataclasses.fields(type(self))]
         field_names = {field.name for field in dataclasses.fields(type(self))}
-        if allow_extra(type(self)):
+        if _allow_extra(type(self)):
             for key, val in vars(self).items():
                 if key not in field_names and not _double_under(key):
                     items.append((key, val))
@@ -163,10 +204,10 @@ class DataObject(metaclass=ABCMeta):
 
         items = {f.name: deserialize_field(f, obj_dict.get(f.name, None)) for f in dataclasses.fields(cls) if f.init}
         field_names = {f.name for f in dataclasses.fields(cls)}
-        if allow_extra(cls):
+        if _allow_extra(cls):
             for key, val in obj_dict.items():
                 if key not in field_names:
-                    items[key] = val
+                    items[key] = _try_deserialize_typed_dataobject(val)
         return cls(**items)
 
 
--- a/kolena/workflow/_datatypes.py
+++ b/kolena/workflow/_datatypes.py
@@ -84,6 +84,8 @@ def _deserialize_typed_dataobject(value: Dict[Any, Any]) -> Any:
     return data_type._from_dict(value)
 
 
+# best effort to deserialize typed data objects in dataclass extra fields
+# note: since a "data_type" string could map to multiple classes through inheritance, only base case would be used.
 def _try_deserialize_typed_dataobject(value: Any) -> Any:
     if isinstance(value, list):
         # only attempt deserialization when it is likely this is a list of typed data objects
--- a/kolena/workflow/_datatypes.py
+++ b/kolena/workflow/_datatypes.py
@@ -30,6 +30,7 @@ import numpy as np
 import pandas as pd
 import pandera as pa
 from pandera.typing import Series
+from pydantic import Extra
 from pydantic.dataclasses import dataclass
 
 from kolena._utils.dataframes.validators import validate_df_schema
@@ -42,6 +43,16 @@ from kolena._utils.serde import with_serialized_columns
 from kolena._utils.validators import ValidatorConfig
 
 T = TypeVar("T", bound="DataObject")
+
+
+def _double_under(input: str) -> bool:
+    return input.startswith("__") and input.endswith("__")
+
+
+def allow_extra(cls: Type[T]) -> bool:
+    # `pydantic.dataclasses.is_built_in_dataclass` would have false-positive when a stdlib-dataclass decorated
+    # class extends a pydantic dataclass
+    return "__pydantic_model__" in vars(cls) and cls.__pydantic_model__.Config.extra == Extra.allow
 
 
 @dataclass(frozen=True, config=ValidatorConfig)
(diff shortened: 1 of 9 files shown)


tests/unit/workflow/test_data_object.py
```diff
@@ -61,3 +63,23 @@ class DataclassesTester(DataObject):
     tester = DataclassesTester(z=False, a="foobar", b=0.3)
     serialized = tester._to_dict()
     assert list(serialized.keys()) == ["b", "a", "z"]
+
+
+def test__data_object__extras() -> None:
+    class Config:
+        extra = Extra.allow
+
+    @dataclass(frozen=True, config=Config)
```
ly-kolena: @gordonhart What do you think of using `pydantic.dataclass` with `extra` (depends on 1.10 above) .

gordonhart: This looks like it could be exactly what we want. Are there additional tests we can write to see how well this works for different cases?

kolena/workflow/_datatypes.py
```diff
@@ -44,6 +45,14 @@
 T = TypeVar("T", bound="DataObject")
 
 
+def _double_under(input: str) -> bool:
+    return input.startswith("__") and input.endswith("__")
+
+
+def allow_extra(cls) -> bool:
+    return "__pydantic_model__" in vars(cls) and cls.__pydantic_model__.Config.extra == Extra.allow
```
ly-kolena, line 55: Could use an extra pair of eye to see if there's any issue with this logic.

tests/integration/workflow/dummy.py
```diff
@@ -43,6 +44,26 @@ class DummyInference(Inference):
     score: float
 
 
+class LocalConfig:
+    extra = Extra.allow
```
gordonhart, line 48: Should we add extra config to every existing dataclass (test samples/annotations/assets)?

Edit: extra has been there all along via [ValidatorConfig](https://github.com/kolenaIO/kolena/blob/trunk/kolena/_utils/validators.py#L26), it's just been invisible because the extra fields are not serialized.

tests/unit/workflow/test_data_object.py
gordonhart: Do you know what in particular is not supported about `extra="allow"`? From [Pydantic docs](https://docs.pydantic.dev/latest/usage/dataclasses/#dataclass-config):

Screen Shot 2023-09-06 at 9 34 21 PM

Edit: asked the author on https://github.com/pydantic/pydantic/pull/6470.

tests/unit/workflow/test_data_object.py
```diff
@@ -61,3 +63,85 @@ class DataclassesTester(DataObject):
...
+def test__data_object__extras_ignore() -> None:
+    @dataclass(frozen=True, config={"extra": "ignore"})
+    class IgnoreExtraTester(DataObject):
+        b: float
+        a: str
+        z: bool
+
+    serialized = dict(z=False, a="foobar", b=0.3, y=["hello"], x="world")
+
+    # pydantic dataclass without `extra = allow` should still work
```
yifanwu-kolena: seems like this comment doesn't reflect what we are testing here

tests/unit/workflow/test_data_object.py
```diff
@@ -61,3 +63,85 @@ class DataclassesTester(DataObject):
...
+        b: float
+        a: str
+        z: bool
+
+    @dataclass(frozen=True)
+    class CustomData:
+        foo: str
+
+    # extras should still be checked
+    with pytest.raises(ValueError):
```
yifanwu-kolena, line 141: the surface failure reason is b/c CustomData is not inherited from `DataObject`, maybe adding some comments or making this test func name more explicit for future reading

kolena/workflow/_datatypes.py
```diff
@@ -44,6 +45,16 @@
 T = TypeVar("T", bound="DataObject")
 
 
+def _double_under(input: str) -> bool:
+    return input.startswith("__") and input.endswith("__")
+
+
+def allow_extra(cls) -> bool:
```
yifanwu-kolena: it's interesting that mypy doesn't complain missing type hints at here

tests/unit/workflow/test_data_object.py
ly-kolena: The test name indicates passing invalid object to deserialization, which in my head is clear. Can you suggest a name or as to why it is not clear?

yifanwu-kolena: cool, i think i got it for now

kolena/workflow/_datatypes.py
yifanwu-kolena: just want to double check that if this is expected

```py
def test__data_object__extras_allow() -> None:
    class Config:
        extra = Extra.allow

    @dataclass(frozen=True, config=Config)
    class DataclassesTester(DataObject):
        b: float
        a: str
        z: bool

    @dataclass(frozen=True) 
    class Base(DataObject):
        ...

    @dataclass(frozen=True)
    class CustomDataInternal(Base):
        foo: str

    @dataclass(frozen=True)
    class CustomData(DataObject):
        foo: CustomDataInternal

    # extras should still be checked
    assert DataclassesTester(z=False, a="foobar", b=0.3, y=CustomData(foo=CustomDataInternal(foo="bar")))._to_dict() == {'a': 'foobar', 'b': 0.3, 'y': {'foo': {'foo': 'bar'}}, 'z': False}
``` 

```py
def test__data_object__extras_allow() -> None:
    class Config:
        extra = Extra.allow

    @dataclass(frozen=True, config=Config)
    class DataclassesTester(DataObject):
        b: float
        a: str
        z: bool

    @dataclass(frozen=True, config=Config) 
    class Base(dict):
        ...

    @dataclass(frozen=True)
    class CustomDataInternal(Base):
        foo: str

    @dataclass(frozen=True)
    class CustomData(DataObject):
        foo: CustomDataInternal

    # extras should still be checked
    assert DataclassesTester(z=False, a="foobar", b=0.3, y=CustomData(foo=CustomDataInternal(foo="bar")))._to_dict() == {'a': 'foobar', 'b': 0.3, 'y': {'foo': {}}, 'z': False}
```

gordonhart:
@ly-kolena do you have a strategy in mind for dealing with the absence of extra fields from the `__str__` representation? IMO this hurts usability a lot for users who opt for the unstructured approach. When using e.g. `GroundTruth` as a freeform grab bag, `print(gt)` will always yield `GroundTruth()`.


```
>>> from kolena.workflow import GroundTruth
>>> from kolena.workflow.annotation import BoundingBox
>>> print(GroundTruth(example="string", bboxes=[BoundingBox(top_left=(0, 0), bottom_right=(1, 1))]))
GroundTruth() 
```

kolena/workflow/_datatypes.py
```diff
@@ -146,6 +162,11 @@ def deserialize_field(
             )
 
         items = {f.name: deserialize_field(f, obj_dict.get(f.name, None)) for f in dataclasses.fields(cls) if f.init}
+        field_names = {f.name for f in dataclasses.fields(cls)}
+        if allow_extra(cls):
+            for key, val in obj_dict.items():
+                if key not in field_names:
+                    items[key] = val
```
gordonhart: This may need more attention, as structured values are not correctly deserialized:

```py
from kolena.workflow import GroundTruth
from kolena.workflow.annotation import BoundingBox

gt = GroundTruth(bboxes=[BoundingBox(top_left=(1, 1), bottom_right=(2, 2))])

# succeeds
assert isinstance(gt.bboxes[0], BoundingBox)

# fails, bboxes is still List[Dict[str, ...]] instead of List[BoundingBox]
assert isinstance(GroundTruth._from_dict(gt._to_dict()).bboxes[0], BoundingBox)
```

ly-kolena: At this point, we are trying to instantiate an instance, based on class info, from a dictionary. For extra fields. We don't have type information, i.e. from `dataclass.fields`, as such can only keep what it is from the dict.

Let me know if I missed something.

gordonhart: For typed data objects, we have `data_type`, which should be sufficient for our deserialization needs. Will need some extra handling here but for any non-scalar object we can deserialize its members using the class corresponding to the specified `data_type`.

kolena/workflow/_datatypes.py
ly-kolena: Given existing serialization logic, it is expected. `CustomDataInternal` is being detected as `dict`, which we then use `.items()` to serialize its content. Therefore `foo` is not considered.

It can be discussed of a better way to handle this case, however I'll leave that to the future.

ly-kolena:
> @ly-kolena do you have a strategy in mind for dealing with the absence of extra fields from the `__str__` representation? IMO this hurts usability a lot for users who opt for the unstructured approach. When using e.g. `GroundTruth` as a freeform grab bag, `print(gt)` will always yield `GroundTruth()`.
> 
> ```
> >>> from kolena.workflow import GroundTruth
> >>> from kolena.workflow.annotation import BoundingBox
> >>> print(GroundTruth(example="string", bboxes=[BoundingBox(top_left=(0, 0), bottom_right=(1, 1))]))
> GroundTruth() 
> ```

From my understanding so far, it is a "limitation" of using `dataclass`, as it doesn't really do inheritance, other than finding all fields from its ancestors. Other reserved double_under methods are all generated from this collected list of fields. We can provide our own implementation of these double_under methods for a class, but it won't have effect on dataclasses that inherit from it.

reference [here](https://docs.python.org/3/library/dataclasses.html#inheritance) and [here](https://docs.python.org/3/library/dataclasses.html#post-init-processing)

kolena/workflow/_datatypes.py
ly-kolena: Added a best-effort attempt to deserialize `TypedDataObject`. Limitations is noted in comment.

ly-kolena:
> @ly-kolena do you have a strategy in mind for dealing with the absence of extra fields from the `__str__` representation? IMO this hurts usability a lot for users who opt for the unstructured approach. When using e.g. `GroundTruth` as a freeform grab bag, `print(gt)` will always yield `GroundTruth()`.
> 
> ```
> >>> from kolena.workflow import GroundTruth
> >>> from kolena.workflow.annotation import BoundingBox
> >>> print(GroundTruth(example="string", bboxes=[BoundingBox(top_left=(0, 0), bottom_right=(1, 1))]))
> GroundTruth() 
> ```

[Prototyped](https://github.com/kolenaIO/kolena/pull/233/files) a custom `dataclass` wrapping around pydantic dataclass. Not a fan, but for reference.

kolena/workflow/_datatypes.py
```diff
@@ -44,10 +45,82 @@
...
+            return _deserialize_typed_dataobject(value)
+
+    return value
+
+
 @dataclass(frozen=True, config=ValidatorConfig)
 class DataObject(metaclass=ABCMeta):
     """The base for various objects in `kolena.workflow`."""
 
+    def __str__(self):
```
ly-kolena, line 107: Implemented custom `__str__` to include extra fields. My original oversight was that I focused on `__repr__` instead. `dataclass` doesn't generate `__str__`, only `__repr__`, as such provides us an window for `print` customization.

cc @gordonhart @munkyshi @yifanwu-kolena for another review. See unit tests for details, specifically `test__data_object__extras_allow`

kolena/workflow/_datatypes.py
```diff
@@ -44,10 +45,82 @@
...
+        extras = [f for f in vars(self) if f not in fields and not _double_under(f)]
+
+        # caveat: dataclass generate `__repr__` for decorated classes, as such calls `__str__` for `DataObject` instead.
+        value_str = ", ".join(
+            [
+                f"{f}={getattr(self, f)}" if isinstance(getattr(self, f), DataObject) else f"{f}={getattr(self, f)!r}"
+                for f in fields + extras
+            ],
+        )
+        return f"""{self.__class__.__qualname__}({value_str})"""
```
yifanwu-kolena: nit: might don't need multiple line string `"""` at here

yifanwu-kolena (approved):

munkyshi:
Is there something I have to do to get the `__str__` representation working? This is rebased off of this branch, with `andrew/load-model-metrics` for both `kolena` and `shipyard`.

Screenshot 2023-09-18 at 16 17 48

ly-kolena:
> Is there something I have to do to get the `__str__` representation working? This is rebased off of this branch, with `andrew/load-model-metrics` for both `kolena` and `shipyard`.

Good catch on the finer points in python's use of `repr` and `str`. If you use `print(metrics[0])` , the console should look something like
```
>>> from kolena.workflow import MetricsTestSample
>>> a = dict(is_correct=True, BERT_prec=0.1)
>>> b = MetricsTestSample(**a)
>>> b
MetricsTestSample()
>>> print(b)
MetricsTestSample(is_correct=True, BERT_prec=0.1)
```

munkyshi:
> > Is there something I have to do to get the `__str__` representation working? This is rebased off of this branch, with `andrew/load-model-metrics` for both `kolena` and `shipyard`.
> 
> Good catch on the finer points in python's use of `repr` and `str`. If you use `print(metrics[0])` , the console should look something like
> 
> ```
> >>> from kolena.workflow import MetricsTestSample
> >>> a = dict(is_correct=True, BERT_prec=0.1)
> >>> b = MetricsTestSample(**a)
> >>> b
> MetricsTestSample()
> >>> print(b)
> MetricsTestSample(is_correct=True, BERT_prec=0.1)
> ```

Got it -- this does end up being pretty awkward:
```
>>> print(metrics[0])
(TestSample(), MetricsTestSample())
>>> metrics[0][1]
MetricsTestSample()
>>> print(metrics[0][1])
MetricsTestSample(BERT_f1=0.677471935749054, BERT_prec=0.6597466468811035, BERT_rec=0.6961759328842163, BLEU=0.0423423055737105, ROUGE_1=0.3915343915343915, ROUGE_2=0.0962566844919786, ROUGE_L=0.201058201058201, inf_to_gt_word_count=1.4133333333333333)
```

munkyshi (approved):
Approving because this is still an improvement in flexibility for retrieving data, but it's really frustrating to not have the extra fields in the `__repr__`

PR #231, opened 2023-09-07
fix: Short circuit local evaluation if remote evaluator is triggered

munkyshi:
### Linked issue(s):
N/A

Short circuits local evaluation if remote evaluator is triggered. This was previously removed in https://github.com/kolenaIO/kolena/pull/186

### DIFF
--- a/tests/integration/workflow/test_test_run.py
+++ b/tests/integration/workflow/test_test_run.py
@@ -343,10 +343,15 @@ def test__test__function_evaluator__with_skip(
     TestRun(dummy_model, dummy_test_suites[0], dummy_evaluator_function_with_config, config)
 
 
+@pytest.mark.depends(on=["test__test"])
 def test__test__remote_evaluator(
     dummy_model: Model,
     dummy_test_suites: List[TestSuite],
 ) -> None:
-    with patch("kolena.workflow.test_run.TestRun._start_server_side_evaluation") as patched:
-        test(dummy_model, dummy_test_suites[0])
-    patched.assert_called_once()
+    with patch.object(TestRun, "_start_server_side_evaluation") as remote_patched:
+        with patch.object(TestRun, "_perform_evaluation") as eval_patched:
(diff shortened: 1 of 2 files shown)


munkyshi:
Not sure if tests are set up properly -- pushed this up to get CI to run things (my local setup doesn't seem to be working)

tests/integration/workflow/test_test_run.py
```diff
@@ -341,3 +341,12 @@ def test__test__function_evaluator__with_skip(
...
     TestRun(dummy_model, dummy_test_suites[0], dummy_evaluator_function_with_config, config)
+
+
+def test__test__remote_evaluator(
+    dummy_model: Model,
+    dummy_test_suites: List[TestSuite],
+) -> None:
+    with patch("kolena.workflow.test_run.TestRun._start_server_side_evaluation") as patched:
+        test(dummy_model, dummy_test_suites[0])
+    patched.assert_called_once()
```
gordonhart: The early return won't change how many times this is called (will always be once), but will change how many times `self._perform{,_streamlined}_evaluation` are called (should be zero after this fix).

tests/integration/workflow/test_test_run.py
```diff
@@ -341,3 +341,12 @@ def test__test__function_evaluator__with_skip(
     config = [DummyConfiguration(value="skip")]
     test(dummy_model, dummy_test_suites[0], dummy_evaluator_function_with_config, config)
     TestRun(dummy_model, dummy_test_suites[0], dummy_evaluator_function_with_config, config)
+
+
+def test__test__remote_evaluator(
```
ly-kolena, line 346: Since we're using patch, we should move this to unit test.

tests/integration/workflow/test_test_run.py
munkyshi: Ah, I was depending on `self._perform{,_streamlined}_evaluation` to err if called without a real evaluator, but it would be better to be more explicit here.

ly-kolena (approved):

PR #239, opened 2023-09-19
add helper functions to export kolena data

ly-kolena:
### Linked issue(s):
Add helper functions to export kolena data to csv/json through pandas dataframe.
1. dataframe_from_csv
2. dataframe_from_json
3. dataframe_to_csv

no `dataframe_to_json` as implementing `toDict` to our `DataObject` integrates with pandas' serialization.

### DIFF
--- a/kolena/workflow/_datatypes.py
+++ b/kolena/workflow/_datatypes.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import dataclasses
-import json
 from abc import ABCMeta
 from abc import abstractmethod
 from collections import OrderedDict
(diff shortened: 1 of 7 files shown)


kolena/workflow/_datatypes.py
```diff
@@ -231,6 +231,10 @@ def deserialize_field(
                     items[key] = _try_deserialize_typed_dataobject(val)
         return cls(**items)
 
+    # integrate with pandas json deserialization
```
munkyshi, line 234: I had trouble finding the documentation on this -- sharing for anyone else: https://pandas.pydata.org/docs/user_guide/io.html#fallback-behavior

kolena/workflow/io.py
```diff
@@ -0,0 +1,108 @@
...
+    if isinstance(x, str):
+        try:
+            return json.loads(x)
+        except Exception:
+            ...
+
+    return x
+
+
+def dataframe_to_csv(df: pd.DataFrame, *args, **kwargs) -> Union[str, None]:
```
munkyshi, line 63: Does this mean that `df.to_csv` doesn't work? In what situations would it fail?

ly-kolena: `d.to_csv` would work in the sense that there would be no error writing source dataframe out to csv. However, the default serialization would be used, which I think is roughly `str`. For kolena dataobject in the dataframe, it would not be able to deserialize back, it would just be a string looking like 
```
"LabeledBoundingBox(top_left=(0.0, 0.0), bottom_right=(10.0, 10.0), width=10.0, height=10.0, area=100.0, aspect_ratio=1.0, label='foo-0')"
```

yifanwu-kolena: [for dict in the df](https://github.com/kolenaIO/shipyard/pull/1570#discussion_r1330320818), it will not generate a valid json serialization when using `to_csv`, i think that's why we need to `df_post[columns] = df_post[columns].applymap(_serialize_json)`

yifanwu-kolena (approved):

munkyshi (approved):

PR #249, opened 2023-10-03
KOL-3467 - Thresholded

diegokolena:
### Linked issue(s):

[KOL-3467](https://linear.app/kolena/issue/KOL-3467/add-thresholded-object-to-client)

- We introduce support for Dict on the MetricsTestSample. This will allow to models thresholded like this: 

 ```
 @dataclass(frozen=True)
class TestSampleMetric(DataObject):
    a: float

@dataclass(frozen=True)
class CustomMetric(MetricsTestSample):
    t1: Dict[float, TestSampleMetric]

 ```


kolena/workflow/_validators.py
```diff
@@ -52,6 +52,8 @@ def validate_data_object_type(
     data_object_type: Type[DataObject],
     supported_field_types: Optional[List[Type]] = None,
     supported_list_types: Optional[List[Type]] = None,
+    supported_dict_key_types: Optional[List[Type]] = [],
+    supported_dict_value_types: Optional[List[Type]] = [],
```
munkyshi: Is there a reason why this is `[]` while the others are `None`?

munkyshi: In general, default mutable objects should be avoided in python: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

munkyshi (commented):
Might be jumping ahead with the review here, but sharing my thoughts:

I'm a bit confused at the direction here. It looks like we're extending the base data object to allow arbitrary dicts, but then restricting that to test cases only.

- What about test samples? The original intent was to support this at the test sample level
- How would the webapp know how to handle these? Does it automatically parse any dict as if it were a thresholded object?

diegokolena:
> Might be jumping ahead with the review here, but sharing my thoughts:
> 
> I'm a bit confused at the direction here. It looks like we're extending the base data object to allow arbitrary dicts, but then restricting that to test cases only.
> 
> * What about test samples? The original intent was to support this at the test sample level

I added now for test samples too. I was to split on two PR. 

> * How would the webapp know how to handle these? Does it automatically parse any dict as if it were a thresholded object?

 Initially, I considered introducing a hidden "__type" parameter to help the webapp differentiate between standard and thresholded objects. However, after reviewing the existing codebase, I observed that this approach would be inconsistent with our current practices. Instead, I'm leaning towards a more seamless integration. The webapp could dynamically adapt its behavior when it detects a MetricsTestSample object containing a dict. What are your thoughts on this?

kolena/workflow/_validators.py
diegokolena: Yes I miss that one. Updated.

docs/advanced-usage/nesting-test-case-metrics.md
```diff
@@ -93,6 +93,7 @@ class AggregateMetrics(MetricsTestCase):
     macro_F1: float
     mAP: float
     PerClass: List[PerClassMetrics]
+    ThresholdMetrics: Dict[float, PerClassMetrics]
```
yifanwu-kolena: just want to double check that we use float as key is intentional. b/c floating number will have precision issues for instance 
```sh
>>> 0.1+0.2
0.30000000000000004

>>> x = {1.12381902839021830912830981209381092380912830982190380192839081230982190: 42}
>>> x
{1.1238190283902183: 42}
```

diegokolena: I've considered the precision issues associated with using floats as keys. However, it's important to note that access to these dictionary values will be managed exclusively by us through the webapp. We plan to control this access via a slider, which will operate on a restricted set of values. This approach mitigates the risks associated with float imprecision.

@munkyshi @yifanwu-kolena Any other use case I am missing where we can have issue?

dylangrandmont: I would consider storing the keys as strings so we can deterministically match them. I don't see any big downside with storing the thresholds in strings and it means we won't have potential precision issues in JS as well once these objects come across the network

ly-kolena: I would avoid using float as key considering we are using json for data serialization/deserialization. Handling float precision is already something we need to cautious of, and using it as key, which is not valid for json could be more landmines.

ly-kolena: Side note: is this the right place to update? This PR is about test-sample level metrics, and thresholds for our discussion is scoped to test-sample only (so far), right?

munkyshi:
@diegokolena I'm wondering if the dictionary based approach is the right change for our client API.

Concerns:
- We have typically forbidden nesting dictionaries as data object values (unless specifically wrapped with a different explicitly declared data object, such as annotations)
- We're running into pain points for explicitly using numeric keys
- The `List[DataObject]` pattern is pretty common in our client API already (and is less friction for users to adapt to), in places like `List[BoundingBox]` or `List[NestedTestCaseMetric]`

Rather than pushing forward with this approach, I wonder if it's worth revisiting the original `List[Thresholded]` approach?

diegokolena:
> List[Thresholded]

diegokolena:
> @diegokolena I'm wondering if the dictionary based approach is the right change for our client API.
> 
> Concerns:
> 
> * We have typically forbidden nesting dictionaries as data object values (unless specifically wrapped with a different explicitly declared data object, such as annotations)
> * We're running into pain points for explicitly using numeric keys
> * The `List[DataObject]` pattern is pretty common in our client API already (and is less friction for users to adapt to), in places like `List[BoundingBox]` or `List[NestedTestCaseMetric]`
> 
> Rather than pushing forward with this approach, I wonder if it's worth revisiting the original `List[Thresholded]` approach?


In this approach, the "Thresholded" dictionary will include the threshold parameter as one of its key/value pairs, as follows:

```
car_values = Thresholded([
    {"threshold": 0.01, "tp": 100, "fp": 50, "fn": 0},
    {"threshold": 0.02, "tp": 99, "fp": 50, "fn": 1},
    {"threshold": 0.03, "tp": 98, "fp": 49, "fn": 2},
    ...
    {"threshold": 0.99, "tp": 1, "fp": 1, "fn": 99},
])

```
How will we determine which key/value pair represents the threshold? Should we use a naming convention?

gordonhart:
Let me know when this is ready for review, @diegokolena!

tests/unit/workflow/test_validators.py
```diff
@@ -171,3 +172,82 @@ class NestedTester(DataObject):
...
+    )
+
+
+def test__validate_field__thresholded__no_initialize_threshold_invalid() -> None:
+    @dataclasses.dataclass(frozen=True)
+    class MyThresholdedMetrics(ThresholdedMetrics):
+        a: List[str]
+
+    with pytest.raises(TypeError):
+        MyThresholdedMetrics(a=List["1"])
```
munkyshi: Including `List` changes the intended type, correct?
```suggestion
        MyThresholdedMetrics(a=["1"])
```

tests/unit/workflow/test_validators.py
```diff
@@ -171,3 +172,82 @@ class NestedTester(DataObject):
...
+        MyThresholdedMetrics(threshold="1", a=1.0)
+
+
+def test__validate_field__thresholded__invalid_dict_field() -> None:
+    @dataclasses.dataclass(frozen=True)
+    class MyThresholdedMetrics(ThresholdedMetrics):
+        a: Dict[str, str]
+
+    with pytest.raises(TypeError):
+        MyThresholdedMetrics(a=Dict["1", "1"])
```
munkyshi: Should this include `threshold` to ensure that this doesn't fail for excluding it?
```suggestion
        MyThresholdedMetrics(threshold=1, a={"key": "value"})
```

kolena/workflow/evaluator.py
```diff
@@ -268,22 +270,25 @@ def _validate_metrics_test_sample_type(metrics_test_sample_type: Type[MetricsTes
...
         # expand e.g. List[Union[MetricsA, MetricsB]] into [MetricsA, MetricsB]
-        list_arg_types = [t for arg_type in get_args(field_type) for t in get_args(arg_type) or [arg_type]]
-        for arg_type in list_arg_types:
-            if arg_type is None:
-                raise ValueError(f"Unsupported optional metrics object in field '{field_name}'")
-            try:
-                validate_scalar_data_object_type(arg_type)
-            except ValueError:
-                raise ValueError(f"Unsupported doubly-nested metrics object in field '{field_name}'")
+        if origin is list:
```
munkyshi: The logic here doesn't seem to have changed:
```
# old
if origin is not list:  # only need to check lists, as MetricsTestCase is only allowed in lists
    continue

# current
if origin is list:
    ...
```

kolena/workflow/evaluator.py
```diff
@@ -94,6 +95,7 @@ class TestCaseMetrics(MetricsTestCase):
         macro_F1: float
         mAP: float
         PerClass: List[PerClassMetrics]
+        ThresholdClass: Dict[str, PerClassMetrics]
```
munkyshi: Usage here appears to be different now, correct?

kolena/workflow/evaluator.py
```diff
@@ -268,22 +270,25 @@ def _validate_metrics_test_sample_type(metrics_test_sample_type: Type[MetricsTes
 
 
 def _validate_metrics_test_case_type(metrics_test_case_type: Type[DataObject]) -> None:
-    validate_scalar_data_object_type(metrics_test_case_type, supported_list_types=[MetricsTestCase])
+    validate_scalar_data_object_type(
+        metrics_test_case_type,
+        supported_list_types=[MetricsTestCase, ThresholdedMetrics],
```
munkyshi: Do we need to validate that `ThresholdedMetrics` is only used as a list type?

In other words, it doesn't make sense to have something like:
```
class ObjectDetectionThresholded(ThresholdedMetrics):
    tp: List[BoundingBox]
    fp: List[BoundingBox]
    fn: List[BoundingBox]
    precision: float
    recall: float

# makes sense
class MyMetrics(MetricsTestSample):
    car_detections: List[ObjectDetectionThresholded]

# doesn't make sense
class MyMetrics2(MetricsTestSample):
    car_detections: ObjectDetectionThresholded
```

tests/unit/workflow/test_validators.py
munkyshi, line 1: We should also validate usage of `ThresholdedMetrics` as a child of other data objects:
```
class MyThresholded(ThresholdedMetrics):
    score: float

class MyMetrics(MetricsTestSample):
    thresholded_scores: List[MyThresholded]

MyMetrics(thresholded_scores=[
    MyThresholded(threshold=0.1, score=0.5),
    MyThresholded(threshold=0.5, score=0.6),
    MyThresholded(threshold=0.9, score=0.7),
])
```

kolena/workflow/evaluator.py
diegokolena: ok rollback to the original code which is equivalent to this; agree not necesary to rewrite if we are not adding any change.

tests/unit/workflow/test_validators.py
diegokolena: Added on the test_evaluator.py

tests/unit/workflow/test_validators.py
diegokolena: Updated.

tests/unit/workflow/test_validators.py
diegokolena: Updated.

kolena/workflow/evaluator.py
diegokolena: Updated.

kolena/workflow/_thresholded.py
```diff
@@ -0,0 +1,41 @@
...
+    def __new__(cls, name, bases, dct):
+        if "threshold" in dct.get("__annotations__", {}):
+            for base in bases:
+                if base.__name__ == "ThresholdedMetrics":
+                    raise TypeError(f"Subclasses of {base.__name__} cannot override 'threshold'")
+        return super().__new__(cls, name, bases, dct)
+
+
+@dataclass(frozen=True)
+class ThresholdedMetrics(TypedDataObject, metaclass=PreventThresholdOverrideMeta):
```
munkyshi: Sorry, last change here -- since this will be public to our users, lets add a docstring about usage.

kolena/workflow/_thresholded.py
munkyshi, line 1: Should this be non-private, since this is meant to be used by our users?

munkyshi (approved):
Approving with one final request

kolena/workflow/_thresholded.py
diegokolena: My rationale for using _thresholded.py was that it essentially serves as an extension of TypedDataObject, which is found in _datatypes.py.

But if you think it makes more sense to rename it to thresholded.py, that's fine with me.

kolena/workflow/_thresholded.py
diegokolena: Added

kolena/workflow/_thresholded.py
munkyshi: I think it makes sense as `thresholded.py` -- it's analogous to the other things we expect our users to see and extend (such as `MetricsTestSample` in `evaluator.py`). We don't expect our users to know of and extend `TypedDataObject` directly, so that's still private.

kolena/workflow/_thresholded.py
munkyshi: Would prefer to have this a bit more detailed / user actionable (example: [nested test case metrics](https://github.com/kolenaIO/kolena/blob/trunk/kolena/workflow/evaluator.py#L76-L104)).

I'd structure this similarly:
```
    A data class representing metrics associated with a specified threshold.

    `List[ThresholdedMetrics]` is intended for usage as a field type for a [`MetricsTestSample`][kolena.workflow.MetricsTestSample] or [`MetricsTestCase`][kolena.workflow.MetricsTestCase]. This list is intended to expected to contain threshold-specific metric values for a discrete set of specific thresholds that are consistent across [`TestSample`][kolena.workflow.TestSample]s within a test run.

    `ThresholdedMetrics` does not allow dictionary objects as field values and ensures
    that the threshold cannot be overridden.
    It should be subclassed with the relevant fields for a given workflow.

    Example usage:

    ```python
    from kolena.workflow import MetricsTestSample
    from kolena.workflow import ThresholdedMetrics
    from kolena.workflow.annotations import BoundingBox

    @dataclass(frozen=True)
    class ClassThresholdedMetrics(ThresholdedMetrics):
        precision: float
        recall: float
        f1: float

    @dataclass(frozen=True)
        car: List[ClassThresholdedMetrics]
        pedestrian: List[ClassThresholdedMetrics]

    # metrics object construction
    metric = TestSampleMetrics(
        car=[
            ClassThresholdedMetrics(
                threshold=0.3,
                precision=0.5,
                recall=0.8,
                f1=0.615,
            ),
            ClassThresholdedMetrics(
                threshold=0.4,
                precision=0.6,
                recall=0.6,
                f1=0.6,
            ),
            ClassThresholdedMetrics(
                threshold=0.5,
                precision=0.8,
                recall=0.4,
                f1=0.533,
            ),
            ...
        ],
        pedestrian=[
            ClassThresholdedMetrics(
                threshold=0.3,
                precision=0.6,
                recall=0.9,
                f1=0.72,
            ),
            ClassThresholdedMetrics(
                threshold=0.4,
                precision=0.7,
                recall=0.7,
                f1=0.7,
            ),
            ClassThresholdedMetrics(
                threshold=0.5,
                precision=0.8,
                recall=0.6,
                f1=0.686,
            ),
            ...
        ],
    )
    ```

    Raises:
        TypeError: If any field value is a dictionary.
```

PR #250, opened 2023-10-03
add telemetry flag to request headers

nankolena:
### Linked issue(s):

Fixes: KOL-3526

Currently, the telemetry flag is returned during auth and is saved as a Kolena client state. It is however not used for subsequent requests. As we start work on metrics logging, this PR updates the client to send this back in the request header using key`X-Kolena-Telemetry`.

### Testing
Using local server and added request header logging, with default setting (telemetry off):
```
'x-request-id': 'd761f5e9c472427db6923f57121b07ee', 'x-kolena-telemetry': 'False', 'content-length': '50'
```

Updated DB to enable telemetry:
```
'x-request-id': 'f6b49885163046588e03580c839473a6', 'x-kolena-telemetry': 'True', 'content-length': '50'
```

### DIFF
--- a/kolena/_utils/krequests.py
+++ b/kolena/_utils/krequests.py
@@ -65,6 +65,7 @@ def _with_default_kwargs(**kwargs: Any) -> Dict[str, Any]:
         "Authorization": f"Bearer {client_state.jwt_token}",
         "X-Request-ID": uuid.uuid4().hex,
         "User-Agent": user_agent(client_name, client_version),
+        "X-Kolena-Telemetry": str(client_state.telemetry),
     }
     return {
         **default_kwargs,
--- a/tests/unit/test_initialize.py
+++ b/tests/unit/test_initialize.py
@@ -50,8 +50,18 @@ def mock_get_token(api_token: str, *args: Any, **kwargs: Any) -> ValidateRespons
     )
 
 
+def mock_get_token_with_telemetry(api_token: str, *args: Any, **kwargs: Any) -> ValidateResponse:
+    return ValidateResponse(
+        tenant="mock tenant",
+        access_token=mock_jwt(api_token),
+        token_type="mock token type",
+        tenant_telemetry=True,
+    )
+
+
 MOCK_TOKEN = "mock token"
 FIXED_TOKEN_RESPONSE = mock_get_token(MOCK_TOKEN)
+FIXED_TOKEN_RESPONSE_WITH_TELEMETRY = mock_get_token_with_telemetry(MOCK_TOKEN)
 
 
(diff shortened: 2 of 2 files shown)


kolena/_utils/krequests.py
```diff
@@ -65,6 +65,7 @@ def _with_default_kwargs(**kwargs: Any) -> Dict[str, Any]:
         "Authorization": f"Bearer {client_state.jwt_token}",
         "X-Request-ID": uuid.uuid4().hex,
         "User-Agent": user_agent(client_name, client_version),
+        "X-Kolena-Telemetry": str(client_state.telemetry),
```
yifanwu-kolena, line 68: is it possible that we can test the serialized header?

yifanwu-kolena (approved):

kolena/_utils/krequests.py
nankolena: I did a manual test with this set to `True/False` and included in the description. Are you referring to adding unit/integ tests for this? From what I saw so far there are no direct testing on these header fields so I only updated the unit tests that stores the corresponding client state

yifanwu-kolena: yeah, i was referring to add unit/integration tests. i guess integration tests for the round trip should be able to verify this later when we have everything implemented

munkyshi (approved):
Dug into more telemetry stuff, seems like I was mistaken about attaching to things server side -- this approach seems fine.

ly-kolena:
I may miss some context: why do we need to send this value as header back to server, when server already has the info?

nankolena:
> I may miss some context: why do we need to send this value as header back to server, when server already has the info?

The server has this info at the jwt token generation step during client initialization and passes it to the client. However for subsequent requests, this info is not supplied by the client. For the server to know if each request should be tracked, without this, the server would have to query the DB again for each request.

An alternative approach is to encode this information in the jwt token on the server logic, which means that the this telemetry flag needs to be an attribute of the Identity class which I did not think was the best idea given how extensively it is used.

yifanwu-kolena:
> > I may miss some context: why do we need to send this value as header back to server, when server already has the info?
> 
> The server has this info at the jwt token generation step during client initialization and passes it to the client. However for subsequent requests, this info is not supplied by the client. For the server to know if each request should be tracked, without this, the server would have to query the DB again for each request.
> 
> An alternative approach is to encode this information in the jwt token on the server logic, which means that the this telemetry flag needs to be an attribute of the Identity class which I did not think was the best idea given how extensively it is used.

another potential approach is to cache the db results, which should decrease the db read a lot

PR #245, opened 2023-09-26
dataset inference and evalutation

ly-kolena:
### What change does this PR introduce and why?

Example updates from object_detection_2d's `seed_test_run.py`

```
def datapoint_metrics_ignored() -> Dict:
    return dict(
        TP=[],
        FP=[],
        FN=[],
        Confused=[],
        count_TP=0,
        count_FP=0,
        count_FN=0,
        count_Confused=0,
        has_TP=False,
        has_FP=False,
        has_FN=False,
        has_Confused=False,
        ignored=True,
        max_confidence_above_t=None,
        min_confidence_above_t=None,
        thresholds=[],
    )

def datapoint_metrics(bbox_matches: MulticlassInferenceMatches, thresholds: Dict[str, float]) -> Dict:
    tp = [inf for _, inf in bbox_matches.matched if inf.score >= thresholds[inf.label]]
    fp = [inf for inf in bbox_matches.unmatched_inf if inf.score >= thresholds[inf.label]]
    fn = [gt for gt, _ in bbox_matches.unmatched_gt] + [
        gt for gt, inf in bbox_matches.matched if inf.score < thresholds[inf.label]
    ]
    confused = [
        inf for _, inf in bbox_matches.unmatched_gt if inf is not None and inf.score >= thresholds[inf.label]
    ]
    non_ignored_inferences = tp + fp
    scores = [inf.score for inf in non_ignored_inferences]
    inference_labels = {inf.label for _, inf in bbox_matches.matched} | {
        inf.label for inf in bbox_matches.unmatched_inf
    }
    fields = [
        ScoredClassificationLabel(label=label, score=thresholds[label])
        for label in sorted(thresholds.keys())
        if label in inference_labels
    ]
    return dict(
        TP=tp,
        FP=fp,
        FN=fn,
        Confused=confused,
        count_TP=len(tp),
        count_FP=len(fp),
        count_FN=len(fn),
        count_Confused=len(confused),
        has_TP=len(tp) > 0,
        has_FP=len(fp) > 0,
        has_FN=len(fn) > 0,
        has_Confused=len(confused) > 0,
        ignored=False,
        max_confidence_above_t=max(scores) if len(scores) > 0 else None,
        min_confidence_above_t=min(scores) if len(scores) > 0 else None,
        thresholds=fields,
    )

def filter_inferences(
    inferences: List[ScoredLabeledBoundingBox],
    confidence_score: Optional[float] = None,
    labels: Optional[Set[str]] = None,
[...]

PR #252, opened 2023-10-05
feat: datasets evaluation public interfaces

yifanwu-kolena:
### Linked issue(s):

based on [this pr](https://github.com/kolenaIO/kolena/pull/245) with extensive testing and handling edge cases.

first iteration of public interfaces for evaluation
- `fetch_inferences`
- `fetch_evaluation_results`
- `test`

Fixes KOL-3421


tests/integration/_experimental/dataset/test_dataset.py
```diff
@@ -15,7 +15,7 @@
 
 import pandas as pd
 import pytest
-from pandas._testing import assert_frame_equal
+from pandas.testing import assert_frame_equal
```
gordonhart, line 18: Does this path change have any impact on the `pandas` versions we support?

yifanwu-kolena: here is our pandas requirements

```toml
pandas = [
    { version = ">=1.1,<1.6", python = ">=3.7,<3.11" },
    { version = ">=1.5,<1.6", python = ">=3.11" },
]
```

`pandas.testing.assert_frame_equal` exists in [1.1](http://pandas.pydata.org/pandas-docs/version/1.1/reference/api/pandas.testing.assert_frame_equal.html?highlight=_testing) even [1.0](https://pandas.pydata.org/pandas-docs/version/1.0/reference/api/pandas.testing.assert_frame_equal.html?highlight=_testing), double checked in the [codebase](https://github.com/pandas-dev/pandas/blob/1.0.x/pandas/testing.py) as well

so i think this change should have no impact on the `pandas` version we are supporting rn

kolena/_experimental/dataset/_evaluation.py
```diff
@@ -0,0 +1,225 @@
...
+            columns=["inference_id", COL_METRICS],
+        )
+    )
+
+    load_uuid = init_upload().uuid
+    upload_data_frame(df=df_metrics, batch_size=BatchSize.UPLOAD_RECORDS.value, load_uuid=load_uuid)
+    request = UploadMetricsRequest(uuid=load_uuid)
+    response = krequests.post(EvaluationPath.UPLOAD_METRICS, json=asdict(request))
+    krequests.raise_for_status(response)
+    return len(df_metrics)
```
grant-Kolena: Can we split this part into a separate function like what we had for _upload_inference so it is more reusable? The input df_metrics will be a dataframe with three columns: `inference_id`, `metrics` and `eval_config`. The logic that links inference id to the metrics relies on the order, I would have custom code to link them in galleon, and wish to use a function in the sdk to upload the dataframe.

kolena/_experimental/dataset/_dataset.py
```diff
@@ -102,7 +102,10 @@ def _to_serialized_dataframe(df: pd.DataFrame, column: str) -> pd.DataFrame:
 
 
 def _to_deserialized_dataframe(df: pd.DataFrame, column: str) -> pd.DataFrame:
-    flattened = pd.json_normalize([json.loads(r[column]) for r in df.to_dict("records")], max_level=0)
+    flattened = pd.json_normalize(
+        [json.loads(r[column]) if r[column] is not None else {} for r in df.to_dict("records")],
```
ly-kolena, line 107: I think `{}` is not right for many data types, e.g. string, numeric, custom classes.

tests/integration/_experimental/dataset/test_evaluation.py
```diff
@@ -70,7 +70,7 @@ def get_df_inf(n: int = 20) -> pd.DataFrame:
     return pd.DataFrame(records)
 
 
-def get_df_metrics(n: int = 20) -> pd.DataFrame:
+def get_df_mtr(n: int = 20) -> pd.DataFrame:
```
ly-kolena, line 63: Nit: I find this shorthand obscured, any benefit to shorten from `metrics` to `mtr`?

yifanwu-kolena: this is to distinguish with `df_metrics` from `fetch_evaluation_results`

```sh
datapoints = dp
inferences = inf
metrics = mtr
``` 
abbr. are our local df, full name ones are from sdk-server

kolena/_experimental/dataset/_dataset.py
yifanwu-kolena: can you expand this a bit?

in my understanding, we only use `_to_deserialized_dataframe` to kolena's df columns like
```py
COL_DATAPOINT = "datapoint"
COL_INFERENCE = "inference"
COL_METRICS = "metrics"
```
and all of them should be a dictionary, so i use `{}` when they are `None` from db

kolena/_experimental/dataset/_evaluation.py
yifanwu-kolena: i think maybe we can use public method `test` on galleon for ingestion with custom `infer` or `eval` functions. WDYT?

kolena/_experimental/dataset/_dataset.py
ly-kolena: You are right, what I was thinking is different.

ly-kolena: Highlighting that, pandas' default "null" value, `np.nan` would be used for all missing field (cell) in expanded dataframe, regardless of actual data types. Users would have to handle it themselves if they want to use a different value, say `None`. I think that is reasonable.

yifanwu-kolena: thanks for the highlighting and this default behaviour was checked in the integration tests `test__test__missing_metrics` and `test__test__missing_inference`

kolena/_experimental/dataset/_evaluation.py
```diff
@@ -0,0 +1,225 @@
...
+    return (
+        pd.concat(df_inference_batch)
+        if df_inference_batch
+        else pd.DataFrame(
+            columns=["datapoint_id", COL_DATAPOINT, "inference_id", COL_INFERENCE],
+        )
+    )
+
+
+def fetch_inferences(dataset: str, model: str) -> Tuple[pd.DataFrame, pd.DataFrame]:
```
ly-kolena, line 94: Nit: although still under `_experimental`, but would still be nice to include docstrings now (admittedly I missed this earlier too)?

tests/integration/_experimental/dataset/test_evaluation.py
```diff
@@ -0,0 +1,306 @@
...
+    dp_columns = ["user_dp_id", "locator", "width", "height", "city"]
+    register_dataset(dataset_name, df_dp[dp_columns])
+
+    df_inf = get_df_inf(10)
+    df_mtr = get_df_mtr(10)
+    inf_columns = ["softmax_bitmap"]
+    mtr_columns = ["score"]
+
+    with pytest.raises(IncorrectUsageError) as exc_info:
+        test(
```
ly-kolena, line 310: Are we able to catch missed inference/metrics and inform users before relying on server side to return `ServiceDataIngestion` error due to null check? That error is harder to understand imho.

On another note, is there legit cases to allow partial results? cc @y27choi

yifanwu-kolena: i don't think sdk-server side will raise `ServiceDataIngestion` due to null check b/c we will always have a dict for the inf or metrics as long as the df length is identical with df_dp.
see test `test__test__upload_none`, and we will save `{"softmax_bitmap": null }` in the db

ly-kolena: Sorry, commented on the wrong line, I was referring to line #294 in test case `test__test__invalid_data__eval_before_inf`. A similar situation would be if `test` invoked after line #230 in test case `test__test__missing_metric`.

yifanwu-kolena: just add the strict client logic to catch this and fail.
based on our [sync](https://kolena-io.slack.com/archives/C01GAKM1W5S/p1697045451173549?thread_ts=1696983622.468879&cid=C01GAKM1W5S), we might want to allow the partial results. technically we just drop all invalid rows before uploading and have some warning msg.

or we could add a parameter in the `test` func to control this behaviour, but i am a bit hesitant for it due to more complex public interface

kolena/_experimental/dataset/_evaluation.py
```diff
@@ -0,0 +1,244 @@
...
+from kolena.errors import IncorrectUsageError
+
+
+@runtime_checkable
+class EvaluationConfiguration(Protocol):
+    def _to_dict(self) -> Dict[str, Any]:
+        ...
+
+
+TYPE_EVALUATION_CONFIG = Union[EvaluationConfiguration, Dict]
```
gordonhart: `Dict[str, Any]`?

gordonhart (approved):
Happy with this as a first pass. I'm expecting that there will be some iteration on the public-facing API here, but that's fine to happen on trunk as long as this still lives in `kolena._experimental`.

kolena/_experimental/dataset/__init__.py
```diff
@@ -11,10 +11,17 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+# noreorder
```
consoull, line 14: Why no reorder?

kolena/_experimental/dataset/_evaluation.py
grant-Kolena: Feels more like a hack to me, for inference upload, we send in the function to link inference with datapoint as `infer`, and for metrics upload we send in the function to link metrics result with inference as `eval`. This would not be the most readable but it should work.

grant-Kolena: Any reason we want to force folks to always go through the test function?

yifanwu-kolena: the `test` interface is the public interface users will interactive, it has all of the functionality. 

we will just rewrite all of the `test` logic in a different way on galleon if we don't use this func in my view.

kolena/_experimental/dataset/_evaluation.py
yifanwu-kolena: just adapted the freeform cfg at [here](https://github.com/kolenaIO/kolena/pull/252/commits/a1cf901f3d2f9fc42098acca550c61179a04b3c2)

kolena/_experimental/dataset/_evaluation.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+TYPE_EVALUATION_CONFIG = Dict[str, Any]
+INFER_FUNC_TYPE = Callable[[pd.DataFrame], pd.DataFrame]
+EVAL_FUNC_TYPE = Callable[[pd.DataFrame, pd.DataFrame, Optional[TYPE_EVALUATION_CONFIG]], pd.DataFrame]
+
+
+def _fetch_dataset(dataset: str) -> pd.DataFrame:
+    df_data_batch = list(_iter_dataset_raw(dataset))
+    return pd.concat(df_data_batch) if df_data_batch else pd.DataFrame(columns=["id", "datapoint"])
```
consoull: Use `COL_DATAPOINT` instead of `"datapoint"`

kolena/_experimental/dataset/_evaluation.py
```diff
@@ -0,0 +1,229 @@
...
+    df_datapoints = _to_deserialized_dataframe(df_data, column=COL_DATAPOINT)
+    df_inferences = _to_deserialized_dataframe(df_data, column=COL_INFERENCE)
+
+    return df_datapoints, df_inferences
+
+
+def _upload_inferences(model: str, df: pd.DataFrame) -> None:
+    load_uuid = init_upload().uuid
+    upload_data_frame(
+        df=df[["id", "inference"]],
```
consoull: Use `COL_INFERENCE` instead of `"inference"`

kolena/_experimental/dataset/_evaluation.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+def _upload_metrics(
+    df: pd.DataFrame,
+    all_metrics: List[Tuple[Optional[TYPE_EVALUATION_CONFIG], pd.DataFrame]],
+) -> int:
+    df_metrics_by_eval = []
+    for eval_config, df_metrics in all_metrics:
+        df_metrics_eval = _to_serialized_dataframe(df_metrics, column=COL_METRICS)
+        df_metrics_eval["eval_config"] = json.dumps(eval_config) if eval_config is not None else None
```
consoull: Use `COL_EVAL_CONFIG` instead of `"eval_config"`

consoull (approved):

kolena/_experimental/dataset/_evaluation.py
yifanwu-kolena: just [abstract _process_metrics out of _upload_metrics](https://github.com/kolenaIO/kolena/pull/252/commits/6df2db663c2e8b821987de16c6a45fb88d9fa60c)

kolena/_experimental/dataset/__init__.py
yifanwu-kolena: i think it's might related to the api doc generation sequence control and seems like we have it for all non dummy `__init__.py`

kolena/_experimental/dataset/_evaluation.py
ly-kolena: @gordonhart did you mean to only support `Dict[str, Any]`? I initally added `EvaluationConfiguration` as duck-typing to support if users want to use `class`, particularly existing `EvaluatorConfiguration`.

yifanwu-kolena: just for the ctx, i followed [this discussion](https://linear.app/kolena/issue/KOL-3422/update-eval-config-concept-for-datasets#comment-ac77b6e7) to update

kolena/_experimental/dataset/__init__.py
gordonhart: Hmm, good question here. These `# noreorder`s may no longer be relevant since we migrated from Sphinx to Mkdocs.

PR #254, opened 2023-10-05
Metrics Glossary: updated diagrams and enabled metrics docs

y27choi:
### Linked issue(s):
Fixes KOL-3353
Fixes KOL-2625

* Updates diagrams with proper styling 
* Enabled docs that were blocked by the diagrams
* Cleaned up long lines on some of the metric docs 
* Surfaced the disabled docs to the main metrics glossary page


docs/metrics/bertscore.md
```diff
@@ -83,11 +111,16 @@ To showcase the value of BERTScore, let's consider the following candidate and r
...
+    | The sun set behind the mountains, casting a warm, orange glow across the horizon. | As the mountains obscured the
+    sun, a warm, orange glow painted the horizon. |
+    | She sipped her coffee and gazed out of the window, lost in thought on a rainy afternoon. | Lost in thought on a
+    rainy afternoon, she sipped her coffee and stared out of the window. |
+    | The adventurous explorer trekked through the dense jungle, searching for hidden treasures. | In search of hidden
+    treasures, the intrepid explorer ventured through the dense jungle. |
+    | Laughter echoed through the park as children played on the swings and slides. | Children's laughter filled the
+    park as they enjoyed the swings and slides. |
+    | The old bookstore was filled with the scent of well-worn pages, a haven for book lovers. | A haven for book lovers,
+    the old bookstore exuded the fragrance of well-read pages. |
```
billbai0102: image

I think the format got a bit messed up here. Should I create a commit to your branch to fix it?

billbai0102:
Aside from the table, everything looks good! I like the new diagrams. I'll take some time next week to figure out how to organize the metrics better than alphabetical order, so visitors know where to look for the metrics they need.

docs/metrics/bertscore.md
y27choi: Ah thanks for catching this @billbai0102.  Will push a commit shortly to fix this.

billbai0102 (approved):

munkyshi (approved):

docs/metrics/bleu.md
```diff
@@ -1,23 +1,27 @@
...
-2. The **Brevity Penalty** is applied to penalize the score for generating sentences that are less in length than the reference text. This is due to the fact that the n-gram Overlap precision tends to give disproportionately high values to candidate texts that are very short in length, but mostly contain n-grams in the reference text.
+1. The **n-gram Overlap** counts the number of 1-grams, 2-grams, 3-grams, and 4-grams of the output text that match the
+1-, ..., 4-grams in the reference text — which is analogous to a precision score for the text. The 1-gram precision
+ensures that the correct vocabulary is used, whereas the 4-gram precision ensures that the candidate text is coherent.
+2. The **Brevity Penalty** is applied to penalize the score for generating sentences that are less in length than the
+reference text. This is due to the fact that the n-gram Overlap precision tends to give disproportionately high values
+to candidate texts that are very short in length, but mostly contain n-grams in the reference text.
```
munkyshi, line 24: nit for line 31: Is it cleaner to take the quarter root of the `i-gram Precision` rather than to the `1/4` power? Also if we do use `1/4`, would it make sense to use `\frac{1}{4}` instead of `1/4`? Currently looks like it could be related to the footnote.

docs/metrics/meteor.md
munkyshi, line 1: One question that I had when reading this was how to compute the word order penalty if FMean was not `1` -- not all unigrams match. Seems like it might be good to have an example illustrating this.

munkyshi (approved):
Adding some small nits after reading through some of the new pages -- non-blocking for this PR, but some content feedback.

One other feedback that doesn't have a page is that I wonder if it would make sense to be explicit about what an `n-gram`, `unigram`, and `bigram` are -- or is that too basic for NLP to be worth defining.

docs/metrics/meteor.md
y27choi: cc @billbai0102

docs/metrics/bleu.md
y27choi: I would lean towards using `1/4` power rather than the quarter root because that exponent is equivalent to weight. 

> would it make sense to use \frac{1}{4} instead of 1/4?  
Yes, I think so. 

Updated:
image

docs/metrics/meteor.md
y27choi: @munkyshi are you asking when the # of matched unigrams is zero?  If you are asking when FMean was not `1`, then does `"the president spoke to the audience"` example answer your question?

image

y27choi:
Will merge this PR as is. It would be great if @billbai0102 can follow up with Andrew's feedback.

PR #256, opened 2023-10-06
Add fallback logic to retrieve client token during initialization 

consoull:
### Linked issue(s):
Fixes KOL-2951

Adds logic when client initialization is not explicitly passed an `api_token`. 
The client initialization will now use an api token with the following priority:

1. Value passed through the optional `api_token` argument.
2. Value stored in `KOLENA_TOKEN` environment variable.
3. Value stored in .netrc file

The purpose of this change is to improve the user experience.

### Updated Doc
Screenshot 2023-10-10 at 2 22 44 PM


tests/unit/test_initialize.py
```diff
@@ -119,6 +122,31 @@ def test__initialize__deprecated_keyword(clean_client_state: None) -> None:
         assert _client_state.jwt_token is not None
 
 
+@patch.dict("os.environ", {KOLENA_TOKEN_ENV: "abc"}, True)
```
yifanwu-kolena, line 144: just curious what's the your criterial to prefer decorator over context manager or vice versa for patching?

kolena/initialize.py
```diff
@@ -36,23 +41,51 @@ def initialize(
...
+    ```bash
+    export KOLENA_TOKEN="********"
+    ```
+    3. Store in [.netrc file](https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html)
+    ```cfg title="~/.netrc"
+    machine api.kolena.io password ********
+    ```
 
-    A session has a global scope and remains active until interpreter shutdown.
+    ```mermaid
```
yifanwu-kolena: i love this graph!

tests/unit/test_initialize.py
consoull: I prefer a decorator if the context is going to span the whole test case since it reduces the indentation level

kolena/initialize.py
```diff
@@ -96,3 +136,20 @@ def initialize(
...
+
+
+def _find_token() -> str:
+    if KOLENA_TOKEN_ENV in os.environ:
+        return os.environ[KOLENA_TOKEN_ENV]
+
+    hostname = urlparse(state._get_api_base_url()).hostname
+    try:
+        netrc_file = netrc.netrc()
+        record = netrc_file.authenticators(hostname)
```
munkyshi, line 148: Do we have any process of adding the token to the `.netrc` file? I would hope to do so automatically on successful initialize.

consoull: Not currently and I don't believe that we should (at least by default). Some users may not want their api token written to disk for security reasons.

kolena/errors.py
```diff
@@ -38,6 +38,10 @@ class InvalidClientStateError(RuntimeError, KolenaError):
     """Exception indicating that client state was invalid."""
 
 
+class MissingTokenError(KeyError, KolenaError):
+    """Exception indicating that the client could not locate an api token."""
```
gordonhart: Nit: capitalize "API"

kolena/initialize.py
```diff
@@ -11,23 +11,28 @@
...
 from kolena._utils.instrumentation import upload_log
 from kolena._utils.state import _client_state
 from kolena.errors import InputValidationError
+from kolena.errors import MissingTokenError
 
 
 def initialize(
-    api_token: str,
+    api_token: Optional[str] = None,
     *args: Any,
```
gordonhart, line 36: 🤔 could it break anything to have the `*args` positional argument catchall come after an optional arg?

kolena/initialize.py
gordonhart: Agree with @consoull's take here. I wouldn't want to automatically manipulate a systemwide file like `.netrc`

kolena/initialize.py
```diff
@@ -36,23 +41,51 @@ def initialize(
...
+    3. Store in [.netrc file](https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html)
+    ```cfg title="~/.netrc"
+    machine api.kolena.io password ********
+    ```
 
-    A session has a global scope and remains active until interpreter shutdown.
+    ```mermaid
+    flowchart TD
+        Start[Get API token]
+        Step1{{`api_token` argument provided?}}
```
gordonhart: Is there a way for us to get this to render in Mermaid as monospaced text, rather than as normal text wrapped in literal backticks?

kolena/initialize.py
```diff
@@ -36,23 +41,51 @@ def initialize(
...
     import os
     import kolena
 
     kolena.initialize(os.environ["KOLENA_TOKEN"], verbose=True)
     ```
+    2. Populate the `KOLENA_TOKEN` environment variable
+    ```bash
+    export KOLENA_TOKEN="********"
+    ```
+    3. Store in [.netrc file](https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html)
```
gordonhart: Nit: monospaced `.netrc`

kolena/initialize.py
```diff
@@ -36,23 +41,51 @@ def initialize(
...
+    ```cfg title="~/.netrc"
+    machine api.kolena.io password ********
+    ```
 
-    A session has a global scope and remains active until interpreter shutdown.
+    ```mermaid
+    flowchart TD
+        Start[Get API token]
+        Step1{{`api_token` argument provided?}}
+        Step2{{$KOLENA_TOKEN environment set?}}
```
gordonhart: Nit: I'd say "environment variable" instead of "environment"

kolena/initialize.py
```diff
@@ -96,3 +136,20 @@ def initialize(
...
+    try:
+        netrc_file = netrc.netrc()
+        record = netrc_file.authenticators(hostname)
+        if record and record[2]:
+            return record[2]
+        raise MissingTokenError
+    except Exception:
+        raise MissingTokenError(
+            f"No api token in `{KOLENA_TOKEN_ENV}` env variable or in .netrc file under {hostname}",
+        )
```
gordonhart: Is this blanket exception catch and re-raise necessary? It leads to a more convoluted stacktrace than necessary ("During handling of the above exception, another exception occurred...").

gordonhart (approved):
Various minor points to clean up but overall looking great — nice touch with the diagram as well. Preemptively approving to unblock.

kolena/initialize.py
consoull: I don't believe anything can be broken since the deprecated signature is `initialize(entity, api_token)` and we check `if len(args) == 1` to overwrite the entity sitting in api_token param with `args[0]`.
An additional change I considered was moving the `api_token` to be keyword only so it is in line with `verbose` and `proxy`. Decided against this to keep the PR trimmed down, but I do think this is the right thing to do going forward.

kolena/initialize.py
consoull: No, definitely not. I debated this portion of the code for a bit before settling on this. I'll rethink the handling here

kolena/initialize.py
consoull: I wanted the same thing but wasn't able to get it to work😞

PR #259, opened 2023-10-10
Fix incorrect max-level for dataframe json-normalize

ly-kolena:

### DIFF
--- a/kolena/_experimental/dataset/_dataset.py
+++ b/kolena/_experimental/dataset/_dataset.py
@@ -101,7 +101,7 @@ def _to_serialized_dataframe(df: pd.DataFrame) -> pd.DataFrame:
 
 
 def _to_deserialized_dataframe(df: pd.DataFrame) -> pd.DataFrame:
-    flattened = pd.json_normalize([json.loads(r[COL_DATAPOINT]) for r in df.to_dict("records")], max_level=1)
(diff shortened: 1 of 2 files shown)


kolena/_experimental/dataset/_dataset.py
```diff
@@ -101,7 +101,7 @@ def _to_serialized_dataframe(df: pd.DataFrame) -> pd.DataFrame:
 
 
 def _to_deserialized_dataframe(df: pd.DataFrame) -> pd.DataFrame:
-    flattened = pd.json_normalize([json.loads(r[COL_DATAPOINT]) for r in df.to_dict("records")], max_level=1)
+    flattened = pd.json_normalize([json.loads(r[COL_DATAPOINT]) for r in df.to_dict("records")], max_level=0)
```
yifanwu-kolena, line 104: seems like we don't want to normalize any dictionary obj

yifanwu-kolena (approved):

munkyshi (approved):
Change looks reasonable -- what's an example of something that was being flattened when it shouldn't have been?

ly-kolena:
> Change looks reasonable -- what's an example of something that was being flattened when it shouldn't have been?

```
{
  "score": 1.5,
  "classification": {
    "label": "dog"
  }
}
```
becomes
| score  | classification.label |
| ------ | ------------- |
| 1.5  | dog |

with the fix, it has the right expection
| score  | classification |
| ------ | ------------- |
| 1.5  | {"label": "dog"} |

PR #258, opened 2023-10-09
remove test sample verbiage from datasets

sandalns:
### Please check if the PR fulfills these requirements

- [ ] Include reference to internal ticket and/or GitHub issue "Fixes #NNNN" (if applicable)
- [ ] Relevant tests for the changes have been added
- [ ] Relevant docs have been added / updated

Fixes: KOL-3519

PR #263, opened 2023-10-10
add initial support for audio type

sandalns:

### DIFF
--- a/kolena/workflow/__init__.py
+++ b/kolena/workflow/__init__.py
@@ -14,6 +14,7 @@
 
 # noreorder
 from ._datatypes import DataObject
+from .test_sample import Audio
 from .test_sample import Metadata
 from .test_sample import Image
 from .test_sample import ImagePair
@@ -87,4 +88,5 @@ __all__ = [
     "TestRun",
     "test",
     "define_workflow",
+    "Audio",
 ]
--- a/kolena/workflow/__init__.py
+++ b/kolena/workflow/__init__.py
@@ -64,6 +64,7 @@ __all__ = [
     "Video",
     "Document",
     "PointCloud",
+    "Audio",
     "GroundTruth",
     "Inference",
     "Workflow",
@@ -88,5 +89,4 @@ __all__ = [
     "TestRun",
     "test",
     "define_workflow",
-    "Audio",
 ]
--- a/kolena/workflow/test_sample.py
+++ b/kolena/workflow/test_sample.py
@@ -113,6 +113,7 @@ class _TestSampleType(DataType):
     DOCUMENT = "DOCUMENT"
     COMPOSITE = "COMPOSITE"
     POINT_CLOUD = "POINT_CLOUD"
+    AUDIO = "AUDIO"
     CUSTOM = "CUSTOM"
 
     @staticmethod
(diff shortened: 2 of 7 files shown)


sandalns:
Open questions that I'm looking at:
- Are there any additional Annotation types that we should add for audio types?
- How would GroundTruths and Inferences potentially look like for for audio data?

dylangrandmont:
We might as well make sure `audio` is supported for datapoints as well

ly-kolena:
> We might as well make sure `audio` is supported for datapoints as well

Why is this highlighted? annotations/assets are not specific to datapoints/testsample.

dylangrandmont:
> Why is this highlighted? annotations/assets are not specific to datapoints/testsample.

Oh, this is my bad. I thought we were going to support audio as a new test sample `data_type` in which case we would want to support it as a datapoint `data_type` as well. Which raises the question... are we going to be supporting this as a test sample type?

sandalns:
> > Why is this highlighted? annotations/assets are not specific to datapoints/testsample.
> 
> Oh, this is my bad. I thought we were going to support audio as a new test sample `data_type` in which case we would want to support it as a datapoint `data_type` as well. Which raises the question... are we going to be supporting this as a test sample type?

I am adding audio as a test sample type, just initially only added the asset type. I will add it to datapoint data_type after https://github.com/kolenaIO/kolena/pull/258 is merged for simplicity.

kolena/workflow/test_sample.py
```diff
@@ -286,7 +287,19 @@ def _data_type(cls) -> _TestSampleType:
         return _TestSampleType.POINT_CLOUD
 
 
-_TEST_SAMPLE_BASE_TYPES = [Composite, Image, Text, BaseVideo, Document, PointCloud]
+@dataclass(frozen=True, config=ValidatorConfig)
+class Audio(TestSample):
+    """An audio file located in a cloud bucket or served at a URL."""
+
+    locator: str
+    """URL (e.g. S3, HTTPS) of the audio file."""
```
gordonhart, line 295: Are there any restrictions on the format (e.g. `.mp3`)?

sandalns: Will update this with the supported file types.

kolena/workflow/__init__.py
```diff
@@ -87,4 +88,5 @@
     "TestRun",
     "test",
     "define_workflow",
+    "Audio",
```
gordonhart: Should be ordered with the rest of the test sample types

kolena/workflow/test_sample.py
```diff
@@ -113,6 +113,7 @@ class _TestSampleType(DataType):
     DOCUMENT = "DOCUMENT"
     COMPOSITE = "COMPOSITE"
     POINT_CLOUD = "POINT_CLOUD"
+    AUDIO = "AUDIO"
```
gordonhart, line 116: Add to datapoint types as well?

gordonhart (approved):
Let's land this (after adding to datapoint types)!

kolena/workflow/test_sample.py
sandalns: done

kolena/workflow/__init__.py
sandalns: done

PR #272, opened 2023-10-12
Move initialize `api_token` from positional to keyword argument

consoull:
Context from prior PR https://github.com/kolenaIO/kolena/pull/256#discussion_r1357095496

Updates `kolena.initialize()` to prefer `api_token` as a keyword argument while maintaining backwards compatibility.


ly-kolena (approved):

gordonhart:
For public-facing docs and example integrations, let's remove the `api_token` entirely — across the board, we're doing `api_token=os.environ["KOLENA_TOKEN"]`, which is exactly why we added `$KOLENA_TOKEN` support to `initialize` in the first place.

PR #273, opened 2023-10-13
Preserve dataframe column order for relevant client operations

ly-kolena:

Preserve dataframe column order for relevant operations that client sdk perform for users. The one piece where order cannot be preserved is when snowflake is involved. Admittedly this impacts most of the meaningful operations in client sdk, however I think it still has value to maintain order as much as we can.

Also introduced some small refactor to reduce serde boilerplates.

### DIFF
--- a/kolena/_experimental/dataset/_dataset.py
+++ b/kolena/_experimental/dataset/_dataset.py
@@ -34,6 +34,7 @@ from kolena.workflow._datatypes import _deserialize_dataobject
 from kolena.workflow._datatypes import _serialize_dataobject
 from kolena.workflow._datatypes import DATA_TYPE_FIELD
 from kolena.workflow._datatypes import TypedDataObject
+from kolena.workflow.io import _dataframe_object_serde
 
 COL_DATAPOINT = "datapoint"
 TEST_SAMPLE_TYPE = "TEST_SAMPLE"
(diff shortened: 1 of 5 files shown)


kolena/_experimental/dataset/_dataset.py
```diff
@@ -90,22 +91,18 @@ def _infer_datatype(df: pd.DataFrame) -> Union[pd.DataFrame, str]:
...
 
 def _to_serialized_dataframe(df: pd.DataFrame) -> pd.DataFrame:
-    object_columns = list(df.select_dtypes(include="object").columns)
-    result = df.select_dtypes(exclude="object")
-    result[object_columns] = df[object_columns].applymap(_serialize_dataobject)
+    result = _dataframe_object_serde(df, _serialize_dataobject)
     result[DATA_TYPE_FIELD] = _infer_datatype(df)
     result[COL_DATAPOINT] = result.to_dict("records")
-    result[COL_DATAPOINT] = result[COL_DATAPOINT].apply(lambda x: json.dumps(x, sort_keys=True))
+    result[COL_DATAPOINT] = result[COL_DATAPOINT].apply(lambda x: json.dumps(x))
```
gordonhart, line 97: Does this have user-facing impact, or is object key order still lost at the persistence layer?

gordonhart (approved):

gordonhart:
Do you have any potential strategies in mind for preserving ordering at the persistence layer? This is a pretty major rough edge already, and its impact will be worsened when values aren't split across e.g. test sample/ground truth/metadata objects.

ly-kolena:
> Do you have any potential strategies in mind for preserving ordering at the persistence layer? This is a pretty major rough edge already, and its impact will be worsened when values aren't split across e.g. test sample/ground truth/metadata objects.

Without ugly hacks, I cannot think of one. One ugly hack is inject something like `_field_order=["image_id, "width", ...]` per datapoint, and drop the field after dataframe reconstruction. Non-hack but adds complexity is `dataset_schema` table.

gordonhart:
> One ugly hack is inject something like _field_order=["image_id, "width", ...] and drop the field after dataframe reconstruction.

Persisting a `field_order` column and applying this order when reading datapoints seems like a pretty clean solution, actually. Any hesitations with that?

ly-kolena:
> > One ugly hack is inject something like _field_order=["image_id, "width", ...] and drop the field after dataframe reconstruction.
> 
> Persisting a `field_order` column and applying this order when reading datapoints seems like a pretty clean solution, actually. Any hesitations with that?

Can put up a PR for that and check how it goes.

gordonhart:
> Can put up a PR for that and check how it goes.

Amazing. Longstanding ticket for reference: [KOL-2959](https://linear.app/kolena/issue/KOL-2959/preserve-order-of-generic-workflow-object-keys)

PR #274, opened 2023-10-13
Proper serialization of nested list of dataobjects in dataframe

ly-kolena:


gordonhart:
Do we want to allow this? How will the frontend handle doubly-nested objects?

ly-kolena:
> Do we want to allow this? How will the frontend handle doubly-nested objects?

I misunderstood that it was an existing capability for dataobject. On another note, given new dataframe approach, at least with current implementation, we don't enforce those checks anymore. Thoughts on this?

PR #278, opened 2023-10-13
no-ops evaluator example

nankolena:
Example commit for a no-ops semantic segmentation evaluator. Will not merge given this is for demo purposes only.

### DIFF
--- a/examples/semantic_segmentation/semantic_segmentation/workflow.py
+++ b/examples/semantic_segmentation/semantic_segmentation/workflow.py
@@ -37,12 +37,7 @@ class TestSample(Image):
 
 @dataclass(frozen=True)
 class GroundTruth(BaseGroundTruth):
-    """Ground truth type for the Semantic Segmentation workflow."""
-
-    mask: SegmentationMask
-    """
-    The ground truth [`SegmentationMask`][kolena.workflow.annotation.SegmentationMask] associated with an image.
-    """
+    gt_placeholder: int = 0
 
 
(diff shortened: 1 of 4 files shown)


examples/semantic_segmentation/semantic_segmentation/workflow.py
```diff
@@ -37,12 +37,7 @@ class TestSample(Image):
 
 @dataclass(frozen=True)
 class GroundTruth(BaseGroundTruth):
```
sandalns, line 38: If you pass BaseGroundTruth to define_workflow directly we can use `kolena.workflow.groundtruth` everywhere

nankolena: I tried that initially, but the base `GroundTruth` class has to be sub-classed: https://github.com/kolenaIO/kolena/blob/2e3417d1e8e690aaafcb0c7b1f75eb51c9f92af4/kolena/workflow/ground_truth.py#L97

munkyshi: Does this not work with `kolena.workflow.GroundTruth`?
```
>>> from kolena.workflow import GroundTruth
>>> issubclass(GroundTruth, GroundTruth)
True
```

munkyshi:
Closing this out -- we've shipped a `no_op_evaluator`: https://docs.kolena.io/reference/workflow/evaluator/#kolena.workflow.evaluator_function.no_op_evaluator

PR #281, opened 2023-10-16
Keep column order with persisted data

ly-kolena:
### Linked issue(s):
Add additional info in dataset serialization/deserialization to restore original dataframe column order

### DIFF
--- a/kolena/_experimental/dataset/_dataset.py
+++ b/kolena/_experimental/dataset/_dataset.py
@@ -90,20 +90,27 @@ def _infer_datatype(df: pd.DataFrame) -> Union[pd.DataFrame, str]:
 
 
 def _to_serialized_dataframe(df: pd.DataFrame) -> pd.DataFrame:
+    columns = [str(col) for col in df.columns]
     result = _dataframe_object_serde(df, _serialize_dataobject)
     result[DATA_TYPE_FIELD] = _infer_datatype(df)
+    result["_field_order"] = [columns for i in df.index]
     result[COL_DATAPOINT] = result.to_dict("records")
-    result[COL_DATAPOINT] = result[COL_DATAPOINT].apply(lambda x: json.dumps(x))
+    result[COL_DATAPOINT] = result[COL_DATAPOINT].apply(json.dumps)
 
     return result[[COL_DATAPOINT]]
 
 
 def _to_deserialized_dataframe(df: pd.DataFrame) -> pd.DataFrame:
-    flattened = pd.json_normalize([json.loads(r[COL_DATAPOINT]) for r in df.to_dict("records")], max_level=0)
-    flattened = flattened.loc[:, ~flattened.columns.str.endswith(DATA_TYPE_FIELD)]
+    records = [json.loads(r[COL_DATAPOINT]) for r in df.to_dict("records")]
+    field_order = records[0]["_field_order"] if records and "_field_order" in records[0] else None
+    flattened = pd.json_normalize(records, max_level=0)
+    flattened = flattened.drop(
+        [col for col in flattened.columns if col.endswith(DATA_TYPE_FIELD) or col == "_field_order"],
+        axis=1,
+    )
     result = _dataframe_object_serde(flattened, _deserialize_dataobject)
 
-    return result
+    return result[field_order] if field_order else result
 
 
 def register_dataset(name: str, df: pd.DataFrame) -> None:
--- a/kolena/_experimental/dataset/_dataset.py
+++ b/kolena/_experimental/dataset/_dataset.py
@@ -36,6 +36,8 @@ from kolena.workflow._datatypes import DATA_TYPE_FIELD
 from kolena.workflow._datatypes import TypedDataObject
 from kolena.workflow.io import _dataframe_object_serde
 
+COLUMN_ORDER_FIELD = "_field_order"
+
 COL_DATAPOINT = "datapoint"
 FIELD_LOCATOR = "locator"
 FIELD_TEXT = "text"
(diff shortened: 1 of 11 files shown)


kolena/_experimental/dataset/_dataset.py
```diff
@@ -36,6 +36,8 @@
 from kolena.workflow._datatypes import TypedDataObject
 from kolena.workflow.io import _dataframe_object_serde
 
+COLUMN_ORDER_FIELD = "_field_order"
```
gordonhart: How will this field get populated? Seems we should add it to the base `DataObject._to_dict` implementation.

ly-kolena: With dataset's dataframe approach, only `datapoint` order info is needed. If individual columns contains annotation/asset (`TypedDataObject`), as they are deserialized back to each's own class, order is irrelevant. As a result, this field can be populated from the source dataframe.

gordonhart: The Python client isn't the only place these objects are deserialized — having a deterministic column order in the app would be useful too. It's not a huge deal one way or the other, but since this approach can be applied to any data object, it seemed worth it.

Screenshot 2023-10-16 at 10 22 55 AM

Note that the frontend will need to be updated to handle this `_field_order` reserved key anyway.

ly-kolena: ~~Are you thinking to add this also to annotation/asset for UI's purpose?~~

Posted before seeing the above comment. Makes sense.

gordonhart: Yes — anything that is serialized as a dict has its order lost at the persistence layer currently. If we're adding a mechanism to order top-level Kolena objects (like datapoint), it's a quick win to extend this pattern to nested Kolena objects (like assets and annotations).

kolena/workflow/io.py
```diff
@@ -108,7 +109,7 @@ def _dataframe_object_serde(df, serde_fn: Callable[[Any], Any]):
     df_post = pd.DataFrame(columns=columns)
     for column in columns:
         if df.dtypes[column] == "object":
-            df_post[column] = df[column].apply(serde_fn)
+            df_post[column] = df[column].apply(serde_fn).replace(np.nan, None)
```
ly-kolena: Also added this logic to handle missing data, in case not all datapoints, through manual curation, have same structure.

ly-kolena: Rolled back this change as behavior between pandas 1.3 and 1.5 are different, and that of 1.3 is not desirable and don't have an alternative to avoid it.

kolena/workflow/_datatypes.py
```diff
@@ -138,11 +139,12 @@ def serialize_value(value: Any) -> Any:
             raise ValueError(f"unsupported value type: '{type(value).__name__}' (value: {value})")
 
         items = [(field.name, getattr(self, field.name)) for field in dataclasses.fields(type(self))]
-        field_names = {field.name for field in dataclasses.fields(type(self))}
+        field_names = [field.name for field in dataclasses.fields(type(self))]
```
gordonhart, line 142: Changing this to a list slows down the `key not in field_names` check a few lines after. Is there an ordered collection with O(1) lookup that we can use here?

kolena/workflow/_datatypes.py
```diff
@@ -42,6 +42,7 @@
 from kolena._utils.serde import with_serialized_columns
 from kolena._utils.validators import ValidatorConfig
 
+FIELD_ORDER_FIELD = "_field_order"
```
gordonhart, line 46: It's odd to have this reserved field underscored and the `data_type` reserved field not underscored. Do you think it's worth the effort to change to `_data_type` for datasets?

kolena/workflow/_datatypes.py
```diff
@@ -138,11 +139,13 @@ def serialize_value(value: Any) -> Any:
...
 
         items = [(field.name, getattr(self, field.name)) for field in dataclasses.fields(type(self))]
-        field_names = {field.name for field in dataclasses.fields(type(self))}
+        field_names = [field.name for field in dataclasses.fields(type(self))]
+        field_set = set(field_names)
         if _allow_extra(type(self)):
             for key, val in vars(self).items():
-                if key not in field_names and not _double_under(key):
+                if key not in field_set and not _double_under(key):
                     items.append((key, val))
```
gordonhart, line 147: Should we also append this key to `field_names`? As-is, extra values are missing from the dumped `_field_order`.

kolena/workflow/_datatypes.py
ly-kolena: We're using the same classes as those used by current `workflow`, wouldn't that break backward compatibility?

gordonhart: The migration would be more complicated than a simple rename for sure

kolena/workflow/_datatypes.py
ly-kolena: thanks for catching, updated with test.

kolena/workflow/_datatypes.py
ly-kolena: I think it's better to evaluate that in a separate PR.

gordonhart (approved):
Functionality looks good. This will need to be paired with a frontend PR to hide the reserved `_field_order` value in various places — do we have a plan for tackling that prior to next week's release?

ly-kolena:
I will follow up on relevant changes on backend side, cc @dylangrandmont for frontend efforts

Note: this should be merged only after these supporting work is done.

PR #282, opened 2023-10-16
Add no_op_evaluator function

munkyshi:
### Linked issue(s):
Related to KOL-3577

Add no_op_evaluator function

Provides users with a prebuilt function for bypassing evaluation and viewing inferences only on Kolena.

Fixes KOL-3556

### DIFF
--- a/kolena/workflow/evaluator_function.py
+++ b/kolena/workflow/evaluator_function.py
@@ -246,3 +246,20 @@ class _TestCases(TestCases):
 def _is_configured(evaluator: BasicEvaluatorFunction) -> bool:
     param_values = list(signature(evaluator).parameters.values())
     return len(param_values) == 5 and issubclass(param_values[4].annotation, EvaluatorConfiguration)
+
+
+def noop_evaluator(
+    test_samples: List[TestSample],
+    ground_truths: List[GroundTruth],
+    inferences: List[Inference],
+    test_cases: TestCases,
+) -> EvaluationResults:
+    test_sample_metrics = [BaseMetricsTestSample() for _ in test_samples]
+    test_case_metrics = [
+        (tc, MetricsTestCase())
+        for tc, *_ in test_cases.iter(test_samples, ground_truths, inferences, test_sample_metrics)
+    ]
+    return EvaluationResults(
+        metrics_test_sample=list(zip(test_samples, test_sample_metrics)),
+        metrics_test_case=test_case_metrics,
+    )
--- a/kolena/workflow/evaluator_function.py
+++ b/kolena/workflow/evaluator_function.py
@@ -248,7 +248,7 @@ def _is_configured(evaluator: BasicEvaluatorFunction) -> bool:
     return len(param_values) == 5 and issubclass(param_values[4].annotation, EvaluatorConfiguration)
 
 
-def noop_evaluator(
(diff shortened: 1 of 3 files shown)


munkyshi:
I haven't tested this out with a real workflow yet, but wanted to get it in front of eyes to discuss if this is a direction we want to go in. CC @consoull

kolena/workflow/evaluator_function.py
```diff
@@ -246,3 +246,31 @@ def _test_sample_key(ts: TestSample) -> str:
 def _is_configured(evaluator: BasicEvaluatorFunction) -> bool:
     param_values = list(signature(evaluator).parameters.values())
     return len(param_values) == 5 and issubclass(param_values[4].annotation, EvaluatorConfiguration)
+
+
+def noop_evaluator(
```
gordonhart: Slight preference here for `no_op` over `noop`. What do you think?

tests/integration/workflow/test_test_run.py
```diff
@@ -343,16 +343,22 @@ def compute_test_case_metrics(
...
     test(dummy_model, dummy_test_suites[0], dummy_evaluator_function)
     TestRun(dummy_model, dummy_test_suites[0], dummy_evaluator_function)
 
 
+def test__test__noop_evaluator(
+    dummy_model: Model,
+    dummy_test_suites: List[TestSuite],
+) -> None:
+    test(dummy_model, dummy_test_suites[0], noop_evaluator)
+    TestRun(dummy_model, dummy_test_suites[0], noop_evaluator)
```
gordonhart: Does creating this `TestRun` object do anything?

kolena/workflow/evaluator_function.py
munkyshi: no strong preference, happy to rename

tests/integration/workflow/test_test_run.py
munkyshi: not sure -- copied from the above tests, but these tests seem more along the lines of "do we run into an error on construction" more so than "does the server do the right thing". Would need to do more testing if we do want to go in the direction suggested by this PR.

consoull (approved):

tests/integration/workflow/test_test_run.py
consoull: I confirmed through an E2E test that the no-op is working as expected. Will expand on this test

munkyshi: @consoull The right place for specific functionality tests might be on the server side. Tests within `kolena` are meant to ensure that these flows don't trigger errors in communication, while the server side integration tests ensure that data state is correct.

gordonhart (approved):

gordonhart:
@consoull are we waiting on any additional validation here?

consoull:
> @consoull are we waiting on any additional validation here?

Kind of. I need to get around to writing an integration test server side to make sure we don't regress. I'm planning to merge this in once that is in place.

gordonhart:
> Kind of. I need to get around to writing an integration test server side to make sure we don't regress. I'm planning to merge this in once that is in place.

Sounds good, let me know if you could use any pointers to get started there. I'd like to have this `no_op_evaluator` in the next client release so we can communicate it externally where applicable.

PR #284, opened 2023-10-18
Remove redundant passing of `KOLENA_TOKEN` env var to initialize

consoull:
Initialize will default to checking the `KOLENA_TOKEN` environment variable if not explicitly passed an api token. Removing redundant passing of `os.environ["KOLENA_TOKEN"]` to initialize as the method will do as a fallback.

Ref https://github.com/kolenaIO/kolena/pull/272#issuecomment-1768479394


gordonhart (approved):
💅

PR #285, opened 2023-10-19
Reduce ActivationMapUploader workers to 1 for Semantic Segmentation example

sandalns:
### Linked issue(s):
Having 32 workers causes the test_run to hang until the process is manually killed. Having 1 worker allows the run to complete and based on testing is faster than 4 workers. 

### DIFF
--- a/examples/semantic_segmentation/semantic_segmentation/data_loader.py
+++ b/examples/semantic_segmentation/semantic_segmentation/data_loader.py
@@ -101,7 +101,7 @@ class DataLoader:
 
 class ActivationMapUploader:
     def __init__(self):
-        self.pool = ThreadPoolExecutor(max_workers=32)
+        self.pool = ThreadPoolExecutor(max_workers=1)
         self.futures = {}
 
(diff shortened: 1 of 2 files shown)


gordonhart (commented):
If we're running this with a single worker, why even use a threadpool? It's also surprising to me that this is faster than 4 workers — this upload process should be IO-bound, no? Slowdown with concurrency hints at this uploader being CPU-bound.

sandalns:
> If we're running this with a single worker, why even use a threadpool? It's also surprising to me that this is faster than 4 workers — this upload process should be IO-bound, no? Slowdown with concurrency hints at this uploader being CPU-bound.

Removed the Threadpool because having one with a single does not have any benefit other than allowing the masks to be uploaded in the background. When I was referering to timing was meant the time for the results to be uploaded (did not have the numbers on the masks itself). However doing this synchronously seems to preform well.

sandalns:
Closing in favor of https://github.com/kolenaIO/kolena/pull/291

PR #288, opened 2023-10-25
Rename 'ClassificationLabel' to 'Label'

gordonhart:
Leave an alias with the old name for backcompat. This is a bit of cleanup that I've wanted to get to for a while — it's a papercut to import `ClassificationLabel` when you're creating a label that isn't for classification purposes. No reason to introduce conceptual friction here by being overly prescriptive with our annotation naming.

Also clean up some bad links, and accommodate deprecations in the latest version of [mkdocs-material-insiders](https://github.com/kolenaIO/mkdocs-material-insiders), which I manually synced today.


y27choi:
Should we update the example workflows to use `Label` instead?

gordonhart:
> Should we update the example workflows to use `Label` instead?

I was debating doing that since it'll make for a lot of diff and `ClassificationLabel` is still perfectly valid. It's probably worth doing though so we put our best foot forward with these examples.

docs/core-concepts/index.md
```diff
@@ -9,7 +9,7 @@ hide:
 In this section, we'll get acquainted with the core concepts on Kolena, and learn in-depth about the various features
 offered. For a brief introduction, see the [Quickstart Guide](../quickstart.md) or the
 [Building a Workflow](../building-a-workflow.md) tutorial. For code-level API documentation, see the
-[API Reference Documentation](../reference/workflow) for the `kolena` Python client.
+[API Reference Documentation](./workflow.md) for the `kolena` Python client.
```
y27choi: Is this change intentional?  This link is not directed to API doc anymore.

mkdocs.yml
```diff
@@ -154,8 +154,8 @@ markdown_extensions:
   - attr_list
   - md_in_html
   - pymdownx.emoji:
-      emoji_index: !!python/name:materialx.emoji.twemoji
-      emoji_generator: !!python/name:materialx.emoji.to_svg
+      emoji_index: !!python/name:material.extensions.emoji.twemoji
```
y27choi, line 157: What is this change for?

y27choi (commented):
Finished reviewing and left some comments. Overall, I'm very happy that we are finally renaming `ClassificationLabel`. Thanks for working on this.

mkdocs.yml
gordonhart: Accommodating upstream changes. I synced our [mkdocs-material-insiders](https://github.com/kolenaIO/mkdocs-material-insiders) fork this morning and it spits out a ton of warnings when using the old `materialx` import.

docs/core-concepts/index.md
gordonhart: Not intentional — updated to point to API doc index file. Thanks for catching.

y27choi (approved):

PR #289, opened 2023-10-25
chore: consolidate inference and metrics to results

yifanwu-kolena:
### Linked issue(s):

Fixes KOL-3706


kolena/_experimental/dataset/_evaluation.py
```diff
@@ -184,12 +187,70 @@ def validate_data(left: pd.DataFrame, right: pd.DataFrame) -> None:
...
+
+def _get_default_infer_func(df_inf: pd.DataFrame, join_keys: list[str]) -> INFER_FUNC_TYPE:
+    def infer(datapoints: pd.DataFrame) -> pd.DataFrame:
+        inferences = datapoints[join_keys].merge(df_inf, how="left", on=join_keys)
+        return inferences
+
+    return infer
+
+
+def _get_default_eval_func(df_metrics: pd.DataFrame, join_keys: list[str]) -> EVAL_FUNC_TYPE:
```
ly-kolena: I feel this is too much black magic to understand the default mechanism. Personally I find, if choosing the two, I'd rather do the `infer` myself.

Bringing this approach to discussion, how about
```
def test(dataset, model, infer_df, metrics_df, on=union[func, str]):
```

y27choi: I like the `on` approach. I'm guessing `str` is a column name to merge on, but what is `func`?

ly-kolena: I mistakenly thought dataframe `merge/join` `on` includes `func` for complex join logic. Any case would be missing with `on` approach through `str, list of str, or array-like`?

Reference

[join](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.join.html)
> on: str, list of str, or array-like, optional
Column or index level name(s) in the caller to join on the index in other, otherwise joins index-on-index. If multiple values given, the other DataFrame must have a MultiIndex. Can pass an array as the join key if it is not already contained in the calling DataFrame. Like an Excel VLOOKUP operation.

[merge](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.merge.html)
> on: label or list
Column or index level names to join on. These must be found in both DataFrames. If on is None and not merging on indexes then this defaults to the intersection of the columns in both DataFrames.

yifanwu-kolena: just updated to expose `on` in the test func

y27choi: > Any case would be missing with on approach through str, list of str, or array-like?
I could see a results df having a `relative_path` column rather than a `locator` column requiring a bit of preprocessing before running `test` but I don't think it's worth worrying about that. 

I think those types have good coverage.

kolena/_experimental/dataset/_evaluation.py
```diff
@@ -184,12 +188,79 @@ def validate_data(left: pd.DataFrame, right: pd.DataFrame) -> None:
...
+def _get_default_eval_func(df_metrics: pd.DataFrame, on: TEST_ON_TYPE) -> EVAL_FUNC_TYPE:
+    if isinstance(on, str):
+        on = [on]
+
+    def eval(
+        datapoints: pd.DataFrame,
+        inferences: pd.DataFrame,
+        eval_config: Optional[TYPE_EVALUATION_CONFIG],
+    ) -> pd.DataFrame:
+        metrics = datapoints[on].merge(df_metrics, how="left", on=on)
```
ly-kolena: This seems to imply `inferences` is not needed for metrics computation, is this right as a default?

kolena/_experimental/dataset/_evaluation.py
```diff
@@ -184,12 +188,79 @@ def validate_data(left: pd.DataFrame, right: pd.DataFrame) -> None:
...
+    if isinstance(on, str):
+        if on not in left.columns or on not in right.columns:
+            raise IncorrectUsageError(f"column {on} doesn't exist in target dataframe")
+    elif isinstance(on, list):
+        for col in on:
+            if col not in left.columns or col not in right.columns:
+                raise IncorrectUsageError(f"column {col} doesn't exist in target dataframe")
+
+
+def _get_default_infer_func(df_inf: pd.DataFrame, on: TEST_ON_TYPE) -> INFER_FUNC_TYPE:
```
ly-kolena: Do we still this indirection?

```
def test():
   <snip>
   if infer:
      ...
      df_inferences = infer.merge(df_datapoints, on=on)
      ...
   if eval:
       ...
      <hand-wavy-similar-merge>
       ...
```

kolena/_experimental/dataset/_evaluation.py
yifanwu-kolena: i think for the simplest use case, mapping metrics only with datapoints makes sense for me, we intentionally ignored the inferences and eval_config at here.
the [partial uploading support](https://linear.app/kolena/issue/KOL-3670/sdk-support-partial-uploading-inference-and-metrics) might be able to suppress potential errors

kolena/_experimental/dataset/_evaluation.py
yifanwu-kolena: i think we still need it b/c we want to support both df and fully customized func

kolena/_experimental/dataset/_evaluation.py
ly-kolena: How does that make sense? Metrics evaluation is doing some computation based on groundtruth and model inference results. Users have to read codes here to know that we are joining on datapoint by default. cc @y27choi for more thoughts.

Side question: this line here would return both datapoint columns and metrics conlumns, right?

yifanwu-kolena: i think under the same config, dp <=> inf <=> metrics is one to one mapping. 
this is the tricky part when exposing the `on`, if we want to accommodate inf or eval_config(pass df for multiple eval_config results), we will need multiple `on` conceptually.

for the side note: it will only return metric columns due to `datapoints[on]`

Screen Shot 2023-10-26 at 11 44 46 AM

tests/integration/_experimental/dataset/test_evaluation.py
```diff
@@ -55,188 +52,169 @@ def get_df_dp(n: int = 20) -> pd.DataFrame:
...
+    assert fetched_eval_config_2 == eval_config_2
+    expected_df_result_1 = df_result_1[3:10].reset_index(drop=True)
+    expected_df_result_2 = df_result_2[3:10].reset_index(drop=True)
+    _assert_frame_equal(fetched_df_result_1, expected_df_result_1, result_columns_1)
+    _assert_frame_equal(fetched_df_result_2, expected_df_result_2, result_columns_2)
+
+
+def test__test__multiple_eval_configs__duplicate() -> None:
+    dataset_name = with_test_prefix(f"{__file__}::test__test__multiple_eval_configs__duplicate")
+    model_name = with_test_prefix(f"{__file__}::test__test__multiple_eval_configs__duplicate")
```
yifanwu-kolena, line 155: need to prevent this on sdk-server side as well

tests/integration/_experimental/dataset/test_evaluation.py
```diff
@@ -55,188 +52,169 @@ def get_df_dp(n: int = 20) -> pd.DataFrame:
...
+    _assert_frame_equal(fetched_df_dp, expected_df_dp, dp_columns)
+    _assert_frame_equal(fetched_df_result, expected_df_result, result_columns)
 
 
-def test__test__missing_inference() -> None:
-    dataset_name = with_test_prefix(f"{__file__}::test__test__missing_inference")
-    model_name = with_test_prefix(f"{__file__}::test__test__missing_inference")
+def test__test__align_manually() -> None:
+    dataset_name = with_test_prefix(f"{__file__}::test__test__align_manually")
+    model_name = with_test_prefix(f"{__file__}::test__test__align_manually")
```
yifanwu-kolena, line 89: don't think this use case is simpler for end users, we might need to require `on` column all the time, but leave it as this for now based on the current discussion

kolena/_experimental/dataset/_evaluation.py
```diff
@@ -48,189 +43,173 @@
...
     :param model: The name of the model to be used.
-    :param infer: The inference function to be used. Defaults to None.
-    :param eval: The evaluation function to be used if any. Defaults to None.
-    :param eval_configs: The evaluation configurations to be used. Defaults to None.
+    :param results: Either a DataFrame or a list of tuples, where each tuple consists of
+                    a eval configuration and a DataFrame.
+    :param on: The column(s) to merge on between datapoint DataFrame and result DataFrame
 
-    :return None: This function doesn't return anything.
+    :return None: This function doesn't return anything. It performs the test and uploads the results.
```
ly-kolena: Non-blocking personal nit: I think `None` suffices, extra explanations seems awkward.

kolena/_experimental/dataset/_evaluation.py
```diff
@@ -48,189 +43,173 @@
...
+    log.info(f"fetched {len(df_data)} for dataset {dataset}")
+
+    if isinstance(results, pd.DataFrame):
+        results = [(None, results)]
+
+    _validate_configs([cfg for cfg, _ in results])
+
+    all_results = []
+    for config, df_result_input in results:
+        log.info(f"start test with configuration {config}" if config else "start evaluation")
```
ly-kolena: Nit: use `start evaluation` for both conditions?

kolena/_experimental/dataset/_evaluation.py
```diff
@@ -48,189 +43,173 @@
...
+        single_result = _get_single_df_result(df_datapoints, df_result_input, on)
+        _validate_data(df_datapoints, single_result)
+        all_results.append((config, single_result))
+        log.info(f"completed test with configuration {config}" if config else "completed evaluation")
```
ly-kolena: Nit: Depending on which term we choose for comment above, use the same, `test` or `evaluation` here.

kolena/_api/v2/model.py
```diff
@@ -45,6 +47,18 @@ class UploadInferencesRequest:
     uuid: str
 
 
+@dataclass(frozen=True)
```
ly-kolena: `LoadInferencesRequest` and `UploadInferencesRequest` can be removed?

kolena/_experimental/dataset/_evaluation.py
```diff
@@ -48,189 +43,173 @@
...
+
+
 def test(
     dataset: str,
     model: str,
-    infer: Optional[INFER_FUNC_TYPE] = None,
-    eval: Optional[EVAL_FUNC_TYPE] = None,
-    eval_configs: Optional[Union[TYPE_EVALUATION_CONFIG, List[TYPE_EVALUATION_CONFIG]]] = None,
+    results: Union[pd.DataFrame, List[Tuple[TYPE_EVALUATION_CONFIG, pd.DataFrame]]],
+    on: TEST_ON_TYPE = None,
```
ly-kolena, line 183: Based on current implementation logic, this should not be `None`, right?

tests/integration/_experimental/dataset/test_evaluation.py
ly-kolena: ack

kolena/_experimental/dataset/_evaluation.py
```diff
@@ -48,189 +43,173 @@
...
-        df_metrics_eval[COL_EVAL_CONFIG] = json.dumps(eval_config) if eval_config is not None else None
-        df_metrics_by_eval.append(pd.concat([df["inference_id"], df_metrics_eval], axis=1))
-    df_metrics = (
-        pd.concat(df_metrics_by_eval, ignore_index=True)
-        if df_metrics_by_eval
+    df_results_by_eval = []
+    for eval_config, df_results in all_results:
+        df_results_eval = _to_serialized_dataframe(df_results, column=COL_RESULT)
+        df_results_eval[COL_EVAL_CONFIG] = json.dumps(eval_config) if eval_config is not None else None
+        df_results_by_eval.append(pd.concat([df["datapoint_id"], df_results_eval], axis=1))
```
ly-kolena, line 87: I think there's an awkwardness right now that `fetch_dataset` returns `id` column, but here we want `datapoint_id`. So a rename is required  I think: `df[["id"]].rename(columns={"id": "datapoint_id"})`.

yifanwu-kolena: yeah, we currently handle it in the source part at [_fetch_dataset](https://github.com/kolenaIO/kolena/pull/289/files#diff-b8784ed1fedc689303755b9c54e6936c49f2ab3715e1c140ac15a6dc402f9187R52-R54)

kolena/_experimental/dataset/_evaluation.py
yifanwu-kolena: currently could be None, see [here](https://github.com/kolenaIO/kolena/pull/289/files#diff-b8784ed1fedc689303755b9c54e6936c49f2ab3715e1c140ac15a6dc402f9187R168-R169) and `test__test__align_manually`

ly-kolena (approved):

tests/integration/_experimental/dataset/test_evaluation.py
yifanwu-kolena: cut the [ticket](https://linear.app/kolena/issue/KOL-3742)

PR #290, opened 2023-10-25
TimeSegment annotation type

gordonhart:
Add time-bound annotation for use in audio applications.

### DIFF
--- a/kolena/workflow/annotation.py
+++ b/kolena/workflow/annotation.py
@@ -37,7 +37,6 @@ from abc import ABCMeta
 from functools import reduce
 from typing import Dict
 from typing import List
-from typing import Optional
 from typing import Tuple
 
 from pydantic.dataclasses import dataclass
@@ -333,17 +332,18 @@ ScoredClassificationLabel = ScoredLabel
 
 @dataclass(frozen=True, config=ValidatorConfig)
 class TimeSegment(Annotation):
-    """Segment of time in the associated audio or video file."""
+    """
+    Segment of time in the associated audio or video file.
+
+    When a `group` is specified, segments are displayed on Kolena with different colors for each group present in a
+    `List[TimeSegment]`.
+    """
 
     start: float
     """Start time, in seconds, of this segment."""
 
     end: float
     """End time, in seconds, of this segment."""
-
-    # TODO: not great to have a required optional param
-    group: Optional[str]
-    """Optionally specify a group for this segment, e.g. speaker."""
 
     @staticmethod
     def _data_type() -> _AnnotationType:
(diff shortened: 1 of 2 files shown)


kolena/workflow/annotation.py
```diff
@@ -328,6 +330,53 @@ class ScoredLabel(Label):
 """Alias for [`ScoredLabel`][kolena.workflow.annotation.ScoredLabel]."""
 
 
+@dataclass(frozen=True, config=ValidatorConfig)
+class TimeSegment(Annotation):
+    """
+    Segment of time in the associated audio or video file.
+
+    When a `group` is specified, segments are displayed on Kolena with different colors for each group present in a
+    `List[TimeSegment]`.
```
gordonhart: Looking for feedback on this. Rather than adding `group` as an optional field to the annotation directly, is it sufficient to document this special behavior here?

ly-kolena: Where should I add `group` to? Is it like

```
class MyGT(GroundTruth):
   group: List[TimeSegment]

mygt = MyGT(groupd=[TimeSegment(..), TimeSegment(...)])
```

gordonhart: You'd add it to each TimeSegment instance. e.g.

```python
class GT(GroundTruth):
   transcription: List[TimeSegment]

gt = GT(transcription=[
    TimeSegment(start=0, end=5, group="speakerA", label="Hello and welcome to our guest XYZ"),
    TimeSegment(start=6, end=10, group="speakerB", label="Thank you for having me!"),
    TimeSegment(start=11, end=20, group="speakerA", label="XYZ is joining us from [...]"),
    ...
])
```

ly-kolena: what's reason not to add it as an optional field? It should be omitted from persistence if `None`.

gordonhart: Until [3.10](https://www.trueblade.com/blogs/news/python-3-10-new-dataclass-features), dataclasses could not be extended (here to `LabeledTimeSegment`, `ScoredTimeSegment`, `ScoredLabeledTimeSegment`) with required fields if the base class has optional fields.

We can circumvent this here by adding `group: Optional[str] = None` to each of those extensions, but that still blocks users from extending them with further required fields.

ly-kolena: Forgot about that. I think maybe an example code snippet could help.

gordonhart: Added example to docstring 👍

ly-kolena (approved):

y27choi (approved):
Too bad we can't add `group` field but I think the comment covers it well.

PR #293, opened 2023-10-27
rename client event names for consistency

nankolena:
### Linked issue(s):
N/A

We added event api in a [previous PR](https://github.com/kolenaIO/kolena/pull/271). One of the events was `generate-token` which represents when the python SDK client authenticates the user and generates an JWT token.

This however is easy to confuse with an existing event name `token-generated` used by the front end which represents when the API token gets refreshed. To avoid confusion, here we rename the event to `sdk-client-initialized` to be explicit.

To be consistent with other events used by the front-end, we also rename all event names to `<noun>-<verb-in-past-tense>`

### DIFF
--- a/kolena/_api/v1/event.py
+++ b/kolena/_api/v1/event.py
@@ -30,7 +30,7 @@ class EventAPI:
 
     class Event(str, Enum):
         # auth
-        GENERATE_TOKEN = "generate-token"
+        INITIALIZE_SDK_CLIENT = "initialize-sdk-client"
 
         # test case
         CREATE_TEST_CASE = "create-test-case"
--- a/kolena/_api/v1/event.py
+++ b/kolena/_api/v1/event.py
@@ -30,28 +30,28 @@ class EventAPI:
 
     class Event(str, Enum):
         # auth
-        INITIALIZE_SDK_CLIENT = "initialize-sdk-client"
+        INITIALIZE_SDK_CLIENT = "sdk-client-initialized"
 
         # test case
-        CREATE_TEST_CASE = "create-test-case"
-        LOAD_TEST_CASE = "load-test-case"
-        LOAD_TEST_CASE_SAMPLES = "load-test-case-samples"
-        EDIT_TEST_CASE = "edit-test-case"
-        INIT_MANY_TEST_CASES = "initialize-many-test-cases"
+        CREATE_TEST_CASE = "test-case-created"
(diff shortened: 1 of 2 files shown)


gordonhart:
:+1: to this change. Taking it a step further, it would be nice to standardize naming here with other platform events — elsewhere, we use the form `<noun>-<verb>`, always using the past tense for `verb`, e.g. `test-suite-renamed`. I'd recommend following that pattern and adding a prefix to client events like `sdk` for ease of filtering. That would make this event `sdk-client-initialized`.

nankolena:
> 👍 to this change. Taking it a step further, it would be nice to standardize naming here with other platform events — elsewhere, we use the form `<noun>-<verb>`, always using the past tense for `verb`, e.g. `test-suite-renamed`. I'd recommend following that pattern and adding a prefix to client events like `sdk` for ease of filtering. That would make this event `sdk-client-initialized`.

That makes sense. I initially started with `<event>-started` and `<event>-ended` but simplified them to just `<event>`. One thing to be cognizant of though, is that currently we track the event whether the call succeeded or failed (code ref [here](https://github.com/kolenaIO/kolena/blob/trunk/kolena/_utils/instrumentation.py#L129-L131)). For example, a `load-test-suite` event could fail and the metadata will track `response_error` as `NotFoundError`. In that sense, calling it `test-suite-loaded` might not be an accurate representation of what happened. Would you think that's okay, or should we in this case rename it `load-test-case-failed`? I feel like the latter will lead to too many distinct event names to organize and visualize.

kolena/_api/v1/event.py
```diff
@@ -30,28 +30,28 @@ class Path(str, Enum):
...
 
         # model
-        CREATE_MODEL = "create-model"
-        LOAD_MODEL = "load-model"
+        CREATE_MODEL = "model-created"
+        LOAD_MODEL = "model-loaded"
 
         # test run
-        EXECUTE_TEST_RUN = "execute-test-run"
+        EXECUTE_TEST_RUN = "test-run-executed"
```
gordonhart: Any objection to prefixing these with `sdk-`? Is there a method already in place for filtering SDK events out from UI events?

gordonhart (approved):

kolena/_api/v1/event.py
nankolena: There is currently no method in place and I was only manually filtering event names. No objection to what you proposed, I can add the `sdk-` prefix to all shortly

PR #296, opened 2023-10-31
Add JWTAuth to krequests to bypass netrc auth priority

consoull:
A fix for https://github.com/kolenaIO/kolena/pull/256 netrc fallback. If no auth class is passed to [requests](https://github.com/psf/requests) it will attempt to get the authorization credentials from the users netrc file and use HTTP Basic Auth https://requests.readthedocs.io/en/latest/user/authentication/#netrc-authentication.
This is undesirable for Kolena client since it is using a JWT token to sign requests https://github.com/kolenaIO/kolena/blob/trunk/kolena/_utils/krequests.py#L65. Adds an Auth class which has a JWT and attaches it to Request objects.

### DIFF
--- a/kolena/_utils/krequests.py
+++ b/kolena/_utils/krequests.py
@@ -52,11 +52,17 @@ CONNECTION_READ_TIMEOUT = 60 * 60  # Give kolena server 1 hour to respond to cli
 # Using the Retry object to configure a backoff which is not supported by using an int here.
 MAX_RETRIES = Retry(total=3, connect=3, read=0, redirect=0, status=0, backoff_factor=2)
 
+class NoOpAuth(requests.auth.AuthBase):
+    """Force requests to bypass `netrc` auth"""
+    def __call__(self, r):
+        return r
+
 
 @kolena_initialized
 def _with_default_kwargs(**kwargs: Any) -> Dict[str, Any]:
     client_state = get_client_state()
     default_kwargs = {
+        "auth": NoOpAuth(),
         "timeout": (CONNECTION_CONNECT_TIMEOUT, CONNECTION_READ_TIMEOUT),
         "proxies": client_state.proxies,
     }
--- a/kolena/_utils/krequests.py
+++ b/kolena/_utils/krequests.py
@@ -15,6 +15,7 @@ import uuid
 from http import HTTPStatus
 from typing import Any
 from typing import Dict
+from typing import Optional
 
 import requests
 from requests import HTTPError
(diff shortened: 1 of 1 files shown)


kolena/_utils/krequests.py
```diff
@@ -52,11 +52,17 @@
 # Using the Retry object to configure a backoff which is not supported by using an int here.
 MAX_RETRIES = Retry(total=3, connect=3, read=0, redirect=0, status=0, backoff_factor=2)
 
+class NoOpAuth(requests.auth.AuthBase):
+    """Force requests to bypass `netrc` auth"""
+    def __call__(self, r):
+        return r
```
gordonhart: Since we're setting the `"Authorization"` header below, rather than implementing this as a passthrough, should this actually set our Kolena bearer token? This seems to be the intended pattern (see [`auth.py` from psf/requests](https://github.com/psf/requests/blob/839a8edec37c81a18ac8332cfbd44f44e1ae6206/src/requests/auth.py#L95C28-L95C28)):

```py
class HTTPBasicAuth(AuthBase):
    """Attaches HTTP Basic Authentication to the given Request object."""

    def __init__(self, username, password):
        self.username = username
        self.password = password

    ...

    def __call__(self, r):
        r.headers["Authorization"] = _basic_auth_str(self.username, self.password)
        return r
```

consoull: Completely agree. Will update

kolena/_utils/krequests.py
```diff
@@ -53,16 +54,27 @@
...
 
 
+class JWTAuth(requests.auth.AuthBase):
+    """Attaches JWT Authorization to the given Request object"""
+
+    def __init__(self, jwt: Optional[str]):
+        self.jwt = jwt
+
+    def __call__(self, r):
+        r.headers["Authorization"] = f"Bearer {self.jwt}"
```
gordonhart, line 63: Should this still set the auth header when `self.jwt` is absent? Should `JWTAuth` even be instantiated when the `jwt` is `None`?

gordonhart: Since `_with_default_kwargs` is decorated with `@kolena_initialized`, it looks like `jwt` will always be present (`str` instead of `Optional[str]`). If that's the case, let's update the types on `JWTAuth.__init__` to reflect this non-nullability

consoull: Yup, nice catch. Updated `JWTAuth.__init__` type hints.
Unfortunately, Pyright isn't as smart and flags a false negative due to https://github.com/kolenaIO/kolena/blob/trunk/kolena/_utils/state.py#L56

gordonhart (approved):

PR #299, opened 2023-11-02
Fix event API conditions, add event action to pre-built workflows

nankolena:
### Linked issue(s):
N/A

- Remove conditions on event API
- In a previous [PR](https://github.com/kolenaIO/kolena/pull/271/files) we've added the event API's to `workflow.*`, but our pre-build workflows (`classification, detection, fr`) have specific implementations and are missing from the previous PR. In this PR, we add event actions to `detection._internal` (detection and classification inherit from this model) as well as `fr

### DIFF
--- a/kolena/detection/_internal/model.py
+++ b/kolena/detection/_internal/model.py
@@ -30,6 +30,7 @@ from pydantic import validate_arguments
 
 from kolena._api.v1.core import Model as CoreAPI
 from kolena._api.v1.detection import Model as API
+from kolena._api.v1.event import EventAPI
 from kolena._api.v1.workflow import WorkflowType
 from kolena._utils import krequests
 from kolena._utils import log
@@ -39,6 +40,7 @@ from kolena._utils.consts import BatchSize
 from kolena._utils.consts import FieldName
 from kolena._utils.endpoints import get_model_url
 from kolena._utils.frozen import Frozen
+from kolena._utils.instrumentation import with_event
 from kolena._utils.instrumentation import WithTelemetry
 from kolena._utils.serde import from_dict
 from kolena._utils.validators import validate_name
@@ -97,6 +99,7 @@ class BaseModel(ABC, Frozen, WithTelemetry):
 
     @classmethod
     @validate_arguments(config=ValidatorConfig)
+    @with_event(event_name=EventAPI.Event.CREATE_MODEL)
     def _create(cls, workflow: WorkflowType, name: str, metadata: Dict[str, Any]) -> CoreAPI.EntityData:
         request = CoreAPI.CreateRequest(name=name, metadata=metadata, workflow=workflow.value)
         res = krequests.post(endpoint_path=API.Path.CREATE.value, data=json.dumps(dataclasses.asdict(request)))
@@ -107,6 +110,7 @@ class BaseModel(ABC, Frozen, WithTelemetry):
 
     @classmethod
     @validate_arguments(config=ValidatorConfig)
+    @with_event(event_name=EventAPI.Event.LOAD_MODEL)
     def _load_by_name(cls, name: str) -> CoreAPI.EntityData:
         request = CoreAPI.LoadByNameRequest(name=name)
         res = krequests.put(endpoint_path=API.Path.LOAD_BY_NAME.value, data=json.dumps(dataclasses.asdict(request)))
--- a/kolena/detection/_internal/test_case.py
+++ b/kolena/detection/_internal/test_case.py
@@ -27,6 +27,7 @@ from pydantic import validate_arguments
 
 from kolena._api.v1.core import TestCase as CoreAPI
 from kolena._api.v1.detection import TestCase as API
+from kolena._api.v1.event import EventAPI
 from kolena._api.v1.workflow import WorkflowType
 from kolena._utils import krequests
 from kolena._utils import log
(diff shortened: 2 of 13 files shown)


kolena/detection/_internal/test_run.py
```diff
@@ -75,6 +77,7 @@ class BaseTestRun(ABC, Frozen, WithTelemetry):
     _LoadTestImagesDataFrameClass: Type[DFType] = DFType
 
     @validate_arguments(config=ValidatorConfig)
+    @with_event(event_name=EventAPI.Event.EXECUTE_TEST_RUN)
```
ly-kolena: should this be decorating `run()` method instead?

kolena/detection/_internal/model.py
```diff
@@ -97,6 +99,7 @@ def __init__(self, name: str, workflow: WorkflowType, metadata: Optional[Dict[st
 
     @classmethod
     @validate_arguments(config=ValidatorConfig)
+    @with_event(event_name=EventAPI.Event.CREATE_MODEL)
```
ly-kolena, line 102: We use these methods in `__init__` like 
```
try:
  load(..)
except NotFound:
  create(..)
```
Would this be adding confusion in analyzing the data?

kolena/detection/_internal/test_case.py
```diff
@@ -125,6 +127,7 @@ def _from(cls, response: CoreAPI.EntityData) -> "BaseTestCase":
         return obj
 
     @classmethod
+    @with_event(event_name=EventAPI.Event.CREATE_TEST_CASE)
```
grant-Kolena, line 130: I am new to this, would we be able to distinguish between CREATE_TEST_CASE Event for detection and fr?

nankolena: Not at the moment. We can expand this decorator to take additional args as a next step if the team want, but currently my understanding is that we are primarily interested in usage metrics.

grant-Kolena: Could be beneficial when analyzing the usage of these legacy stuff, but I am ok to do it after we have the need to have the requirement

grant-Kolena (approved):

kolena/detection/_internal/test_run.py
nankolena: I think you mean the `test()` method for these use cases? That's a good idea as we will be measuring durations in the outer function instead of this init. Will update shortly

kolena/detection/_internal/model.py
nankolena: Do you mean that [this block](https://github.com/kolenaIO/kolena/blob/trunk/kolena/detection/_internal/model.py#L85-L90) will do the following:
1. If model name is present - only a `load_model` event is called
2. If model name is not preset - both `load_model` and `create_model` events are called

My opinion is that this is fine given that the client indeed tried to do both actions. Additionally, for (2), if the model is not present, then the event_api will [record the exception](https://github.com/kolenaIO/kolena/blob/1d93dcaf147723aa1a6aaead5f9cb06e39ad2c9b/kolena/_utils/instrumentation.py#L130) as `NotFoundError` for us to filter as needed

ly-kolena: Non-blocking comment: It'll be a bit of effort to analyze errors for `load` operations, as to if it's actually error or "expected".

kolena/fr/test_run.py
```diff
@@ -75,6 +77,7 @@ class Data:
         model: Model.Data
         test_suites: List[TestSuite.Data]
 
+    @with_event(event_name=EventAPI.Event.EXECUTE_TEST_RUN)
```
ly-kolena: same comment here, just in case.

ly-kolena (approved):

kolena/detection/_internal/model.py
nankolena: Agreed, I raised a similar point in an [earlier pr](https://github.com/kolenaIO/kolena/pull/293#issuecomment-1782995557). It comes down to if we want to have more distinct event names (`event-failed` and `event-succeeded`) vs fewer event names but with more metadata. We can modify things as we start to analyze if needed.

kolena/fr/test_run.py
nankolena: Yup I've updated this in a [new commit](https://github.com/kolenaIO/kolena/pull/299/commits/7785a316a5453fd96aaf34fd356d9f7134d7bb19#diff-702b5740a8de89705989ecd08651cfb88d01c425ad9308050308f3186eb4b1a2) too. Thanks for double checking!

kolena/workflow/test_run.py
```diff
@@ -487,6 +486,7 @@ def _start_server_side_evaluation(self) -> None:
 
 
 @validate_arguments(config=ValidatorConfig)
+@with_event(event_name=EventAPI.Event.EXECUTE_TEST_RUN)
```
ly-kolena: Double checking, it seems like we already decorated `TestRun.run`, then we should not add here.

nankolena: Discussed offline, made [a commit](https://github.com/kolenaIO/kolena/pull/299/commits/178a639d383304c7d8ea7c459c6b3235525c55e0) to leave at the `TestRun.run` level so the event will not missed if a user is not using this `test()` interface.

PR #300, opened 2023-11-05
Update average precision score to be based on sampled PR curve

consoull:
### Linked issue(s):
Fixes KOL-3731

Updates the average precision calculation to use the sampled PR curve. This results in a less accurate AP score, but lowers the compute/memory requirements of the script.
The degree of sampling can be modified with the script argument `--eval_level`.

### DIFF
--- a/examples/semantic_segmentation/README.md
+++ b/examples/semantic_segmentation/README.md
@@ -42,9 +42,6 @@ test suite.
     The result masks will be stored under `s3://{args.out_bucket}/coco-stuff-10k/results/{args.model}` directory in
     your bucket.
 
-	> [!WARNING]
-	> This script has high memory usage(>20GB). If this is an issue, look into using https://github.com/grantjenks/python-diskcache
-
 3. An optional script, [`seed_activation_map.py`](semantic_segmentation/seed_activation_map.py) demonstrates how to generate activation maps from
 the aformentioned model inferences. If you wish to generate your own, use the `--out_bucket` argument to provide an AWS S3 bucket
 where the activation maps will be uploaded to.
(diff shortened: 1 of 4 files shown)


examples/semantic_segmentation/README.md
```diff
@@ -54,16 +51,17 @@ Run a script using the `--help` flag for more information:
...
-                        --out_bucket OUT_BUCKET
+usage: seed_test_run.py [-h] [--eval_level EVAL_LEVEL] [--model MODEL]
+                        [--test_suites TEST_SUITES [TEST_SUITES ...]] --out_bucket OUT_BUCKET
 
-optional arguments:
+options:
   -h, --help            show this help message and exit
+  --eval_level EVAL_LEVEL
+                        Number of points on precision/recall curve. Higher value will be more
+                        accurate, but require more compute
```
gordonhart: Not necessary to expose this as a command line option, IMO. The goal of these examples is to be simple, not full-featured — users take them and extend/modify them as necessary to fit their specific use cases.

gordonhart:
Do you have any perf (runtime, memory usage) results to share?

examples/semantic_segmentation/README.md
consoull: My thought process was this value tunes the scripts bottleneck and is difficult to find otherwise. We can highlight the value with a comment, but I thought this was the better approach.

examples/semantic_segmentation/README.md
```diff
@@ -54,16 +51,14 @@ Run a script using the `--help` flag for more information:
...
-optional arguments:
+options:
   -h, --help            show this help message and exit
   --model MODEL         Name of model in directory to test
   --test_suites TEST_SUITES [TEST_SUITES ...]
                         Name(s) of test suite(s) to test.
   --out_bucket OUT_BUCKET
-                        Name of AWS S3 bucket with write access to
-                        upload result masks to.
+                        Name of AWS S3 bucket with write access to upload result masks to.
```
gordonhart, line 63: Is this the actual output from `poetry run python3 semantic_segmentation/seed_test_run.py --help`?

gordonhart (approved):

examples/semantic_segmentation/README.md
consoull: Ya, I copied it over

PR #302, opened 2023-11-06
[KOL-3748] allow additional headers in client state

nankolena:
### Linked issue(s):
https://linear.app/kolena/issue/KOL-3748/update-python-sdk-to-take-additional-headers-in-client-state

Add `additional_request_headers` to client state that is used to pass in customized headers for each request.

### DIFF
--- a/kolena/_utils/krequests.py
+++ b/kolena/_utils/krequests.py
@@ -78,6 +78,8 @@ def _with_default_kwargs(**kwargs: Any) -> Dict[str, Any]:
         "User-Agent": user_agent(client_name, client_version),
         "X-Kolena-Telemetry": str(client_state.telemetry),
     }
+    if client_state.additional_request_headers:
+        default_headers.update(client_state.additional_request_headers)
     return {
         **default_kwargs,
         **kwargs,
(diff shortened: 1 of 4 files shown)


kolena/_utils/krequests.py
```diff
@@ -78,6 +78,8 @@ def _with_default_kwargs(**kwargs: Any) -> Dict[str, Any]:
         "User-Agent": user_agent(client_name, client_version),
         "X-Kolena-Telemetry": str(client_state.telemetry),
     }
+    if client_state.additional_request_headers:
+        default_headers.update(client_state.additional_request_headers)
```
ly-kolena: Nit: I think we should do the other way to ensure our own headers take priority (could use unit test)

```
headers = dict(client_state.additional_request_headers)
headers.update(default_headers)
```

nankolena: Just updated and added some tests. Let me know what you think. In the order of priority, I had for the request headers:
1. Default ones (based on `client_state.telemetry`, etc)
2. Additional ones from `client_state.additional_request_headers`
3. Custom ones passed by each call

Similarly, for `kwargs`, I updated it such that each custom call's `kwargs` will not be able to override the default ones (`auth, timeout, and proxies`)

ly-kolena (approved):

PR #305, opened 2023-11-07
Upgrade pillow for security

ly-kolena:
### Linked issue(s):

https://github.com/kolenaIO/kolena/security/dependabot/2
https://github.com/kolenaIO/kolena/security/dependabot/3
https://github.com/kolenaIO/kolena/security/dependabot/4

### Please check if the PR fulfills these requirements

- [ ] Include reference to internal ticket and/or GitHub issue "Fixes #NNNN" (if applicable)
- [ ] Relevant tests for the changes have been added
- [ ] Relevant docs have been added / updated

PR #312, opened 2023-11-08
Revert "Upgrade pillow for security"

ly-kolena:
Reverts kolenaIO/kolena#305

Encountered issues with our existing services/libraries that uses kolena, which some of them adopted old version ceiling.


sandalns:
Is this still needed?

PR #315, opened 2023-11-09
Add `load_all` method to Model

consoull:
### Linked issue(s):
Fixes KOL-3746

Adds `load_all` method to Models class, similar to the one present in TestSuites. This was done to provide more feature parity between Models and TestSuites.


kolena/workflow/model.py
```diff
@@ -159,6 +159,33 @@ def load(cls, name: str, infer: Optional[Callable[[TestSample], Inference]] = No
         log.info(f"loaded model '{name}' ({get_model_url(obj._id)})")
         return obj
 
+    @classmethod
+    @with_event(event_name=EventAPI.Event.LOAD_ALL_MODEL)
+    def load_all(
+        cls,
+        *,
+        infer: Optional[Callable[[TestSample], Inference]] = None,
```
ly-kolena: Do we need this argument?

consoull: I asked myself the same thing. We have it for `load` so I added it here as well for parity. A valid use for a mutli-model inference function be one which uses model name as part of mapping to a stored precomputed inference.

sandalns: Is this blocking this PR?

ly-kolena: The `infer` is a function that takes a sample and output an inference, the example you mentioned doesn't exactly fit. A single infer for multiple models does not make sense to me.

For parity, I'd say we should provide a list of infer functions, but that conflicts with `load_all`, as users likely wouldn't know.

I think we can go without  this arguments, and users should be able to use the results from this endpoint to `reinitialize` a model instance with the right `infer`.

gordonhart: > I think we can go without this arguments, and users should be able to use the results from this endpoint to reinitialize a model instance with the right `infer`.

This is a pretty bad devx, but I agree that it's better than applying one `infer` function to potentially many models returned from `load_all`.

```python3
models = Model.load_all(tags=["example"])
model = Model.load(models[0].name, infer=infer) if models else Model.create(new_name, infer=infer)
```

kolena/workflow/model.py
```diff
@@ -218,6 +243,7 @@ def _from_data_with_infer(
         obj.name = data.name
         obj.metadata = data.metadata
         obj.tags = data.tags
+        obj.workflow = cls.workflow
```
ly-kolena: nit: do we need this?

ly-kolena (approved):

kolena/workflow/model.py
consoull: Yes, otherwise equality checks like this fail
```
    def test__init() -> None:
        name = with_test_prefix(f"{__file__}::test__init")
        model = Model(name=name)
        loaded = Model.load(name)
    
>       assert model == loaded
E       AssertionError: assert Model(name='REEK9KB972PG /Users/consoull/workspace/kolena/tests/integration/workflow/test_model.py::test__init', metadata={}, tags=set(), workflow=Workflow(name='Dummy Workflow 🤖', test_sample_type=<class 'tests.integration.workflow.dummy.DummyTestSample'>, ground_truth_type=<class 'tests.integration.workflow.dummy.DummyGroundTruth'>, inference_type=<class 'tests.integration.workflow.dummy.DummyInference'>), infer=None) == Model(name='REEK9KB972PG /Users/consoull/workspace/kolena/tests/integration/workflow/test_model.py::test__init', metadata={}, tags=set(), infer=None)
```

consoull: Actually I think a better solution is to remove setting the workflow in `_populate_from_other` and also remove this line. Workflow is populated at the class level when getting the Model from `define_workflow`. After that it should not be set.
What do you think?

ly-kolena: I cannot reproduce where `Model()` and `Model.load()` returns differently. If I add a custom `__eq__` , `assert model == loaded` would pass.

consoull: The problem appears to be that in `Model.__init__` the [class variable `workflow`](https://github.com/kolenaIO/kolena/blob/trunk/kolena/workflow/model.py#L67) is shadowed by a `workflow` instance variable [here](https://github.com/kolenaIO/kolena/blob/trunk/kolena/workflow/model.py#L206).
When [loading a Model from data we do not set an instance variable for workflow](https://github.com/kolenaIO/kolena/blob/trunk/kolena/workflow/model.py#L209-L223) and instead use the class workflow. This is why
```
name = with_test_prefix(f"{__file__}::test__init")
model = Model(name=name)
loaded = Model.load(name)
assert model.workflow == loaded.workflow # passes
assert model == loaded
*** AssertionError
```
I believe we should update to the solution described in https://github.com/kolenaIO/kolena/pull/315#discussion_r1397679293

kolena/workflow/model.py
```diff
@@ -203,11 +228,10 @@ def _populate_from_other(self, other: "Model") -> None:
             self.name = other.name
             self.metadata = other.metadata
             self.tags = other.tags
-            self.workflow = other.workflow
```
gordonhart: Why?

consoull: https://github.com/kolenaIO/kolena/pull/315#discussion_r1397890819

consoull: It seems to me that [only workflow is meant to be a class attribute](https://github.com/kolenaIO/kolena/blob/trunk/kolena/workflow/define_workflow.py#L84) and [the other class attributes](https://github.com/kolenaIO/kolena/blob/trunk/kolena/workflow/model.py#L73-L87) should be moved to instance attributes only

consoull: Going to revert this diff and suggest in a new PR

.circleci/continue_config.yml
```diff
@@ -374,7 +374,7 @@ workflows:
       - example-test:
           matrix:
             parameters:
-              subproject: [ keypoint_detection, age_estimation, object_detection_3d, classification, question_answering, face_recognition_11 ]
```
gordonhart, line 377: @a75w-kolena please fix and re-add in your upcoming FR PR.

PR #326, opened 2023-11-15
Improvement to identification error in Speaker Diarization

billbai0102:
### Linked issue(s):
Fixes KOL-3800

Changes Identification error to include FN (previously it was only FP)

### DIFF
--- a/examples/speaker_diarization/speaker_diarization/utils.py
+++ b/examples/speaker_diarization/speaker_diarization/utils.py
@@ -240,7 +240,7 @@ def generate_missed_speech_error(gt: GroundTruth, inf: Inference) -> List[TimeSe
     gt = create_non_overlapping_segments(gt.transcription)
     inf = create_non_overlapping_segments(inf.transcription)
 
-    return [TimeSegment(start=r[0], end=r[1]) for r in generate_fp(inf, gt)]
+    return [TimeSegment(start=r[0], end=r[1]) for r in generate_error(inf, gt)]
 
 
 def calculate_tertiles(tc: TestCase, feature: str) -> dict:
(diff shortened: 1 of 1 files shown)


y27choi:
Can you add a link to the latest results?

billbai0102:
https://trunk.kolena.io/bill/results?modelIds=9&workflow=Speaker+Diarization&testSuiteId=76&evaluators=N4Ig7AbCBcoJwwIwBYBMBfdQ

Latest results can be found here, will seed to app/demo once approved.

examples/speaker_diarization/speaker_diarization/utils.py
```diff
@@ -204,10 +209,11 @@ def generate_identification_error(gt: GroundTruth, inf: Inference) -> List[TimeS
         gt_no = create_non_overlapping_segments(gt.transcription, id)
         inf_no = create_non_overlapping_segments(inf.transcription, id)
         res.extend(generate_fp(gt_no, inf_no))
+        res.extend(generate_fp(inf_no, gt_no))
```
y27choi: What are we trying to capture here by flipping the `inf` and `gt`?

billbai0102: fp(gt, inf) calculates the FP, whereas fp(inf, gt) calculates the FN (which only works since we're working with intervals) - originally I had defined the error as just FP but it's more correct to include both FP and FN

y27choi: Got it. Do you think it makes sense to rename the function to be more relevant?

billbai0102: Hmm, I think generate_identification_error is a descriptive enough function name as it is. Maybe we can create a generate_fn(gt, inf) function that just returns generate_fp(inf, gt), but that might be overkill? What are your thoughts?

y27choi: I think `generate_identification_error` is fine.  I just thought that `generate_fp` can be confusing when it can be used to extract fp and fn.  Why don't we call it `generate_error_segments` and leave a comment in its function header that explains two different usages?

y27choi (commented):
Line 243 needs an update as well but other than that it looks great.

y27choi (approved):

PR #328, opened 2023-11-16
invalidate doc cdn cache on deploy

grant-Kolena:
- invoke a lambda function to invalidate cdn cache when doing prod doc deploy

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -64,6 +64,3 @@ jobs:
         run: |
           poetry run mkdocs build --verbose --strict --config-file mkdocs.insiders.yml
           aws s3 sync ./site "s3://docs.kolena.io"
-        env:
-          DD_RUM_CLIENT_TOKEN: ${{ vars.DD_RUM_CLIENT_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -64,6 +64,7 @@ jobs:
         run: |
           poetry run mkdocs build --verbose --strict --config-file mkdocs.insiders.yml
           aws s3 sync ./site "s3://docs.kolena.io"
+          aws lambda invoke --function-name docs-cdn-invalidation-lambda --invocation-type RequestResponse --no-cli-pager --region us-west-2 /dev/stdout
```
ly-kolena: Have not tried, but would this also work (or any concern) ?

```
aws cloudfront create-invalidation --distribution-id {{ secret.distribution_id }} --paths "/*"
```

grant-Kolena: This should work, let me check. Original thinking was use lambda to only invalidate the file that changed, but that's very hard to do. We can cut the middle man

ly-kolena (approved):
Looks good to me if you have manually verified this works.

PR #225, opened 2023-09-01
Change Prebuilt OD For Polygons

marko-polo-cheno:

PR #338, opened 2023-11-22
Add experimental prebuilt Instance Segmentation workflow

consoull:
### Linked issue(s):
Closes KOL-3869

Adds an experimental Instance Segmentation workflow based off of the
experimental Object Detection workflow. The primary difference is that
Instance Segmentation deals with Polygons while Object Detection deals
with Bounding Boxes.

### DIFF
--- a/kolena/_experimental/instance_segmentation/__init__.py
+++ b/kolena/_experimental/instance_segmentation/__init__.py
@@ -1,0 +1,47 @@
+# Copyright 2021-2023 Kolena Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# noreorder
+from .workflow import ClassMetricsPerTestCase
+from .workflow import GroundTruth
+from .workflow import Inference
+from .workflow import Model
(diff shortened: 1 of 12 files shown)


gordonhart:
Can we make this without copying over a bunch of code from `kolena._experimental.object_detection`? Should be able to just redefine `workflow.py` with polygons, then import everything else from OD. Some minor type changes may be required, but that will make it much much easier to maintain, and reduce the diff here down to a few hundred lines at most.

y27choi:
> Can we make this without copying over a bunch of code from `kolena._experimental.object_detection`? Should be able to just redefine `workflow.py` with polygons, then import everything else from OD. Some minor type changes may be required, but that will make it much much easier to maintain, and reduce the diff here down to a few hundred lines at most.

There is already an [open PR](https://github.com/kolenaIO/kolena/pull/225) for adding polygon support to our pre-built OD workflow which was put on hold due to an API change.

gordonhart:
> There is already an open PR for adding polygon support to our pre-built OD workflow which was put on hold due to an API change.

What I have in mind is the middle ground between this PR and the other PR:

1. Add a new `kolena._experimental.instance_segmentation` workflow
2. Update internal definitions in `object_detection` (evaluator, utils) to accommodate bounding box or polygon geometry
3. Share this polymorphic evaluator implementation between `object_detection` and `instance_segmentation`, removing the need to copy/paste lots of complicated code and tests

This achieves the objectives of adding IS without breaking the API, and reusing as much code as possible.

consoull:
Where is the breaking API change in https://github.com/kolenaIO/kolena/pull/225?

gordonhart:
Renaming `bboxes` to `objects` in [`workflow.py`](https://github.com/kolenaIO/kolena/pull/225/files#diff-8abe933b3194970f1a4e03a8da5f0439a6499a65e092df92dddf5d03b572c9e0R55).

consoull:
Closed in favour of https://github.com/kolenaIO/kolena/pull/225

PR #339, opened 2023-11-23
Remove Model class attributes that are shadowed

consoull:
Model has class attributes defined for name, metadata, tags, infer,
and _id which are shadowed as instance attributes
during initialization. This change removes these class attributes since
they belong as instance attributes.

### DIFF
--- a/kolena/workflow/model.py
+++ b/kolena/workflow/model.py
@@ -217,7 +217,6 @@ class Model(Frozen, WithTelemetry, metaclass=ABCMeta):
                 yield test_sample, ground_truth, inference
         log.info(f"loaded inferences from model '{self.name}' on test case '{test_case.name}'")
 
-
     @classmethod
     def _from_data(
         cls,
--- a/kolena/workflow/model.py
+++ b/kolena/workflow/model.py
@@ -70,24 +70,6 @@ class Model(Frozen, WithTelemetry, metaclass=ABCMeta):
     [`define_workflow`][kolena.workflow.define_workflow.define_workflow].
     """
 
-    name: str
-    """Unique name of the model."""
-
-    metadata: Dict[str, Any]
-    """Unstructured metadata associated with the model."""
-
-    tags: Set[str]
-    """Tags associated with this model."""
-
-    infer: Optional[Callable[[TestSample], Inference]]
-    """
-    Function transforming a [`TestSample`][kolena.workflow.TestSample] for a workflow into an
-    [`Inference`][kolena.workflow.Inference] object. Required when using [`test`][kolena.workflow.test] or
-    [`TestRun.run`][kolena.workflow.TestRun.run].
-    """
-
-    _id: int
-
     @telemetry
     def __init_subclass__(cls) -> None:
         if not hasattr(cls, "workflow"):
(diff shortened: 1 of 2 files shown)


gordonhart:
Is there any functional motivation for this change, or just cleanliness? Initially preferred to define attributes at the class level for readability.

kolena/workflow/model.py
```diff
@@ -114,7 +96,20 @@ def __init__(
...
+            self.metadata = loaded.metadata
+            """Unstructured metadata associated with the model."""
+            self.tags = loaded.tags
+            """Tags associated with this model."""
+            self.infer = loaded.infer
+            """
+            Function transforming a [`TestSample`][kolena.workflow.TestSample] for a workflow into an
+            [`Inference`][kolena.workflow.Inference] object. Required when using [`test`][kolena.workflow.test] or
+            [`TestRun.run`][kolena.workflow.TestRun.run].
+            """
```
gordonhart, line 112: Example of undesirable docs diff here — type hint is missing, and `loaded.<x>` is noisy. Both are correctly labeled as instance attributes:

| Trunk | Branch |
| --- | --- |
| Screenshot 2023-11-27 at 12 19 13 PM |  Screenshot 2023-11-27 at 12 19 20 PM |

consoull: Ah good catch. I missed this when I took a look at the documentation.
Seems to be a limitation for mkdocstrings Sphinx support https://mkdocstrings.github.io/griffe/docstrings/#sections.

Adding `__init__` params as shown on https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#an-example-class-with-docstrings renders as
Screenshot 2023-11-28 at 4 19 39 PM

I think we should backlog moving towards google or numpy style comments since mkdocstrings support is lacking for Sphinx.

consoull:
> Is there any functional motivation for this change, or just cleanliness? Initially preferred to define attributes at the class level for readability.

A bit of both. Models inherit from [`Frozen`](https://github.com/kolenaIO/kolena/blob/7dce11f41a3327565616360ee773210359b20d39/kolena/_utils/frozen.py#L49) whose `__eq__` compares `__dict__`. These differ when loading models since the Model created with `__init__` shadows the class attribute `workflow` while the loaded model does not. An example of the issue in this comment https://github.com/kolenaIO/kolena/pull/315#discussion_r1397890819.
Another potential fix is to shadow the workflow when loading the model, but I prefer getting rid of the shadowing altogether since it is confusing/bad practice.

kolena/workflow/model.py
gordonhart: We just migrated away from Sphinx to MkDocs! There are tradeoffs with all of these systems — I don't think we're going to move back any time soon 🙂

consoull: I'm referring to mkdocstring `docstring_style` which [is set to sphinx](https://github.com/kolenaIO/kolena/blob/7dce11f41a3327565616360ee773210359b20d39/mkdocs.yml#L191), but is poorly supported https://mkdocstrings.github.io/griffe/docstrings/#sections

PR #342, opened 2023-11-27
Docs: disable instant nav

gordonhart:
Disable `navigation.instant` mode, which appears to have bugs when rendering math formulas (see [related discussion](https://github.com/squidfunk/mkdocs-material/discussions/6138)).

Fixes KOL-3848

### DIFF
--- a/docs/metrics/fpr.md
+++ b/docs/metrics/fpr.md
@@ -27,7 +27,7 @@ FPR is often used in conjuction with [TPR (recall)](./recall.md). By measuring b
 of a model's performance can be drawn.
 
 <div class="grid cards" markdown>
-- :kolena-manual-16: API Reference: [`FPR` ↗][kolena.workflow.metrics.fpr]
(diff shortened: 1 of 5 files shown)


y27choi:
I also verified that it is fixed locally. 
A couple of things that I noticed:

Some formula don't load (even after a refresh):
Screen Shot 2023-11-27 at 8 59 44 AM

Some multi-line equations have line numbers on the right side:
Screen Shot 2023-11-27 at 8 55 39 AM

gordonhart:
Thanks for reviewing, @y27choi. Given that there are meaningful differences in display, I think I'll switch back to MathJax and just disable `navigation.instant` for now. Based on discussions and issues in mkdocs-material, there's almost certainly a way to get math to work with instant navigation, I just haven't had the time to really dig into it.

gordonhart:
Noting that disabling instant mode introduces this undesirable scrollbar flicker:

Nov-28-2023 08-29-06

Definitely a better problem to have than broken formulas.

munkyshi (approved):

PR #346, opened 2023-12-04
Add support for composite datasets

consoull:
### Linked issue(s):
Towards KOL-3558

Adds support for composite datasets with `DATAPOINT/COMPOSITE` value and tags nested datasets with their corresponding dataset type.

The below screenshot illustrates this change. The dataset consists of nested text datasets `a` and `b`. The top level `data_type` is marked `DATASET/COMPOSITE` while the nested datasets `a` and `b` have their `data_type` marked `DATASET/TEXT`
Screenshot 2023-12-11 at 1 19 19 PM

### DIFF
--- a/kolena/_experimental/dataset/_dataset.py
+++ b/kolena/_experimental/dataset/_dataset.py
@@ -107,7 +107,7 @@ def _format_composite(df: pd.DataFrame, prefix: str, sep: str) -> None:
         )
     if sep in prefix:
         raise InputValidationError(
-            f"More than one delimeter '{sep}' in {prefix=}.",
+            f"More than one delimeter '{sep}' in prefix: '{prefix}'.",
         )
 
     composite_columns = df.filter(regex=rf"^{prefix}", axis=1).columns.to_list()
--- a/kolena/_experimental/dataset/_dataset.py
+++ b/kolena/_experimental/dataset/_dataset.py
@@ -86,7 +86,11 @@ def _infer_datatype_value(x: str) -> str:
 
 def add_datatype(df: pd.DataFrame) -> None:
     """Adds `data_type` column(s) to input DataFrame."""
-    prefixes = {column.rsplit(sep=".", maxsplit=1)[0] + "." for column in df.columns.values if "." in column}
+    prefixes = {
+        column.rsplit(sep=".", maxsplit=1)[0] + "."
+        for column in df.columns.values
+        if isinstance(column, str) and "." in column
+    }
     if prefixes:
         df[DATA_TYPE_FIELD] = DatapointType.COMPOSITE.value
         for prefix in prefixes:
--- a/kolena/_experimental/dataset/_dataset.py
+++ b/kolena/_experimental/dataset/_dataset.py
@@ -113,7 +113,7 @@ def _format_composite(df: pd.DataFrame, prefix: str, sep: str) -> None:
     composite_columns = df.filter(regex=rf"^{prefix}", axis=1).columns.to_list()
     composite = df.loc[:, composite_columns]
     df.drop(columns=composite_columns, inplace=True)
-    composite.rename(columns=lambda col: col.removeprefix(prefix + sep), inplace=True)
+    composite.rename(columns=lambda col: col.split(sep)[-1], inplace=True)
     composite[DATA_TYPE_FIELD] = _infer_datatype(composite)
     df[prefix] = composite.to_dict("records")
     df[prefix] = df[prefix].apply(json.dumps)
--- a/kolena/_experimental/dataset/_dataset.py
+++ b/kolena/_experimental/dataset/_dataset.py
@@ -41,6 +41,7 @@ from kolena.workflow.io import _dataframe_object_serde
 
 FIELD_LOCATOR = "locator"
 FIELD_TEXT = "text"
+SEP = "."
 
 
 class DatapointType(str, Enum):
@@ -85,35 +86,35 @@ def _infer_datatype_value(x: str) -> str:
     return DatapointType.TABULAR.value
 
 
-def _add_datatype(df: pd.DataFrame, sep: str = ".") -> None:
+def _add_datatype(df: pd.DataFrame) -> None:
     """Adds `data_type` column(s) to input DataFrame."""
     prefixes = {
-        column.rsplit(sep=sep, maxsplit=1)[0]
+        column.rsplit(sep=SEP, maxsplit=1)[0]
         for column in df.columns.values
-        if isinstance(column, str) and sep in column
+        if isinstance(column, str) and SEP in column
     }
     if prefixes:
         df[DATA_TYPE_FIELD] = DatapointType.COMPOSITE.value
         for prefix in prefixes:
-            _format_composite(df, prefix, sep)
+            _format_composite(df, prefix)
     else:
         df[DATA_TYPE_FIELD] = _infer_datatype(df)
 
 
-def _format_composite(df: pd.DataFrame, prefix: str, sep: str) -> None:
+def _format_composite(df: pd.DataFrame, prefix: str) -> None:
     if prefix in df.columns:
         raise InputValidationError(
             f"Conflicting column '{prefix}' encountered when formatting composite dataset.",
         )
-    if sep in prefix:
+    if SEP in prefix:
         raise InputValidationError(
-            f"More than one delimeter '{sep}' in prefix: '{prefix}'.",
+            f"More than one delimeter '{SEP}' in prefix: '{prefix}'.",
         )
 
     composite_columns = df.filter(regex=rf"^{prefix}", axis=1).columns.to_list()
     composite = df.loc[:, composite_columns]
     df.drop(columns=composite_columns, inplace=True)
-    composite.rename(columns=lambda col: col.split(sep)[-1], inplace=True)
+    composite.rename(columns=lambda col: col.split(SEP)[-1], inplace=True)
     composite[DATA_TYPE_FIELD] = _infer_datatype(composite)
     df[prefix] = composite.to_dict("records")
     df[prefix] = df[prefix].apply(json.dumps)
--- a/kolena/_experimental/dataset/_dataset.py
+++ b/kolena/_experimental/dataset/_dataset.py
@@ -152,7 +152,7 @@ def _flatten_composite(df: pd.DataFrame) -> pd.DataFrame:
     for key, value in df.iloc[0].items():
         if isinstance(value, dict) and DatapointType.has_value(value.get(DATA_TYPE_FIELD)):
             flattened = pd.json_normalize(df[key]).rename(
-                columns=lambda col: f"{key}.{col}",
+                columns=lambda col: f"{key}{SEP}{col}",
             )
             df = df.join(flattened)
             df.drop(columns=[key], inplace=True)
--- a/kolena/_experimental/dataset/_dataset.py
+++ b/kolena/_experimental/dataset/_dataset.py
@@ -54,6 +54,10 @@ class DatapointType(str, Enum):
     TEXT = "DATAPOINT/TEXT"
     VIDEO = "DATAPOINT/VIDEO"
 
+    @classmethod
+    def has_value(cls, item) -> bool:
+        return item in cls.__members__.values()
+
 
 _DATAPOINT_TYPE_MAP = {
     "image": DatapointType.IMAGE.value,
@@ -146,7 +150,7 @@ def _to_deserialized_dataframe(df: pd.DataFrame, column: str) -> pd.DataFrame:
 
 def _flatten_composite(df: pd.DataFrame) -> pd.DataFrame:
     for key, value in df.iloc[0].items():
-        if isinstance(value, dict) and DATA_TYPE_FIELD in value:
+        if isinstance(value, dict) and DatapointType.has_value(value.get(DATA_TYPE_FIELD)):
             flattened = pd.json_normalize(df[key]).rename(
                 columns=lambda col: f"{key}.{col}",
             )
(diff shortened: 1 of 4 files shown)


kolena/_experimental/dataset/_dataset.py
```diff
@@ -83,10 +84,21 @@ def _infer_datatype_value(x: str) -> str:
     return DatapointType.TABULAR.value
 
 
-def _infer_datatype(df: pd.DataFrame) -> Union[pd.DataFrame, str]:
-    if FIELD_LOCATOR in df.columns:
-        return df[FIELD_LOCATOR].apply(_infer_datatype_value)
-    elif FIELD_TEXT in df.columns:
+def add_datatype(df: pd.DataFrame) -> None:
```
consoull: Marked public since this function is utilized by backend. This function opts to modify the passed DataFrame so the prefix parsing logic can be shared

tests/unit/_experimental/dataset/test_dataset.py
```diff
@@ -52,6 +53,43 @@ def test__infer_datatype_value(uri: str, expected: str) -> None:
...
+            locator=["s3://test.pdf", "https://test.png", "/home/test.mp4", "/tmp/test.pcd"],
+        ),
+    )
+    add_datatype(df)
+    assert df[DATA_TYPE_FIELD].equals(
+        pd.Series([DatapointType.DOCUMENT, DatapointType.IMAGE, DatapointType.VIDEO, DatapointType.POINT_CLOUD]),
+    )
+
+
+def test__add_datatype__composite() -> None:
```
yifanwu-kolena, line 72: in my understanding, composite supports arbitrary amount of sub datapoints. should we test for 1 sub dp and more than 2 dps as well?

consoull: Ya, your understanding is correct. I will add a test for 1 sub datapoint. Do you see benefit in having a test for 3 datapoints over 2?

consoull: I'm going to add some edge cases which are missing from this test (eg. doubly nested composites)

yifanwu-kolena: yeah, agree that 3 or 2 sub dp have no difference

kolena/_experimental/dataset/_dataset.py
ly-kolena: I kinda disagree with this. This additional info is for internal use only, we don't expose this to users or want them to think about it. When users get retrieve the data back, they also don't see this temporary internal column.

kolena/_experimental/dataset/_dataset.py
```diff
@@ -83,6 +86,40 @@ def _infer_datatype_value(x: str) -> str:
...
+        if isinstance(column, str) and sep in column
+    }
+    if prefixes:
+        df[DATA_TYPE_FIELD] = DatapointType.COMPOSITE.value
+        _format_composite(df, prefixes, sep)
+    else:
+        df[DATA_TYPE_FIELD] = _infer_datatype(df)
+
+
+def _format_composite(df: pd.DataFrame, prefixes: Set[str], sep: str) -> None:
```
consoull: Not really a fan of the arguments here, but `_add_datatype` method was getting bulky. I'll likely update this to work on a single composite (prefix at a time and keep the for loop in `_add_datatype`

tests/unit/_experimental/dataset/test_dataset.py
```diff
@@ -52,6 +57,89 @@ def test__infer_datatype_value(uri: str, expected: str) -> None:
...
+
+    composite_dataset = pd.DataFrame(
+        {
+            "too.many.dots": [i for i in range(5)],
+        },
+    )
+    with pytest.raises(InputValidationError):
+        _add_datatype(composite_dataset)
+
+
```
ly-kolena, line 150: I added this test

```
def test__datapoint_dataframe__serde_composite() -> None:
    datapoints = [
        {
            "category": "outdoor" if i < 5 else "indoor",
            "is_same": True if i % 3 else False,
            "a.locator": f"https://test-iamge-{i}.png",
            "a.width": 500 + i,
            "a.height": 400 + i,
            "a.bboxes": [
                LabeledBoundingBox(label="car", top_left=[i, i], bottom_right=[i + 50, i + 50])
                for i in range(random.randint(2, 6))
            ],
            "a.label": ScoredClassificationLabel(label="dog", score=0.1 + i * 0.05),
            "b.locator": f"https://test-iamge-{i}.png",
            "b.width": 500 + i,
            "b.height": 400 + i,
            "b.bboxes": [
                LabeledBoundingBox(label="car", top_left=[i, i], bottom_right=[i + 50, i + 50])
                for i in range(random.randint(2, 6))
            ],
            "b.label": ScoredClassificationLabel(label="cat", score=0.1 + i * 0.05),
        }
        for i in range(10)
    ]
    df = pd.DataFrame(datapoints)
    df_expected = pd.DataFrame(
        dict(
            datapoint=[
                dict(
                    category=dp["category"],
                    is_same=dp["is_same"],
                    a={
                        DATA_TYPE_FIELD: DatapointType.IMAGE,
                        "label": dp["a.label"]._to_dict(),
                        "width": dp["a.width"],
                        "height": dp["a.height"],
                        "locator": dp["a.locator"],
                        "bboxes": [bbox._to_dict() for bbox in dp["a.bboxes"]],
                    },
                    b={
                        DATA_TYPE_FIELD: DatapointType.IMAGE,
                        "label": dp["b.label"]._to_dict(),
                        "width": dp["b.width"],
                        "height": dp["b.height"],
                        "locator": dp["b.locator"],
                        "bboxes": [bbox._to_dict() for bbox in dp["b.bboxes"]],
                    },
                    data_type=DatapointType.COMPOSITE,
                )
                for dp in datapoints
            ],
        ),
    )
    df_serialized = _to_serialized_dataframe(df, column=COL_DATAPOINT)

    assert df_serialized[COL_DATAPOINT].apply(json.loads).equals(df_expected[COL_DATAPOINT])

    df_expected = pd.DataFrame(
        [
            {
                "category": dp["category"],
                "is_same": dp["is_same"],
                "a.locator": dp["a.locator"],
                "a.width": dp["a.width"],
                "a.height": dp["a.height"],
                "a.bboxes": [
                    BoundingBox(label=bbox.label, top_left=bbox.top_left, bottom_right=bbox.bottom_right)
                    for bbox in dp["a.bboxes"]],
                "a.label": Label(label=dp["a.label"].label, score=dp["a.label"].score),
                "b.locator": dp["b.locator"],
                "b.width": dp["b.width"],
                "b.height": dp["b.height"],
                "b.bboxes": [
                    BoundingBox(label=bbox.label, top_left=bbox.top_left, bottom_right=bbox.bottom_right)
                    for bbox in dp["b.bboxes"]],
                "b.label": Label(label=dp["b.label"].label, score=dp["b.label"].score),
            }
            for dp in datapoints
        ]
    )
    df_deserialized = _to_deserialized_dataframe(df_serialized, column=COL_DATAPOINT)
    assert_frame_equal(df_deserialized, df_expected)
```

the final deserialization looks to have flattened `a.label` and `b.label`, which should remain `ClassificationLabel`.

kolena/_experimental/dataset/_dataset.py
```diff
@@ -83,6 +90,35 @@ def _infer_datatype_value(x: str) -> str:
...
+                raise InputValidationError(
+                    f"More than one delimeter '{SEP}' in prefix: '{prefix}'.",
+                )
+
+            composite_columns = df.filter(regex=rf"^{prefix}", axis=1).columns.to_list()
+            composite = df.loc[:, composite_columns].rename(columns=lambda col: col.split(SEP)[-1])
+            composite[DATA_TYPE_FIELD] = _infer_datatype(composite)
+
+            df[prefix] = composite.to_dict("records")
+            df.drop(columns=composite_columns, inplace=True)
```
ly-kolena, line 122: Can we avoid using `inplace=True`. My general reading is that there is no clear cut performance benefit and with our usage of passing dataframe as input, it's easy to introduce hard-to-debug bugs because of inplace manipulation. And there's no way to tell function caller that we do that to their input.

Edit: Looks like we're manipulating df in general in this function anyway, gonna think if there's alternative, otherwise the above comment doesn't matter much/

tests/unit/_experimental/dataset/test_dataset.py
consoull: Thanks! This test uncovered a few issues. The recursive flattening is fixed by adding `max_level=0` option to https://pandas.pydata.org/docs/reference/api/pandas.json_normalize.html.

Another issue this uncovered is LabeledBoundingBox being cast to BoundingBox during deserialization. It appears many of the [subtypes (eg. LabeledBoundingBox, ScoredBoundingBox) are missing annotation types](https://github.com/kolenaIO/kolena/blob/trunk/kolena/workflow/annotation.py#L106) and being cast to their parent class due to inheriting the annotation label.
```
(Pdb) df_deser['a.bboxes'][0]
[BoundingBox(top_left=(0.0, 0.0), bottom_right=(50.0, 50.0), width=50.0, height=50.0, area=2500.0, aspect_ratio=1.0), BoundingBox(top_left=(1.0, 1.0), bottom_right=(51.0, 51.0), width=50.0, height=50.0, area=2500.0, aspect_ratio=1.0), BoundingBox(top_left=(2.0, 2.0), bottom_right=(52.0, 52.0), width=50.0, height=50.0, area=2500.0, aspect_ratio=1.0)]
(Pdb) df_orig['a.bboxes'][0]
[LabeledBoundingBox(top_left=(0.0, 0.0), bottom_right=(50.0, 50.0), width=50.0, height=50.0, area=2500.0, aspect_ratio=1.0, label='car'), LabeledBoundingBox(top_left=(1.0, 1.0), bottom_right=(51.0, 51.0), width=50.0, height=50.0, area=2500.0, aspect_ratio=1.0, label='car'), LabeledBoundingBox(top_left=(2.0, 2.0), bottom_right=(52.0, 52.0), width=50.0, height=50.0, area=2500.0, aspect_ratio=1.0, label='car')]
```

kolena/_experimental/dataset/_dataset.py
ly-kolena: Although weary of the usage of `inplace=True` and prefer not to have function that manipulates input dataframe, this comment is nonblocking, as currently `_add_datatype` has single invocation point and it is modifying a dataframe internally created, not user's dataframe.

tests/unit/_experimental/dataset/test_dataset.py
ly-kolena: For the 2nd point, it is currently a known limitation for the tradeoff for ease of use, as we don't have enough information to deserialize to subtypes.

kolena/_experimental/dataset/_dataset.py
```diff
@@ -95,7 +131,7 @@ def _infer_datatype(df: pd.DataFrame) -> Union[pd.DataFrame, str]:
 def _to_serialized_dataframe(df: pd.DataFrame, column: str) -> pd.DataFrame:
     result = _dataframe_object_serde(df, _serialize_dataobject)
     if column == COL_DATAPOINT:
-        result[DATA_TYPE_FIELD] = _infer_datatype(df)
+        _add_datatype(result)
```
ly-kolena, line 139: Current logic has no issue, but I'm wondering if we should add a test to ensure `_to_serialized_dataframe` doesn't apply composite heuristic to, say, serializing result dataframe.

kolena/_experimental/dataset/_dataset.py
consoull: `_add_datatype` is going to have a few more invocations on the backend so I think it is worthwhile to have the discussion now.  We can update the function to return a new copy of the DataFrame leaving the previous one intact, but if all functions along the dataset deser call stack ensure the input dataframe is not manipulated there will certainly be performance implications for large datasets.

In my opinion, the public api `register_dataset` should create a copy immediately and called private functions are free to mutate. This ensures that the users DataFrame is not mutated and keeps copies to a minimum.

ly-kolena: Even for private functions, they could be reused/shared by other public/private methods, now or future. A function that modifies input data is easy to cause bugs when call paths are non-trivial. Current scoping to `_add_datatype` (as small as possible) looks ok to me, still future developers should be aware of this behavior if they want to use it. I don't think we should expand it to `register_dataset`.

A slightly different discussion of use of `inplace=True`, I don't want to blindly believe in performance gain from `inplace`, but also don't want to unnecessarily fear of it. For dicussion's sake, references that in general not in favor of `inplace`
1. https://stackoverflow.com/questions/45570984/in-pandas-is-inplace-true-considered-harmful-or-not
2. https://sourcery.ai/blog/pandas-inplace/

tests/unit/_experimental/dataset/test_dataset.py
```diff
@@ -52,6 +57,89 @@ def test__infer_datatype_value(uri: str, expected: str) -> None:
...
+    composite_dataset = pd.DataFrame(
+        {
+            "a": [i for i in range(5)],
+            "a.locator": [f"{i}.png" for i in range(5)],
+        },
+    )
+    with pytest.raises(InputValidationError):
+        _add_datatype(composite_dataset)
+
+    composite_dataset = pd.DataFrame(
```
ly-kolena, line 134: Really really nit: a test case like this has unexpected output

```
    composite_dataset = pd.DataFrame(
        {
            ".dots": [i for i in range(5)],
            "x": [i for i in range(5)],
        },
    )
    _add_datatype(composite_dataset)
```

ly-kolena (approved):
With fix of max_level to `flatten_composite`, logic looks good.

tests/unit/_experimental/dataset/test_dataset.py
consoull: I think this is a good callout! I will add a validation check for empty prefix

PR #356, opened 2023-12-13
 client require id_field when register dataset , allow chunk processing

grant-Kolena:
This is currently blocked by https://github.com/kolenaIO/shipyard/pull/1737

### DIFF
--- a/kolena/_experimental/dataset/_dataset.py
+++ b/kolena/_experimental/dataset/_dataset.py
@@ -25,6 +25,7 @@ from kolena._api.v2.dataset import LoadDatapointsRequest
 from kolena._api.v2.dataset import Path
 from kolena._api.v2.dataset import RegisterRequest
 from kolena._experimental.dataset.common import COL_DATAPOINT
+from kolena._experimental.dataset.common import COL_DATAPOINT_ID_OBJECT
 from kolena._experimental.dataset.common import validate_batch_size
 from kolena._experimental.dataset.common import validate_id_fields
 from kolena._utils import krequests_v2 as krequests
@@ -124,7 +125,10 @@ def register_dataset(name: str, df: pd.DataFrame, id_fields: List[str]) -> None:
 
     load_uuid = init_upload().uuid
 
-    df_serialized = _to_serialized_dataframe(df, column=COL_DATAPOINT)
+    df_serialized_datapoint = _to_serialized_dataframe(df, column=COL_DATAPOINT)
+    df_serialized_datapoint_id_object = _to_serialized_dataframe(df[sorted(id_fields)], column=COL_DATAPOINT_ID_OBJECT)
+    df_serialized = pd.concat([df_serialized_datapoint, df_serialized_datapoint_id_object], axis=1)
+
     upload_data_frame(df=df_serialized, batch_size=BatchSize.UPLOAD_RECORDS.value, load_uuid=load_uuid)
     request = RegisterRequest(name=name, id_fields=id_fields, uuid=load_uuid)
     response = krequests.post(Path.REGISTER, json=asdict(request))
--- a/kolena/_experimental/dataset/_dataset.py
+++ b/kolena/_experimental/dataset/_dataset.py
@@ -16,6 +16,7 @@ import mimetypes
 from dataclasses import asdict
 from enum import Enum
 from typing import Iterator
+from typing import List
 from typing import Union
 
 import pandas as pd
@@ -25,6 +26,7 @@ from kolena._api.v2.dataset import Path
 from kolena._api.v2.dataset import RegisterRequest
 from kolena._experimental.dataset.common import COL_DATAPOINT
 from kolena._experimental.dataset.common import validate_batch_size
+from kolena._experimental.dataset.common import validate_id_fields
 from kolena._utils import krequests_v2 as krequests
 from kolena._utils.batched_load import _BatchedLoader
 from kolena._utils.batched_load import init_upload
@@ -113,15 +115,17 @@ def _to_deserialized_dataframe(df: pd.DataFrame, column: str) -> pd.DataFrame:
     return result
 
 
-def register_dataset(name: str, df: pd.DataFrame) -> None:
+def register_dataset(name: str, df: pd.DataFrame, id_fields: List[str]) -> None:
     """
-    Create or update a dataset with datapoints.
+    Create or update a dataset with datapoints and id_fields.
     """
+    validate_id_fields(df, id_fields)
+
     load_uuid = init_upload().uuid
 
     df_serialized = _to_serialized_dataframe(df, column=COL_DATAPOINT)
     upload_data_frame(df=df_serialized, batch_size=BatchSize.UPLOAD_RECORDS.value, load_uuid=load_uuid)
-    request = RegisterRequest(name=name, uuid=load_uuid)
+    request = RegisterRequest(name=name, id_fields=id_fields, uuid=load_uuid)
     response = krequests.post(Path.REGISTER, json=asdict(request))
     krequests.raise_for_status(response)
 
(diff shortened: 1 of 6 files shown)


kolena/_experimental/dataset/_dataset.py
```diff
@@ -113,15 +121,52 @@ def _to_deserialized_dataframe(df: pd.DataFrame, column: str) -> pd.DataFrame:
...
     """
-    Create or update a dataset with datapoints.
+    Create or update a dataset with datapoints and id_fields. If the dataset already exists, in order to associate the
+    existing result with the new datapoints, the id_fields need be the same as the existing dataset.
+
+    :param name: name of the dataset
+    :param df: an iterator of pandas dataframe or a pandas dataframe, you can pass in the iterator if you want to have
+                batch processing,
+                 example iterator usage: csv_reader = pd.read_csv("PathToDataset.csv", chunksize=10)
+    :param id_fields: a list of id fields, this will be used to link the result with the datapoints
```
gordonhart: @ly-kolena thoughts on inferring these? For the simple case, I'd like for users to not have to think about ID.

ly-kolena: The discussion was

1. On top of existing datapoint_type infer, we can also auto-configure for dataframe with either `locator` or `text` column to use them. respectively, and raise exception when neither is found and user does not provide `id_fields`.
2. Warn users if the final resolved `id_fields` differ from existing one of the dataset (if this is an update)
3. (Optional future feature) augment handling of 2) to prompt for user confirmation, with option like `force=True` to bypass prompt, in the future for better visibility on `id_fields` change.

One downside is that, `id_fields` is still critical for "consistent" data handling, e.g. datapoint and their results. It would still come as a surprise if users are not aware of it in the first place and then have to come back to comb through documentation to realize this is the cause of data ingestion error, or something else.

Does that sound reasonable?

kolena/_experimental/dataset/common.py
```diff
@@ -11,14 +11,41 @@
...
 
 
-COL_DATAPOINT = "datapoint"
-COL_EVAL_CONFIG = "eval_config"
-COL_RESULT = "result"
+def validate_id_fields(df: pd.DataFrame, id_fields: List[str], existing_id_fields: List[str] = None) -> None:
+    if len(id_fields) == 0:
+        raise InputValidationError(f"invalid id_fields '{id_fields}': expected at least one field")
+    if existing_id_fields:
+        if Counter(id_fields) != Counter(existing_id_fields):
```
yifanwu-kolena: should we use `set` at here?

grant-Kolena: set does not validate on the dups, but we won't have dups?

yifanwu-kolena: oh, i see. maybe we need to do an explicit no dups check for `id_fields`? b/c the current warning msg is only for the id_fields changes, and it will not fail when there are dups in the `id_fields`

grant-Kolena: got it I will add explicit no dup check

kolena/_experimental/dataset/_dataset.py
```diff
@@ -113,15 +121,52 @@ def _to_deserialized_dataframe(df: pd.DataFrame, column: str) -> pd.DataFrame:
     return result
 
 
-def register_dataset(name: str, df: pd.DataFrame) -> None:
+def _upload_dataset_chunk(df: pd.DataFrame, load_uuid: str, id_fields: List[str]) -> None:
```
ly-kolena, line 133: We're doing chunking inside `upload_data_frame`. Is there reason for 2-level of chunk control?

kolena/_experimental/dataset/_dataset.py
```diff
@@ -113,15 +121,52 @@ def _to_deserialized_dataframe(df: pd.DataFrame, column: str) -> pd.DataFrame:
...
+    :return None
     """
     load_uuid = init_upload().uuid
-
-    df_serialized = _to_serialized_dataframe(df, column=COL_DATAPOINT)
-    upload_data_frame(df=df_serialized, batch_size=BatchSize.UPLOAD_RECORDS.value, load_uuid=load_uuid)
-    request = RegisterRequest(name=name, uuid=load_uuid)
+    existing_dataset = load_dataset(name)
+    existing_id_fields = []
+    if existing_dataset:
```
ly-kolena, line 171: Any reason to continue if dataset does not exist?

grant-Kolena: This function allows both modify and creation, `register_dataset` is a misleading name, but changing it would also be breaking. Can we change to `create_or_update_dataset`?

ly-kolena: temporary insanity, mixed up with `test`. Sorry for the confusion. No issue here.

kolena/_experimental/dataset/_dataset.py
grant-Kolena: Chunking here allows the user to send their dataframe to upload in chunks, and the bach_load chunking makes sure we are not upload a giant file at once.
There might not be a need for it in dataset as the chunks size on the dataset should always be smaller than batch_load chunk size.  However, for other things that does not have chunk support in the respective upload function, we still need the batch_load chunking

kolena/_experimental/dataset/common.py
```diff
@@ -11,14 +11,41 @@
...
+                f"ID field for the existing dataset has been changed from {existing_id_fields} to {id_fields},"
+                f" this will disassociate the existing result from the new datapoints",
+            )
+    for id_field in id_fields:
+        if id_field not in df.columns:
+            raise InputValidationError(
+                f"invalid id_fields '{id_fields}': field '{id_field}' does not exist in dataframe",
+            )
+    # check uniqueness of the id columns
+    unique_id_fields = df[id_fields].drop_duplicates()
```
ly-kolena, line 53: Is `df.duplicated(subset=id_fields).any()` "better"?

Can use a unit test for this case if there isn't one already

grant-Kolena: Sure I will add a unit test for this

kolena/_experimental/dataset/_dataset.py
ly-kolena: @grant-Kolena , have you addressed this concern?

kolena/_experimental/dataset/_dataset.py
```diff
@@ -113,15 +121,52 @@ def _to_deserialized_dataframe(df: pd.DataFrame, column: str) -> pd.DataFrame:
...
+    response = krequests.put(Path.LOAD_DATASET, json=asdict(LoadDatasetByNameRequest(name=name)))
+    if response.status_code == requests.codes.not_found:
+        return None
+
+    response.raise_for_status()
+
+    return from_dict(EntityData, response.json())
+
+
+def register_dataset(name: str, df: Union[Iterator[pd.DataFrame], pd.DataFrame], id_fields: List[str]) -> None:
```
ly-kolena: Non blocking: Only highlighting one limitation is that id uniqueness check would not be complete if users pass an iterator.

grant-Kolena: We do have dedup on the temp table on server side, the behavior would be different. Would this be acceptable?

ly-kolena: I'm good with it.

kolena/_experimental/dataset/_dataset.py
grant-Kolena: We think the infer should happen on galleon side, if someone is integrating with SDK they should provide the id explicitly as there is no interface to let them select /inform them what has been selected as id field.

Also I am not clear on the rules of what should be inferred as id

grant-Kolena: discussed offline, just inferring locator and text for now

ly-kolena (approved):

grant-Kolena:
Merging and start the work on galleon

PR #359, opened 2023-12-14
Fix dataset integration tests

ly-kolena:

### DIFF
--- a/kolena/_experimental/dataset/_dataset.py
+++ b/kolena/_experimental/dataset/_dataset.py
@@ -31,7 +31,7 @@ from kolena._api.v2.dataset import RegisterRequest
 from kolena._experimental.dataset.common import COL_DATAPOINT
 from kolena._experimental.dataset.common import COL_DATAPOINT_ID_OBJECT
 from kolena._experimental.dataset.common import validate_batch_size
-from kolena._experimental.dataset.common import validate_id_fields
+from kolena._experimental.dataset.common import validate_dataframe_ids
 from kolena._utils import krequests_v2 as krequests
 from kolena._utils.batched_load import _BatchedLoader
 from kolena._utils.batched_load import init_upload
@@ -45,7 +45,6 @@ from kolena.workflow._datatypes import _serialize_dataobject
 from kolena.workflow._datatypes import DATA_TYPE_FIELD
 from kolena.workflow._datatypes import TypedDataObject
 from kolena.workflow.io import _dataframe_object_serde
-
 
 FIELD_LOCATOR = "locator"
 FIELD_TEXT = "text"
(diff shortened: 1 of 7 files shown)


tests/integration/_experimental/dataset/test_dataset.py
```diff
@@ -148,17 +149,18 @@ def test__register_dataset__composite() -> None:
         for i in range(1, 20)
     ]
     columns = datapoints[0].keys()
+    id_fields = ["id"]
```
consoull: I haven't done thorough testing but id_fields appear to work for composite style keys. I'd recommend we use "a.text" and "b.text" here instead of top level id field

grant-Kolena (approved):

tests/integration/_experimental/dataset/test_dataset.py
```diff
@@ -148,17 +149,18 @@ def test__register_dataset__composite() -> None:
...
+    register_dataset(name, df, id_fields=id_fields)
 
     loaded_datapoints = fetch_dataset(name)
     loaded_datapoints = loaded_datapoints.sort_values("total_word_count", ignore_index=True).reindex(columns=columns)
     assert_frame_equal(df, loaded_datapoints)
 
     # update dataset
     df = pd.DataFrame(datapoints[:5] + datapoints[7:15], columns=columns)
-    register_dataset(name, df)
+    register_dataset(name, df, id_fields=id_fields)
```
nankolena, line 162: Not sure if we should do it in this PR, but I think we should definitely add some tests to cover [this logic](https://github.com/kolenaIO/kolena/blob/ab94b07b0951d104d78fa70ff60c64c8e518e39f/kolena/_experimental/dataset/_dataset.py#L221-L227)

tests/integration/_experimental/dataset/test_dataset.py
ly-kolena: i think there could be discussion on that, and will leave that to your PR if we do want infer id even for composite. Technically the test setup here is correct.

nankolena (approved):

tests/integration/_experimental/dataset/test_dataset.py
consoull: Main reason I pointed this out is we currently have no test coverage for composite style id_fields

kolena/_experimental/dataset/_dataset.py
```diff
@@ -217,23 +232,17 @@ def register_dataset(
...
     if isinstance(df, pd.DataFrame):
-        validate_id_fields(df, id_fields, existing_id_fields)
+        id_fields = resolve_id_fields(df, id_fields, existing_dataset)
+        validate_dataframe_ids(df, id_fields)
         _upload_dataset_chunk(df, load_uuid, id_fields)
     else:
         validated = False
         for chunk in df:
             if not validated:
+                id_fields = resolve_id_fields(chunk, id_fields, existing_dataset)
```
ly-kolena, line 243: cc @grant-Kolena for reviewing this change. I noticed original logic could break when iterator is passed

tests/integration/_experimental/dataset/test_dataset.py
ly-kolena: added unit tests

tests/integration/_experimental/dataset/test_dataset.py
ly-kolena: updated as suggested.

munkyshi (approved):

PR #366, opened 2023-12-18
List dataset commit history and get snapshots

nankolena:
### Linked issue(s):
https://linear.app/kolena/issue/KOL-3548/sdk-load-previous-version-dataset

- Add client-side functionalities to retrieve the commit history of a dataset
- Update the `fetch_dataset` method to allow passing in a commit to get historical snapshots

Note that CI will fail until https://github.com/kolenaIO/shipyard/pull/1742 reaches prod.

### DIFF
--- a/kolena/_experimental/dataset/_dataset.py
+++ b/kolena/_experimental/dataset/_dataset.py
@@ -313,26 +313,36 @@ def _list_commits(name: str, desc: bool = False, offset: int = 0, limit: int = 5
     return from_dict(ListCommitHistoryResponse, response.json())
 
 
-def _iter_commits(name: str, desc: bool = False, limit: int = None) -> Iterator[Tuple[int, List[CommitData]]]:
+def _iter_commits(
+    name: str,
+    desc: bool = False,
+    limit: int = None,
+    page_size: int = 50,
+) -> Iterator[Tuple[int, List[CommitData]]]:
     """
     Get an iterator over the commit history of the dataset.
     """
-    initial_response = _list_commits(name, desc=desc)
+    initial_response = _list_commits(name, desc=desc, limit=page_size)
     total_commit_count = initial_response.total_count
     if not limit:
         limit = total_commit_count
     yield total_commit_count, initial_response.records
     current_count = len(initial_response.records)
     while current_count < min(limit, total_commit_count):
-        yield total_commit_count, _list_commits(name, desc=desc, offset=current_count).records
+        yield total_commit_count, _list_commits(name, desc=desc, offset=current_count, limit=page_size).records
         current_count += len(initial_response.records)
 
 
-def fetch_commits(name: str, desc: bool = False, limit: int = None) -> Tuple[int, List[CommitData]]:
+def fetch_commits(
+    name: str,
+    desc: bool = False,
+    limit: int = None,
+    page_size: int = 50,
+) -> Tuple[int, List[CommitData]]:
     """
     Get the commit history of a dataset.
     """
-    iter_commit_responses = list(_iter_commits(name, desc, limit))
+    iter_commit_responses = list(_iter_commits(name, desc, limit, page_size))
     total_commit_count = iter_commit_responses[0][0]
     commits = [commit for response in iter_commit_responses for commit in response[1]][:limit]
     return total_commit_count, commits
--- a/kolena/_experimental/dataset/_dataset.py
+++ b/kolena/_experimental/dataset/_dataset.py
@@ -262,12 +262,15 @@ def register_dataset(
 
 def _iter_dataset_raw(
     name: str,
+    commit: str = None,
     batch_size: int = BatchSize.LOAD_SAMPLES.value,
 ) -> Iterator[pd.DataFrame]:
     validate_batch_size(batch_size)
     init_request = LoadDatapointsRequest(
         name=name,
+        commit=commit,
         batch_size=batch_size,
+
     )
     yield from _BatchedLoader.iter_data(
         init_request=init_request,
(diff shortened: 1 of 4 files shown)


kolena/_experimental/dataset/_dataset.py
```diff
@@ -273,23 +279,70 @@ def _iter_dataset_raw(
...
+    if not limit:
+        limit = total_commit_count
+    yield total_commit_count, initial_response.records
+    current_count = len(initial_response.records)
+    while current_count < min(limit, total_commit_count):
+        yield total_commit_count, _list_commits(name, desc=desc, offset=current_count, limit=page_size).records
+        current_count += len(initial_response.records)
+
+
+def fetch_commits(
```
ly-kolena: Nit: do we want a name that makes it more obvious this is for dataset? Maybe `fetch_dataset_history` or something? Debating if we should include `commit` in method name as it is a "new" concept that may need explanation.

kolena/_experimental/dataset/_dataset.py
```diff
@@ -273,23 +279,70 @@ def _iter_dataset_raw(
...
+    name: str,
+    desc: bool = False,
+    limit: int = None,
+    page_size: int = 50,
+) -> Tuple[int, List[CommitData]]:
+    """
+    Get the commit history of a dataset.
+    """
+    iter_commit_responses = list(_iter_commits(name, desc, limit, page_size))
```
grant-Kolena: what would be the resp if name is invalid?

tests/integration/_experimental/dataset/test_dataset.py
```diff
@@ -169,3 +177,92 @@ def test__fetch_dataset__not_exist() -> None:
...
+    assert total_commits == TEST_COMMIT_HISTORY_VERSIONS
+    previous_commit_time = -1
+    for version, commit in enumerate(commits):
+        assert commit.timestamp >= previous_commit_time
+        assert commit.n_removed == version
+        assert commit.n_added == version + 1
+        previous_commit_time = commit.timestamp
+
+
+@pytest.mark.depends(on=["test__fetch_commits"])
```
ly-kolena: Do we this depends? Looks like `test__fetch_commits` is not doing modification to data `with_dataset_commits` and these tests can just run without the dependency.

kolena/_experimental/dataset/_dataset.py
```diff
@@ -273,23 +279,70 @@ def _iter_dataset_raw(
...
+    yield total_commit_count, initial_response.records
+    current_count = len(initial_response.records)
+    while current_count < min(limit, total_commit_count):
+        yield total_commit_count, _list_commits(name, desc=desc, offset=current_count, limit=page_size).records
+        current_count += len(initial_response.records)
+
+
+def fetch_commits(
+    name: str,
+    desc: bool = False,
```
ly-kolena: Got confused for a few seconds, thinking "Why is description a boolean". Maybe a docstring or param name update.

kolena/_experimental/dataset/_dataset.py
```diff
@@ -273,23 +279,70 @@ def _iter_dataset_raw(
...
+    """
+    initial_response = _list_commits(name, desc=desc, limit=page_size)
+    total_commit_count = initial_response.total_count
+    if not limit:
+        limit = total_commit_count
+    yield total_commit_count, initial_response.records
+    current_count = len(initial_response.records)
+    while current_count < min(limit, total_commit_count):
+        yield total_commit_count, _list_commits(name, desc=desc, offset=current_count, limit=page_size).records
+        current_count += len(initial_response.records)
```
ly-kolena: why are we using `initial_response.records` here not the last request length? Also, should the while loop also use the `total` from latest response?

kolena/_experimental/dataset/_dataset.py
nankolena: A 404 not found error would be thrown if the dataset does not exist

kolena/_experimental/dataset/_dataset.py
nankolena: Good idea, will rename to `fetch_dataset_history`

tests/integration/_experimental/dataset/test_dataset.py
nankolena: Right we don't need this dependency marked explicitly. My mental model was to skip other tests if `test__fetch_commits` already fails, but there is no logical dependency between them

kolena/_experimental/dataset/_dataset.py
nankolena: Good catch, should increment with the new response.

As for using the `total_count` from the initial response, that's because in case of edits that happen in parallel to the `get_dataset_history()` call, the `total_count` may increase between pages and this while loop might end up being an infinite one.

ly-kolena: With offset and limit given every request, and that server respond includes latest total_count, why would this be infinite loop?

nankolena: Thought through a bit more, I think you are right and we should not expect infinite loop given that the commit history can only increase over time.
- With `descending=False`: there is no problem, as the `offset` will serve as the cursor on chronological order
- With `descending=True`: there might be situations where the same commit is retrieved multiple times if new commits are added

ly-kolena: For 2nd case, supposedly we can dedup on client side, with caveat we'll for sure miss the new commits, but I think in a way acceptable.

kolena/_experimental/dataset/_dataset.py
nankolena: Raised quick PR on server side for the same: https://github.com/kolenaIO/shipyard/pull/1754

Will update client side in next commit

ly-kolena (approved):

PR #368, opened 2023-12-18
fix: robust uniqueness checking

yifanwu-kolena:
### Linked issue(s):

https://linear.app/kolena/issue/KOL-4137/sdk-client-enforce-id-columns-to-only-be-scalar-of-string-and-integer

### DIFF
--- a/kolena/_experimental/dataset/common.py
+++ b/kolena/_experimental/dataset/common.py
@@ -45,6 +45,11 @@ def validate_id_fields(id_fields: List[str], existing_id_fields: List[str] = Non
             )
 
 
+def _convert_non_int_str_to_str(df: pd.DataFrame) -> pd.DataFrame:
+    df = df.applymap(lambda x: str(x) if (not isinstance(x, (int, str))) else x)
+    return df
+
+
 def validate_dataframe_ids(df: pd.DataFrame, id_fields: List[str]) -> None:
     for id_field in id_fields:
         if id_field not in df.columns:
@@ -52,7 +57,7 @@ def validate_dataframe_ids(df: pd.DataFrame, id_fields: List[str]) -> None:
                 f"invalid id_fields: field '{id_field}' does not exist in dataframe",
             )
     # check uniqueness of the id columns
-    unique_id_fields = df[id_fields].drop_duplicates()
+    unique_id_fields = _convert_non_int_str_to_str(df[id_fields]).drop_duplicates()
     if len(unique_id_fields) != len(df):
         raise InputValidationError(
             f"invalid id_fields: " f"input dataframe's id field values are not unique for {id_fields}",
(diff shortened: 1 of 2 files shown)


kolena/_experimental/dataset/common.py
```diff
@@ -45,14 +45,19 @@ def validate_id_fields(id_fields: List[str], existing_id_fields: List[str] = Non
             )
 
 
+def _convert_non_int_str_to_str(df: pd.DataFrame) -> pd.DataFrame:
+    df = df.applymap(lambda x: str(x) if (not isinstance(x, (int, str))) else x)
```
ly-kolena: I think this would have unexpected result when `str(x)` is not applied on trivial objects, for example class instances. If we choose to restrict id column type, I'd prefer not to auto convert but be explicit.

Thinking about this again, I'm retracting my initial suggestion and think we probably don't need this enforcement on int or str type. I think the only "problematic" part is floating number comparison, and we have our existing precision truncation on server side for that. What do you think?

yifanwu-kolena: - ingestion via web UI
  - only basic types via json parsing
- ingestion via python script
  - it might fail for non-trivial class for `str(x)`, but users can convert it into a dict

i think this change will make the system more robust. currently it will just throw some pandas error when selecting dict and other type as id fields as unit tests.

i think we want to have this uniqueness check on the client side no matter what dedup we have on server side

ly-kolena: I want to highlight the problem of using `str` as auto type coercion. Using the new unit test, this example would pass, but it should fail. (there could be others, as `str` is not designed for id)

```
        (pd.DataFrame(dict(a=[{"a": 1, "b": 2}, {"a": 2, "b": 1}, {"b": 2, "a": 1}], b=[1, 2, 1])), ["a", "b"]),
```

yifanwu-kolena: synced offline, string coercion is not the right approach to compare uniqueness, [converted the multiple columns df into a single column](https://github.com/kolenaIO/kolena/pull/368/commits/c1e855c3212633a888649edb12c6d0e8902a8efc) and leverage the default dictionary comparison might do the trick

grant-Kolena (approved):

kolena/_experimental/dataset/common.py
```diff
@@ -45,9 +45,11 @@ def validate_id_fields(id_fields: List[str], existing_id_fields: List[str] = Non
             )
 
 
-def _convert_non_int_str_to_str(df: pd.DataFrame) -> pd.DataFrame:
-    df = df.applymap(lambda x: str(x) if (not isinstance(x, (int, str))) else x)
-    return df
+def _convert_to_single_column(df):
+    # create a new dataframe with dictionary in each row
+    new_df = pd.DataFrame(df.apply(lambda x: x.to_dict(), axis=1))
```
ly-kolena: Nit: Series also support drop_duplicates. For our current usage, looks like we can just use `df[id_fields].apply(...).drop_duplcates()`

Alternatively, not sure if it saves a bit of memory or else
```
if <a_serie>.duplicated().any():
  .....
```

ly-kolena (approved):

PR #375, opened 2023-12-22
Example of thresholded metrics

diegokolena:


munkyshi:
@diegokolena is this meant to be an example that we share publicly, or just an example for devs to use?

kolena/_experimental/thresholded_metrics/__init__.py
munkyshi, line 1: Do we need new code in experimental for our customers to use `thresholded_metrics`? Is there a reason this lives here instead of in `examples/thresholded_metrics/thresholded_metrics`?

diegokolena: This example is not a real example, mainly to test the feature itself; that is why we put on experimental.

gordonhart:
Should this be closed?

PR #387, opened 2024-01-03
chore: export asset and annotation from kolena.dataset

yifanwu-kolena:
### Linked issue(s):

https://linear.app/kolena/issue/KOL-4394/re-export-common-classes-from-kolenadatasets

new docs

Screen Shot 2024-01-03 at 10 24 46 AM


tests/unit/dataset/test_annotation.py
```diff
@@ -0,0 +1,18 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from kolena.dataset.annotation import Annotation
+
+
+def test__import():
```
ly-kolena, line 16: nit^n: can have a single test file with

```

def test__import_annoation():
  from kolena.dataset.annoation import XXX

def test__import_asset():
   ....
```

kolena/dataset/annotation.py
```diff
@@ -0,0 +1,64 @@
...
+from kolena.workflow.annotation import ScoredLabeledBoundingBox
+from kolena.workflow.annotation import ScoredLabeledBoundingBox3D
+from kolena.workflow.annotation import ScoredLabeledPolygon
+from kolena.workflow.annotation import ScoredLabeledTimeSegment
+from kolena.workflow.annotation import ScoredPolygon
+from kolena.workflow.annotation import ScoredTimeSegment
+from kolena.workflow.annotation import SegmentationMask
+from kolena.workflow.annotation import TimeSegment
+
+__all__ = [
```
gordonhart, line 39: Is it possible for us to achieve this re-export without having to manually specify each of the exported items? Will be easy to forget to update this.

yifanwu-kolena: all of these are mainly for the mkdocs rather than python module behaviours. i tried `from kolena.workflow.annotation import *` without `__all__`, but docs will not be generated

kolena/dataset/asset.py
```diff
@@ -0,0 +1,21 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from kolena.workflow.asset import BaseVideoAsset
+from kolena.workflow.asset import BinaryAsset
+from kolena.workflow.asset import ImageAsset
+from kolena.workflow.asset import PlainTextAsset
+from kolena.workflow.asset import PointCloudAsset
+from kolena.workflow.asset import VideoAsset
+
+__all__ = ["ImageAsset", "PlainTextAsset", "BinaryAsset", "PointCloudAsset", "BaseVideoAsset", "VideoAsset"]
```
ly-kolena, line 21: Are we intentionally skipping `Annotation` and `Asset`?

yifanwu-kolena: i just use `_ASSET_TYPES` and `_ANNOTATION_TYPES`, i don't think we expect users will need to inherit the base class

yifanwu-kolena:
prefer https://github.com/kolenaIO/kolena/pull/395 over re-export under dataset

PR #389, opened 2024-01-03
Enforce `id_fields` to be hashable

consoull:
### Linked issue(s):
Towards KOL-4385

During stress testing, the id_field uniqueness check was found to be a significant bottleneck. The cause was converting all id_fields to a dict to compare by value.

This commit addresses the issue by enforcing id_fields to be hashable so the `pandas.DataFrame.duplicated` can carry out the check using its underlying hashtable. This speeds up id_field validation checks.


tests/unit/_experimental/dataset/test_common.py
```diff
@@ -72,9 +70,12 @@ def test__validate_dataframe_ids(df: pd.DataFrame, id_fields: List[str]) -> None
...
         # dataframe is missing one of id_fields
         (pd.DataFrame(dict(a=[1, 2, 3])), ["a", "b"]),
+        # dataframe id_fields datatypes are not primitives
+        (pd.DataFrame(dict(a=[dict(c=i) for i in range(3)], b=[1, 2, 1])), ["a", "b"]),
+        (pd.DataFrame(dict(a=[[1], [2], [3]], b=[1, 2, 1])), ["a", "b"]),
         # dataframe values in id_fields is not unique
         (pd.DataFrame(dict(a=[1, 2, 1], b=[1, 2, 1])), ["a", "b"]),
-        (pd.DataFrame(dict(a=[[1], [1], [1]], b=[1, 2, 1])), ["a", "b"]),
+        (pd.DataFrame(dict(a=[1, 1, 1], b=[1, 2, 1])), ["a", "b"]),
         # the key sequence difference will not make it unique
```
yifanwu-kolena: i think this comment is obsolete, maybe we should remove all tests under this comment as well?

kolena/_experimental/dataset/common.py
```diff
@@ -58,4 +58,6 @@ def validate_dataframe_ids(df: pd.DataFrame, id_fields: List[str]) -> None:
             raise InputValidationError(
                 f"invalid id_fields: field '{id_field}' does not exist in dataframe",
             )
+        elif df[id_field].dtype == object and not isinstance(df.iloc[0][id_field], str):
```
ly-kolena: Just double checking: types like pandas categorical or timestamp would also fall here, i.e. `object`?

ly-kolena (approved):

gordonhart:
Is this restriction 100% necessary for perf, or can we avoid the expensive conversion when ID fields _are_ primitives? Ideally we'd have as few restrictions here as possible — there are valid reasons to use e.g. `locator, bbox` as a primary key, if you wanted to process different regions of the same image as different datapoints.

kolena/_experimental/dataset/common.py
yifanwu-kolena: datetime will be it's own dtype
ref: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.dtypes.html

consoull: Confirming these won't fall under object
```
>>> pd.Series(["a", "b", "c"], dtype="category").dtype == object
False
>>> pd.Series(pd.date_range("2000", freq="D", periods=3)).dtype == object
False
```

consoull:
@gordonhart I couldn't think of a valid use case for an `Object` to be an id_field. Thanks for providing one! I'm going to update the PR to only run the dictionary conversion on non-string objects
> Is this restriction 100% necessary for perf, or can we avoid the expensive conversion when ID fields are primitives?

PR #404, opened 2024-01-04
added __repr__ for annotation classes

grant-Kolena:
### Linked issue(s):
fixes KOL-4420 this is a continuation of this [draft](https://github.com/kolenaIO/kolena/pull/393/files)


kolena/annotation.py
```diff
@@ -42,15 +43,18 @@ def _data_category() -> str:
         return "ANNOTATION"
 
 
-@dataclass(frozen=True, config=ValidatorConfig)
+@dataclass(frozen=True, config=ValidatorConfig, repr=False)
```
grant-Kolena, line 46: dataclass creates a default one otherwise

kolena/annotation.py
```diff
@@ -42,15 +43,18 @@ def _data_category() -> str:
...
 class Annotation(TypedDataObject[_AnnotationType], metaclass=ABCMeta):
     """The base class for all annotation types."""
 
     def __init_subclass__(cls, **kwargs):
         _register_data_type(cls)
 
+    def __repr__(self):
```
consoull, line 53: I suggest using `__str__` instead. That is what pandas is using to dump the object into a csv cell.
It convention that `__repr__` outputs valid python code to construct the object

consoull: https://stackoverflow.com/questions/1436703/what-is-the-difference-between-str-and-repr

consoull: Nevermind. Containers will use the contained objects __repr__ so we need to use __repr__ here

grant-Kolena: Both repr and str needs to be changed after experimentation

PR #410, opened 2024-01-05
recursively serialize dataobject

grant-Kolena:

### DIFF
--- a/kolena/io.py
+++ b/kolena/io.py
@@ -24,8 +24,10 @@ from kolena._utils.datatypes import DataObject
 
 
 def _serialize_dataobject(x: Any) -> Any:
-    if isinstance(x, list):
-        return [item._to_dict() if isinstance(item, DataObject) else item for item in x]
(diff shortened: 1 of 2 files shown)


yifanwu-kolena (approved):

kolena/io.py
```diff
@@ -24,8 +24,10 @@
 
 
 def _serialize_dataobject(x: Any) -> Any:
-    if isinstance(x, list):
-        return [item._to_dict() if isinstance(item, DataObject) else item for item in x]
+    if isinstance(x, (list, tuple)):
```
yifanwu-kolena, line 27: seems like Union is only supported [after 3.10](https://docs.python.org/3/library/functions.html#isinstance)

grant-Kolena: oh, I will use `or` instead, thanks for pointing this out

ly-kolena:
For reference https://github.com/kolenaIO/kolena/pull/274#issuecomment-1761653852

PR #416, opened 2024-01-05
Legacy docs

diegokolena:

### DIFF
--- a/.github/workflows/legacy_docs.yml
+++ b/.github/workflows/legacy_docs.yml
@@ -4,8 +4,6 @@ on:
   push:
     branches:
       - hotfix/0.100.x
-      - diego/legacy-docs
-
 jobs:
   docs:
     name: Build and publish documentation
--- a/.github/workflows/legacy_docs.yml
+++ b/.github/workflows/legacy_docs.yml
@@ -1,0 +1,52 @@
+name: Docs
+
+on:
+  push:
+    branches:
+      - legacy-docs
+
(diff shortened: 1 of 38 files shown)


munkyshi:
Updated `hotfix/0.100.x` to be cut from the `0.100.0` tag, rather than an arbitrary point in time from `trunk`. I've gone ahead and cherry picked the changes back to this older point in time, though it might not have updated smoothly, due to code restructuring over the past few weeks.

.github/workflows/legacy_docs.yml
```diff
@@ -0,0 +1,53 @@
+name: Docs
+
+on:
+  push:
+    branches:
+      - hotfix/0.100.x
+      - diego/legacy-docs
```
munkyshi: Should probably remove this before merging.

mkdocs.yml
```diff
@@ -1,11 +1,11 @@
-site_name: Testing with Kolena
+site_name: Testing with Kolena (legacy)
```
munkyshi: Might want this to say `0.100.0` instead of `legacy`

munkyshi (approved):

mkdocs.yml
diegokolena: Updated

.github/workflows/legacy_docs.yml
diegokolena: Updated

PR #419, opened 2024-01-05
Update "Face Recognition" test to use a random suite name

consoull:
### Linked issue(s):
Towards KOL-4255

This PR updates the Face Recognition example test to use a randomized test suite name. Previously the test suite name was not randomized causing parallel test runs to share a test suite and produce false negatives.

### DIFF
--- a/examples/workflow/face_recognition_11/face_recognition_11/seed_test_suite.py
+++ b/examples/workflow/face_recognition_11/face_recognition_11/seed_test_suite.py
@@ -124,7 +124,7 @@ def main(args: Namespace) -> int:
                 ),
             )
         TestSuite(
-            name=f"{DATASET} :: {category} [FR]",
+            name=f"{args.suite_name} :: {category} [FR]",
             test_cases=[complete_test_case, *test_cases],
             reset=True,
         )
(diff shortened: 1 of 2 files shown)


munkyshi (approved):

examples/workflow/face_recognition_11/face_recognition_11/seed_test_suite.py
```diff
@@ -150,4 +150,10 @@ def main(args: Namespace) -> int:
         default=f"s3://{BUCKET}/{DATASET}/meta/metadata.csv",
         help="CSV file containing the metadata of each image. See default CSV for details.",
     )
+    ap.add_argument(
+        "--suite_name",
```
munkyshi, line 154: Actually, want to flag that I think we've used `--test-suite-name` in the past. Probably worth it to be explicit that it's a test suite for kolena workflow.

munkyshi: Quick grep indicates that other places use `--test-suite` or `--test_suites` as well -- not too consistent all over the place, but we should include `test` in the arg name.

consoull: Sounds good! I'll loop back and raise a PR updating all them to be consistent

PR #429, opened 2024-01-06
Use pre-computed RetinaFace inferences for keypoint detection example

bagalaster:

Removes the dependency on RetinaFace and uses pre-computed RetinaFace inferences instead to avoid known Tensorflow badness on M1 Macs.

Tested by running `upload_results.py`


examples/dataset/keypoint_detection/keypoint_detection/upload_results.py
```diff
@@ -13,21 +13,36 @@
...
 from keypoint_detection.model import infer_random
-from keypoint_detection.model import infer_retinaface
 from tqdm import tqdm
 
 import kolena
 from kolena.dataset import fetch_dataset
 from kolena.dataset import test
+from kolena.io import dataframe_from_csv
+from kolena.workflow.annotation import BoundingBox
+from kolena.workflow.annotation import Keypoints
```
munkyshi: We should import these from `kolena.annotation` -- it's the longer term home for these post v1 (and we don't want to reference anything in `kolena.workflow` for datasets).

munkyshi (approved):
One small comment, but approving otherwise. Thanks for cleaning this up!

examples/dataset/keypoint_detection/keypoint_detection/upload_results.py
```diff
@@ -13,21 +13,36 @@
...
+
+RETINAFACE_S3_PATH = "s3://kolena-public-examples/300-W/results/raw/retinaface.csv"
 
 
 def run(args: Namespace) -> None:
     kolena.initialize(verbose=True)
+
+    retinaface_raw_inferences_df = dataframe_from_csv(RETINAFACE_S3_PATH, storage_options={"anon": True})
```
marko-polo-cheno: May not need to touch S3 if `args.model != "RetinaFace"`

PR #437, opened 2024-01-07
Enforce id fields to be hashable

consoull:
### Linked issue(s):
Towards KOL-4382 and KOL-4383

During stress testing, the id_field uniqueness check was found to be a significant bottleneck. The cause was converting all id_fields to a dict to compare objects by value.

This commit addresses the issue by enforcing id_fields to be hashable so the conversion to dicts is not necessary. This speeds up id_field validation checks.


kolena/dataset/common.py
```diff
@@ -47,10 +47,13 @@ def validate_id_fields(id_fields: List[str], existing_id_fields: Optional[List[s
...
 
 def _validate_dataframe_ids_uniqueness(df: pd.DataFrame, id_fields: List[str]) -> None:
-    if df[id_fields].apply(lambda x: x.to_dict(), axis=1).duplicated().any():
-        raise InputValidationError(
-            f"invalid id_fields: " f"input dataframe's id field values are not unique for {id_fields}",
-        )
+    try:
+        if df[id_fields].duplicated().any():
+            raise InputValidationError(
+                f"invalid id_fields: " f"input dataframe's id field values are not unique for {id_fields}",
```
munkyshi: nit: is the double formatted string necessary?
```suggestion
                f"invalid id_fields: input dataframe's id field values are not unique for {id_fields}",
```

munkyshi (approved):

kolena/dataset/common.py
```diff
@@ -47,10 +47,13 @@ def validate_id_fields(id_fields: List[str], existing_id_fields: Optional[List[s
...
-        raise InputValidationError(
-            f"invalid id_fields: input dataframe's id field values are not unique for {id_fields}",
-        )
+    try:
+        if df[id_fields].duplicated().any():
+            raise InputValidationError(
+                f"invalid id_fields: input dataframe's id field values are not unique for {id_fields}",
+            )
+    except TypeError as e:
+        raise InputValidationError("id fields must be hashable: ", e)
```
ly-kolena, line 56: Nit: can use [raise from](https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement) for slightly clearer reporting

PR #440, opened 2024-01-07
Improve id fields hashable exception signature

consoull:
Improves id fields hashable expection signature by using `raise ... from`.


ly-kolena (approved):

munkyshi (dismissed):
This looks like it was added in python 3.9 (I don't see it in the 3.8 [docs](https://docs.python.org/3.8/tutorial/errors.html)). Can we confirm if it exists in 3.8 before merging this?

ly-kolena:
> This looks like it was added in python 3.9 (I don't see it in the 3.8 [docs](https://docs.python.org/3.8/tutorial/errors.html)). Can we confirm if it exists in 3.8 before merging this?

I tried with python 3.8.18 and it is working. Also seem like introduced for python 3 ([PEP](https://legacy.python.org/dev/peps/pep-3134/)) 

```
Python 3.8.18 (default, Sep 11 2023, 13:40:15) 
[GCC 11.2.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> try:
...   raise IOError("aaa")
... except IOError as e:
...   raise RuntimeError("bbb") from e
... 
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
OSError: aaa

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
RuntimeError: bbb
>>> 
```

consoull:
`raise ... from` added in https://peps.python.org/pep-0344/ which landed in Python 2.5

PR #443, opened 2024-01-07
Polish dataset examples

marko-polo-cheno:
Made examples slightly more consistent.


ly-kolena (approved):

.circleci/continue_config.yml
```diff
@@ -408,7 +408,7 @@ workflows:
       - example-test-dataset:
           matrix:
             parameters:
-              subproject: [ rain_forecast, speaker_diarization ]
+              subproject: [ classification, keypoint_detection, rain_forecast, speaker_diarization, text_summarization]
```
ly-kolena: should `text_summarization` be here, it's in the `large` section?

examples/dataset/classification/tests/test_multiclass_classification.py
```diff
@@ -39,12 +39,14 @@ def dataset_name() -> str:
     return f"{TEST_PREFIX} - {DATASET}"
```
ly-kolena: Not sure the best way to do this, but change scope to `package` for both `dataset_name` fixture looks to fix the issue, explained below. (Or we could just combined the 2 tests into 1)

Certainly learned something today. Looks like this fixture is called twice, once for `test__upload_dataset` and again for `test__upload_results` even though they are in the same module. As a result, `test__upload_results` is looking for a dataset that does not exist.

The tests execution order for this `tests` directory currently is

```
test_binary_classification.py::test__upload_dataset
test_multiclass_classification.py::test__upload_dataset
test_binary_classification.py::test__upload_results
test_multiclass_classification.py::test__upload_results
```

My guess is the because of the interleaving, the modules are "imported"/"executed" twice, thus defying intuition of `module` scope.

Quote from pytest tests [docs](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#fixture-scopes)

> Note
> Pytest only caches one instance of a fixture at a time, which means that when using a parametrized fixture, pytest may invoke a fixture more than once in the given scope.

examples/dataset/classification/classification/multiclass/upload_dataset.py
```diff
@@ -20,13 +20,14 @@
...
 from kolena.dataset import upload_dataset
-from kolena.workflow.annotation import ClassificationLabel
 
 
 def run(args: Namespace) -> None:
     df = pd.read_csv(f"s3://{BUCKET}/{DATASET}/raw/{DATASET}.csv", storage_options={"anon": True})
     df["ground_truth"] = df["ground_truth"].apply(lambda label: ClassificationLabel(label))
+
     kolena.initialize(verbose=True)
     upload_dataset(args.dataset, df, ID_FIELDS)
```
ly-kolena: Need to update this to new signature

```
upload_dataset(args.dataset, df, id_fields=ID_FIELDS)
```

marko-polo-cheno: Oh I see. Thank you!

PR #452, opened 2024-01-08
Update workflow examples to use consistent arg names

consoull:
Update workflow examples to use consistent arg names https://github.com/kolenaIO/kolena/pull/419#discussion_r1443338454

### DIFF
--- a/examples/workflow/object_detection_3d/README.md
+++ b/examples/workflow/object_detection_3d/README.md
@@ -95,15 +95,15 @@ Run a script using the `--help` flag for more information:
 
 ```
 $ poetry run python3 object_detection_3d/seed_test_suite.py --help
-usage: seed_test_suite.py [-h] [--test-suite TEST_SUITE] sample-file
+usage: seed_test_suite.py [-h] [--test_suite TEST_SUITE] sample_file
 
 positional arguments:
   sample_file           File containing test sample and ground truth data
 
 options:
   -h, --help            show this help message and exit
-  --test-suite TEST_SUITE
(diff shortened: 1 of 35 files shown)


examples/workflow/object_detection_3d/README.md
```diff
@@ -95,15 +95,15 @@ Run a script using the `--help` flag for more information:
...
+usage: seed_test_suite.py [-h] [--test_suite TEST_SUITE] sample_file
 
 positional arguments:
   sample_file           File containing test sample and ground truth data
 
 options:
   -h, --help            show this help message and exit
-  --test-suite TEST_SUITE
-                        Name of test suite
+  --test_suite TEST_SUITE
```
munkyshi: Thoughts on kebab case vs snake case? They both get parsed to the same variable in python, but I lean towards kebab because you already have hyphens in the option.

consoull: I prefer kebab, but biased towards what we already had. Will happily make this change

ly-kolena: kebab gets another vote here

yifanwu-kolena: second, i hate typing `_` in cli cmd

munkyshi (approved):
Approving pending kebab change.

examples/workflow/object_detection_3d/README.md
gordonhart: Fourth'd

PR #460, opened 2024-01-12
updated the upload_results doc to more accurately describe its input and output

grant-Kolena:
### Linked issue(s):

Screenshot 2024-01-15 at 1 54 01 PM

### DIFF
--- a/kolena/dataset/evaluation.py
+++ b/kolena/dataset/evaluation.py
@@ -164,7 +164,6 @@ def _validate_configs(configs: List[EvalConfig]) -> None:
                 raise IncorrectUsageError("duplicate eval configs are invalid")
 
 
-@with_event(EventAPI.Event.UPLOAD_DATASET_MODEL_RESULT)
 def _upload_results(
     dataset: str,
     model: str,
--- a/kolena/dataset/evaluation.py
+++ b/kolena/dataset/evaluation.py
@@ -98,7 +98,7 @@ def _process_result(
     return df_result_eval
 
 
-def _upload_results(
+def _send_upload_results_request(
     model: str,
     load_uuid: str,
     dataset_id: int,
@@ -165,22 +165,11 @@ def _validate_configs(configs: List[EvalConfig]) -> None:
 
 
 @with_event(EventAPI.Event.UPLOAD_DATASET_MODEL_RESULT)
-def upload_results(
+def _upload_results(
     dataset: str,
     model: str,
     results: Union[DataFrame, List[Tuple[EvalConfig, DataFrame]]],
 ) -> UploadResultsResponse:
-    """
-    This function is used for uploading the results from a specified model on a given dataset.
-
-    :param dataset: The name of the dataset.
-    :param model: The name of the model.
-    :param results: Either a DataFrame or a list of tuples, where each tuple consists of
-                    an eval configuration and a DataFrame.
-
-    :return: Containing information about the number of result rows created and updated
-    """
-
     existing_dataset = _load_dataset_metadata(dataset)
     if not existing_dataset:
         raise NotFoundError(f"dataset {dataset} does not exist")
@@ -192,9 +181,11 @@ def upload_results(
     load_uuid = init_upload().uuid
 
     _validate_configs([cfg for cfg, _ in results])
+    total_rows = 0
     for config, df_result_input in results:
         log.info(f"uploading test results with configuration {config}" if config else "uploading test results")
         if isinstance(df_result_input, pd.DataFrame):
+            total_rows += df_result_input.shape[0]
             validate_dataframe_ids(df_result_input, id_fields)
             validate_dataframe_have_other_columns_besides_ids(df_result_input, id_fields)
             df_results = _process_result(config, df_result_input, id_fields)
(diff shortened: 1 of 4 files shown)


kolena/dataset/evaluation.py
```diff
@@ -169,8 +169,11 @@ def upload_results(
     :param dataset: The name of the dataset.
     :param model: The name of the model.
     :param results: Either a DataFrame or a list of tuples, where each tuple consists of
-                    an eval configuration and a DataFrame.
-    :return: None
+                    an eval configuration and a DataFrame. The DataFrame type has been aliased to be either a pandas
+                    DataFrame or an iterator of pandas DataFrame, this allows for batch upload when passing iterators.
```
munkyshi: I'd prefer to link to the alias for `DataFrame` that we have + have the description there.

kolena/dataset/evaluation.py
```diff
@@ -169,8 +169,11 @@ def upload_results(
...
+
+    :return: An UploadResultsResponse object containing information about the number of result rows created
+             and updated
```
munkyshi: It's not very meaningful to refer to `UploadResultsResponse` when we don't have the type documented -- how will users know what the fields to look into actually are?

kolena/dataset/evaluation.py
```diff
@@ -169,8 +169,11 @@ def upload_results(
     :param dataset: The name of the dataset.
     :param model: The name of the model.
     :param results: Either a DataFrame or a list of tuples, where each tuple consists of
-                    an eval configuration and a DataFrame.
-    :return: None
+                    an eval configuration and a DataFrame. The DataFrame type has been aliased to be either a pandas
```
ly-kolena: Is there a way for `DataFrame` type to be expanded or linkable to the source definition? Exaplaining type alias seems a bit verbose.

grant-Kolena: currently there is a not very descriptive tooltip, seems only the python builtin types gets a link
Screenshot 2024-01-12 at 11 28 22 AM

kolena/dataset/evaluation.py
grant-Kolena: I will add the doc for it

grant-Kolena:
We have description linked for every self defined type now
Screenshot 2024-01-16 at 10 14 25 AM

kolena/dataset/evaluation.py
grant-Kolena: They are expandable now

kolena/dataset/evaluation.py
gordonhart: This actually raises a larger problem: `UploadResultsResponse` is an internal API object and shouldn't leak into the external-facing API as the response from `upload_results`.

What are users expected to do with an `UploadResultsResponse`? We should consider changing this function to return `None`.

ly-kolena: @gordonhart What would you recommend if we want to make `UploadResultsResponse` public?

grant-Kolena: Or can we create a public data model, I still think number of rows created/ updated is valuable info

kolena/dataset/evaluation.py
```diff
@@ -197,12 +196,33 @@ def upload_results(
...
                     id_column_validated = True
+                total_rows += df_result.shape[0]
                 df_results = _process_result(config, df_result, id_fields)
                 upload_data_frame(df=df_results, batch_size=BatchSize.UPLOAD_RECORDS.value, load_uuid=load_uuid)
 
-    response = _upload_results(model, load_uuid, existing_dataset.id)
+    response = _send_upload_results_request(model, load_uuid, existing_dataset.id)
     log.info(
         f"uploaded test results for model '{model}' on dataset '{dataset}': "
+        f"There are a total of {total_rows} rows in the results input"
```
ly-kolena: Prefer to keep log succinct, suggest `{} uploaded, {} inserted, {} updated`

grant-Kolena: sounds good

ly-kolena: Do we also want to include result link to web UI?

kolena/dataset/evaluation.py
```diff
@@ -51,7 +51,14 @@
 from kolena.errors import NotFoundError
 
 EvalConfig = Optional[Dict[str, Any]]
+"""
+User defined configuration for evaluating results, for example `{"threshold": 7}`.
+"""
 DataFrame = Union[pd.DataFrame, Iterator[pd.DataFrame]]
+"""
+A type alias representing a DataFrame, which can be either a pandas DataFrame
+or an iterator of pandas DataFrames, data will be processed in chunks when iterator is used.
```
ly-kolena: Is `data will be processed...` needed? Seems `iterator` by itself is sufficient. (side nit, `in chunks` of iterator is confusing)

kolena/dataset/evaluation.py
grant-Kolena: We do have a task of logging the result url, do you mean we should return the result url in the response?

kolena/dataset/evaluation.py
grant-Kolena: how about we change it to `A type alias representing a DataFrame, which can be either a pandas DataFrame
or an iterator of pandas DataFrames`?

kolena/dataset/evaluation.py
ly-kolena: I meant only logging the result url, fine for the other ticket to cover it.

kolena/dataset/evaluation.py
ly-kolena: Sounds good.

kolena/dataset/evaluation.py
grant-Kolena: I still think this is valuable to see `{} uploaded, {} inserted, {} updated`. I don't know a good way to get these info for a result file upload from sdk on the UI, plus this is one less click

ly-kolena (approved):

PR #466, opened 2024-01-15
enable mypy for examples/workflow

sandalns:
### Linked issue(s):
KOL-4575
Enables the temporarily removed mypy checks from example/workflow and adds them back. mypy previously was also not ran on the `experimental/classification` or experimental/object_detection` directories.


sandalns:
Main reasons for `type: ignores`

1) I ignored errors around numpy return and parameter types after time boxing 1hr and not making progress
2) Many errors around the expected type being `kolena.workflow.TestSample` and not `example.workflow.TestSample` etc

Any suggestions here would be appreciated.

grant-Kolena (approved):

kolena/_experimental/object_detection/workflow.py
```diff
@@ -17,7 +17,7 @@
 try:
     from typing import Literal
 except ImportError:
-    from typing_extensions import Literal
+    from typing_extensions import Literal  # type: ignore
```
yifanwu-kolena: `typing.Literal` was introduced in python 3.8, considering we already dropped python 3.7 support after 1.0, i think we can remove these try catch everywhere

sandalns: Will remove

sandalns: done

examples/workflow/automatic_speech_recognition/automatic_speech_recognition/utils.py
```diff
@@ -89,13 +89,13 @@ def generate_diff_word_level(reference: str, candidate: str) -> Dict[str, Any]:
...
 
-def preprocess_transcription(txt: Union[GroundTruth, Inference]) -> str:
+def preprocess_transcription(txt: str) -> str:
     """
     Preprocesses and standardizes text to prepare for metrics evaluations.
     Removes punctuation, changes text to lower case, and converts all NumWords to Numbers.
     """
     num = NumWordsToNum()
-    txt = re.sub(r"[^\w\s]", "", txt.transcription.label.lower())
+    txt = re.sub(r"[^\w\s]", "", txt)
```
yifanwu-kolena, line 98: based on the usage of `preprocess_transcription`(examples/workflow/automatic_speech_recognition/automatic_speech_recognition/evaluator.py), i don't think we should modify the code at here

```py
def compute_test_sample_metrics(gt: GroundTruth, inf: Inference) -> TestSampleMetric:
    gt = preprocess_transcription(gt)
    inf = preprocess_transcription(inf)
```

yifanwu-kolena: nvm, i noticed that we updated the usage part as well

yifanwu-kolena (approved):

PR #453, opened 2024-01-09
docs: how to shape a CSV file for ingestion

yifanwu-kolena:
### Please check if the PR fulfills these requirements

- [ ] Include reference to internal ticket and/or GitHub issue "Fixes #NNNN" (if applicable)
- [ ] Relevant tests for the changes have been added
- [ ] Relevant docs have been added / updated

PR #470, opened 2024-01-18
style: add markdown lint

yifanwu-kolena:
### Linked issue(s):

follow up of https://github.com/kolenaIO/kolena/pull/453#discussion_r1447579988

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -1,24 +1,40 @@
+# https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md
 default: true
 
+# Hard tabs
 MD010:
   spaces_per_tab: 4
 
+# Line length
 MD013:
   code_blocks: false
   headings: false
   line_length: 120
   tables: false
 
+# Heading levels should only increment by one level at a time
 MD001: false
+# Inconsistent indentation for list items at the same level
 MD005: false
+# Multiple headings with the same content
 MD024: false
+# Ordered list item prefix
 MD029: false
+# Fenced code blocks should be surrounded by blank lines
 MD031: false
+# Inline HTML
 MD033: false
+# Bare URL used
 MD034: false
+# Emphasis used instead of a heading
 MD036: false
+# Fenced code blocks should have a language specified
 MD040: false
+# First line in a file should be a top-level heading
 MD041: false
+# Images should have alternate text (alt text)
 MD045: false
+# Code block style
 MD046: false
+# Reference links and images should use a label that is defined
 MD052: false
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -9,6 +9,7 @@ MD013:
   tables: false
 
 MD001: false
+MD005: false
 MD024: false
 MD029: false
 MD031: false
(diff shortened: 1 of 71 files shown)


docs/metrics/geometry-matching.md
```diff
@@ -195,17 +192,16 @@ the IoU threshold to be considered as a valid match.
 
 ??? info "Pseudocode: PASCAL VOC Matching"
 
-	1. Loop through all images in your dataset;
```
yifanwu-kolena, line 198: tabs to spaces

.markdownlint.yaml
```diff
@@ -0,0 +1,24 @@
+default: true
+
+MD010:
```
ly-kolena, line 5: Nit: can we use comment to inform what each rule is, so that I don't have to google the library for its meaning?

docs/index.md
```diff
@@ -9,12 +9,14 @@ hide:
   ![Kolena](assets/images/wordmark-white.svg#only-dark){ width="400" }
 </figure>
 
+<!-- markdownlint-disable MD013 -->
```
gordonhart: It's not great to have to disable this all over the place — do we ever need `MD013`?

yifanwu-kolena: [MD013](https://github.com/DavidAnson/markdownlint/blob/main/doc/md013.md) is the rule for line length

.markdownlint.yaml
yifanwu-kolena: good idea, just added the comment with the rules url as well

.markdownlint.yaml
```diff
@@ -0,0 +1,40 @@
...
+default: true
+
+# Hard tabs
+MD010:
+  spaces_per_tab: 4
+
+# Line length
+MD013:
+  code_blocks: false
+  headings: false
```
ly-kolena: Nit: do we not want to apply this rule to code_blocks and headings?

README.md
```diff
@@ -4,12 +4,14 @@
   </a>
 </p>
 
+<!-- markdownlint-disable MD013 -->
```
ly-kolena: Can this section be line-broken to comply to the length rule without exception? (assuming this can help

> This rule has an exception when there is no whitespace beyond the configured line length. This allows you to include items such as long URLs without being forced to break them in the middle.

ly-kolena (approved):

docs/metrics/wer-cer-mer.md
```diff
@@ -66,14 +70,20 @@ $$
 $$
 
 ### Example
+
 Let's calculate the word error rate between the following reference and candidate texts:
 
+<!-- markdownlint-disable MD013 -->
```
ly-kolena: For my knowledge, why is this needed when we have `table` disabled in MD013 configuration?

yifanwu-kolena: after updating the table syntax to have explicit header separator, it is recognized by the linter properly.
in term of other tables nested in the ??? block, it will be recognized as code rather than table

.markdownlint.yaml
yifanwu-kolena: just enabled

PR #474, opened 2024-01-19
Disable docs publish as part of the hotfix

sandalns:
Disable docs publish to avoid clobbering the docs as part of this hotfix.


munkyshi (approved):
Would be nice to have a comment inline for "this used to be here but we're getting rid of it for the hotfix"

PR #475, opened 2024-01-19
Experimental code to upload dataset embeddings

nankolena:
Part of KOL-4788

### Linked issue(s):
https://linear.app/kolena/issue/KOL-4788/update-kolena-embeddings-to-support-uploading-dataset-embeddings

Add client-side code for uploading embeddings on a given dataset. Also add integration tests for embedding-related experimental code to CI.

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -56,7 +56,7 @@ jobs:
           key: &ci-base-cache ci-cache-<< parameters.python-version >>-<< parameters.extras >>-{{ checksum "pyproject.toml" }}
       - when:
           condition:
-            equal: [ "none", << parameters.extras >> ]
+            not_equal: [ "metrics", << parameters.extras >> ]
           steps:
             - run:
                 name: Install Dependencies
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -435,11 +435,21 @@ workflows:
           matrix:
             parameters:
               python-version: [ "3.9.18" ]
-              extras: [ "metrics", "embeddings" ]
+              extras: [ "metrics" ]
           pytest-group: _experimental
           enabled: true
           requires:
             - ci-base-<< matrix.python-version >>-<< matrix.extras >>
+      - integration-test:
+          name: integration-test-experimental-<< matrix.python-version >>-<< matrix.extras >>
+          matrix:
+            parameters:
+              python-version: [ "3.9.18" ]
+              extras: [ "embeddings" ]
+          pytest-group: _experimental
+          enabled: true
+          requires:
+            - ci-base-<< matrix.python-version >>-none
       - integration-test-dataset:
           name: integration-test-dataset-<< matrix.python-version >>-<< matrix.extras >>
           matrix:
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -56,7 +56,7 @@ jobs:
           key: &ci-base-cache ci-cache-<< parameters.python-version >>-<< parameters.extras >>-{{ checksum "pyproject.toml" }}
       - when:
           condition:
-            not_equal: [ "metrics", << parameters.extras >> ]
+            equal: [ "none", << parameters.extras >> ]
           steps:
             - run:
                 name: Install Dependencies
@@ -233,14 +233,11 @@ jobs:
                     -o junit_family=legacy --junitxml=test-results/result.xml $TESTFILES
       - when:
           condition:
-            equal: [ "_experimental", << parameters.pytest-group >> ]
+            equal: [ "metrics", << parameters.extras >> ]
           steps:
             - run:
-                name: Run experimental << parameters.extras >> integration tests
-                command: |
-                  TEST_MARKER="<< parameters.extras >>"
-                  poetry run pytest --cache-clear -m $TEST_MARKER -vv --cov=kolena --cov-branch \
-                    tests/integration/_experimental
+                name: Run experimental metrics integration tests
+                command: poetry run pytest --cache-clear -m 'metrics' -vv --cov=kolena --cov-branch tests/integration/
       - when:
           # Generate coverage only from one python version
           condition:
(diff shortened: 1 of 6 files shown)


tests/integration/_experimental/search/test_embeddings.py
```diff
@@ -62,10 +69,79 @@ def test_upload_embeddings(embedding: np.ndarray) -> None:
...
+
+@pytest.fixture(scope="module", autouse=True)
+def dataset_name() -> str:
+    name = with_test_prefix(f"{__file__}::test__embedding_dataset {uuid.uuid4()}")
+    datapoints = [dict(locator=f"locator-{i}", value=i) for i in range(N_DATAPOINTS)]
+    upload_dataset(name, pd.DataFrame(datapoints), id_fields=["locator"])
+    return name
+
+
+@pytest.mark.skip("disabled until server side changes are deployed")
```
nankolena: Will uncomment these tests after https://github.com/kolenaIO/shipyard/pull/1823 is in production. Was successful running them locally:

```
(kolena-py3.10) ➜  kolena git:(nan/KOL-4788) ✗ poetry run pytest -vv tests/integration/_experimental/search/test_embeddings.py::test__upload_dataset_embeddings
====================================================================================== test session starts ======================================================================================
platform darwin -- Python 3.10.13, pytest-7.4.4, pluggy-1.3.0 -- /Users/nanwei/Library/Caches/pypoetry/virtualenvs/kolena-hYtzzkLc-py3.10/bin/python
cachedir: .pytest_cache
rootdir: /Users/nanwei/Desktop/kolena_dev/kolena
configfile: pyproject.toml
plugins: cov-4.1.0, typeguard-4.1.5, depends-1.0.1
collected 4 items                                                                                                                                                                               

tests/integration/_experimental/search/test_embeddings.py::test__upload_dataset_embeddings[embedding0] PASSED                                                                             [ 25%]
tests/integration/_experimental/search/test_embeddings.py::test__upload_dataset_embeddings[embedding1] PASSED                                                                             [ 50%]
tests/integration/_experimental/search/test_embeddings.py::test__upload_dataset_embeddings[embedding2] PASSED                                                                             [ 75%]
tests/integration/_experimental/search/test_embeddings.py::test__upload_dataset_embeddings[embedding3] PASSED                                                                             [100%]

```

sandalns (approved):

kolena/_experimental/search/embeddings.py
```diff
@@ -65,3 +77,61 @@ def upload_embeddings(key: str, embeddings: List[Tuple[str, np.ndarray]]) -> Non
     krequests.raise_for_status(res)
     data = from_dict(data_class=API.UploadEmbeddingsResponse, data=res.json())
     log.success(f"uploaded embeddings for key '{key}' on {data.n_samples} samples")
+
+
+def upload_dataset_embeddings(dataset_name: str, key: str, df_embedding: pd.DataFrame) -> None:
```
yifanwu-kolena, line 82: do we support partial dataset embedding upload? 
what happen if they provide different model locators for the datapoints in the same dataset? do we handle it gracefully when searching the embeddings? or do we need to forbid it?

nankolena: > do we support partial dataset embedding upload?

Yes we do support upload embedding to a subset of the dataset. I will add an integ test on this. The existing behavior for this is that only test samples with embeddings are returned in studio (i.e. this `sort by` in essence acts like a filter). We can probably handle this better in the UI messaging as a next step.

> what happen if they provide different model locators for the datapoints in the same dataset? do we handle it gracefully when searching the embeddings?

The current experience is a bit rough IMO. One can upload multiple embeddings on the same datapoint for different models. However, during search, only [one model key](https://github.com/kolenaIO/shipyard/blob/7df225eb45d4882e86ec4c431fd4cdcb9b7385e1/galleon/galleon/services/v2/search/service.py#L175) is used which is specified in `galleon.galleon` at the tenant level. Future work is warranted to make this better - for example allow customized or potentially multi-model per tenant.

yifanwu-kolena (approved):

.circleci/continue_config.yml
```diff
@@ -437,6 +437,16 @@ workflows:
           enabled: true
           requires:
             - ci-base-<< matrix.python-version >>-<< matrix.extras >>
+      - integration-test:
+          name: integration-test-experimental-embedding-<< matrix.python-version >>-<< matrix.extras >>
```
ly-kolena, line 441: side question: did we not run this test before?

tests/integration/_experimental/search/test_embeddings.py
```diff
@@ -44,8 +50,8 @@
         np.array([], dtype=np.float16),
     ],
 )
-def test_upload_embeddings(embedding: np.ndarray) -> None:
-    test_case_name = with_test_prefix(f"{__file__} test_upload_embeddings")
+def test__upload_embeddings(embedding: np.ndarray) -> None:
+    test_case_name = with_test_prefix(f"{__file__} test_upload_embeddings {uuid.uuid4()}")
```
ly-kolena: For my own knowledge, why is this additional uuid needed?

.circleci/continue_config.yml
nankolena: No it was not run in CI before. My take is that it was [intentionally skipped as a todo](https://github.com/kolenaIO/kolena/pull/475/files#diff-8450575a09d4183f4c781d02b723d4c26fa3f3a313d9861f9bca4259e70e6e42L37) but lost track

tests/integration/_experimental/search/test_embeddings.py
nankolena: I don't remember the specific error code, but iirc this test was intended to repeat on the 4 different arguments of embedding, and the first one would be successful while the rest would fail, complaining that a test case with this name already exists. An alternative approach is to re-use the test case as a fixture, similar to what I did with `dataset_name` below.

kolena/_experimental/search/_internal/datatypes.py
```diff
@@ -24,5 +28,21 @@ class LocatorEmbeddingsDataFrameSchema(pa.SchemaModel):
 
     embedding: Series[pa.typing.String] = pa.Field(coerce=True)
     """
-    Embedding vector (`np.ndarray`) corresponding to a searchable representation of the sample.
+    Embedding vector (base64-encoded string of `np.ndarray`) corresponding to a searchable representation of the sample.
+    """
+
+
+class DatasetEmbeddingsDataFrameSchema(pa.SchemaModel):
```
ly-kolena, line 35: Do we apply any validation for this schema?

nankolena: Could you elaborate the question? This schema is [used to validate the df](https://github.com/kolenaIO/kolena/pull/475/files#diff-1dce437eca8dea142eb32b0f04f31ccf94f51f6c01304f6729f861482963b7efR121) before uploading

ly-kolena: I meant what constraints do we need to enforce on these values, minimal string length, fixed values, etc?

We're using `coerce=True`, which from my reading would convert input values to target type first. Do we need this?

ly-kolena (approved):

kolena/_experimental/search/_internal/datatypes.py
nankolena: Regarding the content (length, values etc) I think it's difficult to put on guards. The model locator/id objects/embedding string could each vary in size. Good question about `coerce=True`, tbh I just took this logic from the `LocatorEmbeddingsDataFrameSchema` above assuming there was some rationale behind it.

PR #479, opened 2024-01-25
Fix state handling in experimental object detection evaluator

gordonhart:
Fixes observed bug due to stateful objects being assigned as class members instead of instance members.

### DIFF
--- a/kolena/_experimental/object_detection/evaluator_multiclass.py
+++ b/kolena/_experimental/object_detection/evaluator_multiclass.py
@@ -39,6 +39,7 @@ from kolena._experimental.object_detection.utils import compute_pr_curve
 from kolena._experimental.object_detection.utils import compute_pr_plot_multiclass
 from kolena._experimental.object_detection.utils import filter_inferences
 from kolena.workflow import Evaluator
+from kolena.workflow import EvaluatorConfiguration
 from kolena.workflow import Plot
 from kolena.workflow.annotation import ScoredLabel
 from kolena.workflow.metrics import f1_score as compute_f1_score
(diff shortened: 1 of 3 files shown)


kolena/_experimental/object_detection/evaluator_multiclass.py
```diff
@@ -60,25 +61,29 @@ class MulticlassObjectDetectionEvaluator(Evaluator):
     For additional functionality, see the associated [base class documentation][kolena.workflow.evaluator.Evaluator].
     """
 
-    threshold_cache: Dict[str, Dict[str, float]] = {}  # configuration -> label -> threshold
+    threshold_cache: Dict[str, Dict[str, float]]  # configuration -> label -> threshold
```
consoull, line 64: Should the class level caches be removed now that we are defining them at the instance level?

gordonhart: This is just a type hint — I don't see any issues with leaving them there. Do you?

consoull: `MulticlassObjectDetectionEvaluator.threshold_cache` will exist and always be `None`. Demonstrating with `ObjectDetectionEvaluator` we have `dynamic_evaluator` defined at both the class and instance level (only assigned at instance).
```
(Pdb) l
145  	    evaluator = setup_evaluator()
146  	
147  	    # runs the evaluation
148  	    test_suite = TestSuite(test_suite_name)
149  	    test(model, test_suite, evaluator, reset=True)
150  ->	    breakpoint()

(Pdb) evaluator
<kolena._experimental.object_detection.evaluator.ObjectDetectionEvaluator object at 0x2a3f4b1d0>
(Pdb) evaluator.dynamic_evaluator
<kolena._experimental.object_detection.evaluator_multiclass.MulticlassObjectDetectionEvaluator object at 0x2a0b97fd0>
(Pdb) ObjectDetectionEvaluator.dynamic_evaluator == None
True
```

ly-kolena (approved):

PR #481, opened 2024-01-25
chore: expose sources for uploading

yifanwu-kolena:
### Linked issue(s)

counterpart of: https://github.com/kolenaIO/shipyard/pull/1854

- expose the sources for uploading endpoints
- refactor to have these wrapper mapping: `upload_x` => `_upload_x` => `_prepare_upload_x_request` + `_send_upload_x_request`

### DIFF
--- a/kolena/_api/v2/dataset.py
+++ b/kolena/_api/v2/dataset.py
@@ -33,6 +33,7 @@ class RegisterRequest:
     name: str
     id_fields: List[str]
     uuid: str
+    sources: Optional[List[str]] = None
 
 
 @dataclass(frozen=True)
--- a/kolena/_api/v2/dataset.py
+++ b/kolena/_api/v2/dataset.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from enum import Enum
+from typing import Dict
 from typing import List
 from typing import Optional
 
@@ -33,7 +34,7 @@ class RegisterRequest:
     name: str
     id_fields: List[str]
     uuid: str
-    sources: Optional[List[str]] = None
+    sources: Optional[List[Dict[str, str]]] = None
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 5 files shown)


kolena/_api/v2/dataset.py
```diff
@@ -33,6 +34,7 @@ class RegisterRequest:
     name: str
     id_fields: List[str]
     uuid: str
+    sources: Optional[List[Dict[str, str]]]
```
gordonhart, line 37: Should this have a `None` default value?

yifanwu-kolena: i think we don't need the default value at here.
this is the caller side definition, and we already pass the `DEFAULT_SOURCES` as default value in the proper functions.
we have the default value `None` on the server side for backward compatibility

kolena/dataset/dataset.py
```diff
@@ -280,13 +267,59 @@ def upload_dataset(
...
+def _upload_dataset(
+    name: str,
+    df: Union[pd.DataFrame, Iterator[pd.DataFrame]],
+    *,
+    id_fields: Optional[List[str]] = None,
+    sources: Optional[List[Dict[str, str]]] = DEFAULT_SOURCES,
+) -> None:
+    prepared_id_fields, load_uuid = _prepare_upload_dataset_request(name, df, id_fields=id_fields)
+
+    data = _send_upload_dataset_request(name, prepared_id_fields, load_uuid, sources=sources)
```
ly-kolena, line 295: Are extractions of these 2 functions needed? I don't see them being shared/reused.

yifanwu-kolena: they are mainly for easier reading rather than reusing.

`_send_upload_x_request` is a good abstraction i noticed in the evaluation, b/c it's the wrapper to call the sdk-server endpoint. `_prepare_upload_dataset_request` method might be arguable in someway even i feel like it's easier to track.

ly-kolena (approved):

PR #486, opened 2024-01-27
Upgrade to Pydantic V2

gordonhart:
Upgrading `pydantic` dependency to pick up their [now-mature V2](https://docs.pydantic.dev/latest/migration/).

Fixes KOL-4940
Fixes KOL-4972


kolena/_utils/validators.py
```diff
@@ -13,16 +13,16 @@
...
-from pydantic import Extra
+from pydantic import ConfigDict
 
 from kolena._utils.consts import FieldName
 
 
-# Pydantic configuration for dataclasses and @validate_arguments decorators
-class ValidatorConfig:
-    arbitrary_types_allowed = True
-    smart_union = True
```
munkyshi, line 24: I see that this was removed from Pydantic -- do they provide an alternative? I remember we added this because numerics were being coerced into strings because of union types.

kolena/_api/v1/core.py
```diff
@@ -57,9 +57,6 @@ class DeleteRequest:
         id: int
 
 
-Model.LoadAllResponse.__pydantic_model__.update_forward_refs()  # type: ignore
```
munkyshi, line 60: Do we not need this anymore? It looks like the migration guide states that this got renamed:
Screenshot 2024-01-31 at 10 32 37

munkyshi (commented):
Two main questions

docs/installing-kolena.md
```diff
@@ -29,18 +29,6 @@ or [Poetry](https://python-poetry.org/):
...
-    Kolena uses `pydantic` for data validation, but is compatible only with `pydantic` V1. Prior to `0.76.0`, the
-    `kolena` client did not set an upper bound on compatible `pydantic` version, and as a result, fresh
-    installations of older versions of `kolena` may install incompatible version of `pydantic`, resulting in errors such
-    as:
-
-    ```
-    AttributeError: type object 'SingleProcessResponse' has no attribute '__pydantic_model__'
-    ```
-
-    If you see the error above, please install the latest version of `kolena`.
```
munkyshi, line 42: 🎉

kolena/_utils/validators.py
gordonhart: According to https://github.com/pydantic/pydantic/discussions/7264, `smart_union=True` is now the default behavior:

> By default, v2 has the "smart union" behaviour. If you want to keep the behaviour of v1 you can use ...

kolena/_api/v1/core.py
gordonhart: Nothing seems to break when I remove it, so I see no reason to keep it. The reason I removed it here is not because it was renamed but because [Pydantic V2 dataclasses no longer have a `__pydantic_model__`](https://docs.pydantic.dev/latest/migration/#changes-to-dataclasses):

> Pydantic dataclasses no longer have an attribute `__pydantic_model__`, and no longer use an underlying `BaseModel` to perform validation or provide other functionality.

kolena/_utils/datatypes.py
```diff
@@ -69,7 +69,8 @@ def _double_under(input: str) -> bool:
 def _allow_extra(cls: Type[T]) -> bool:
     # `pydantic.dataclasses.is_built_in_dataclass` would have false-positive when a stdlib-dataclass decorated
     # class extends a pydantic dataclass
-    return "__pydantic_model__" in vars(cls) and cls.__pydantic_model__.Config.extra == Extra.allow  # type: ignore
+    pydantic_config: dict = getattr(cls, "__pydantic_config__", {})
```
yifanwu-kolena, line 72: `Dict` for style consistency before we dropping python 3.8 support

yifanwu-kolena (approved):

PR #522, opened 2024-02-22
Persist additional metadata on BBoxes to results during evaluation

nankolena:
### Linked issue(s)
https://linear.app/kolena/issue/KOL-5283/update-sdk-client-to-persist-bouding-box-metadata-when-uploading

The goal is to support object-level metrics for OD (for example, "what's my recall metrics for bounding boxes on the left vs on the right). In the datapoints, each LabeledBoundingBox could be associated with additional metadata. These additional metadata however, other than the `label`, is ignored when we upload results. Given that the evaluation logic (IoU threshold etc) is on the client side, it's infeasible for the server to associate one specific inference to a ground truth after-the-fact.

This PR persists the bounding box metadata to the applicable column when uploading results (`TP`, `TN`, `unmatched_ground_truth`, `matched_inference`).

An alternative is to instead persist only the index of the ground truth to which an inference is matched against. While it saves storage, it makes the metrics computation logic very complicated - the server needs to query results table, get inference, then somehow join with the datapoints using the bounding box index (not sure if do-able directly in SQL), etc which is both complicated in logic and also likely be slower to compute metrics. 

### Verification

Ran the example OD script with changes in this branch (adding `supercategory` field to the bboxes). [Confirmed that the results have this additional field too](https://trunk.kolena.com/nan/dataset/studio?datasetId=64&datasetCommit=aa561275eb825d8afa6b1f011e0df149b32a8ad1&models=N4IglgJiBcBsCcBWANCAxgewHYDMwHMBXAJwEMAXMbGUSGWAXwaA&offset=0&datapointId=64958446):
image

### DIFF
--- a/examples/dataset/object_detection_2d/pyproject.toml
+++ b/examples/dataset/object_detection_2d/pyproject.toml
@@ -7,7 +7,8 @@ license = "Apache-2.0"
 
 [tool.poetry.dependencies]
 python = ">=3.9,<3.12"
-kolena = {extras = ["metrics"], version = "^1.4.0"}
+# TODO: use official release
+kolena = {extras = ["metrics"], path = "../../..", develop = true}
 s3fs = "^2023.5.0"
 fsspec = "^2023.5.0"
 pandera = ">=0.9.0,<0.16"
--- a/examples/dataset/object_detection_2d/pyproject.toml
+++ b/examples/dataset/object_detection_2d/pyproject.toml
@@ -7,8 +7,7 @@ license = "Apache-2.0"
 
 [tool.poetry.dependencies]
 python = ">=3.9,<3.12"
-# TODO: use official release
-kolena = {extras = ["metrics"], path = "../../..", develop = true}
+kolena = {extras = ["metrics"], version = "^1.4.0"}
 s3fs = "^2023.5.0"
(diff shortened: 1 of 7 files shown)


examples/dataset/object_detection_2d/pyproject.toml
```diff
@@ -7,7 +7,8 @@ license = "Apache-2.0"
 
 [tool.poetry.dependencies]
 python = ">=3.9,<3.12"
-kolena = {extras = ["metrics"], version = "^1.4.0"}
+# TODO: use official release
+kolena = {extras = ["metrics"], path = "../../..", develop = true}
```
diegokolena: This is not necessary right?

nankolena: Not necessary for the example to be functional, but without this, running the example right now will not persist the `supercategory` field in the model results. I can update this after the next release assuming this PR is merged

nankolena: Stashed these changes for now. I will update the example after today's release, assuming this PR is merged.

ly-kolena (approved):

yifanwu-kolena (approved):

PR #525, opened 2024-02-28
Stash KITTI 3D OD work

munkyshi:
Raw workflow -- pretty un-polished.
- Not sure if the pyproject toml works exactly as is
- Hard coded references to `s3://kolena-dev-data/` instead of `kolena-public-examples`
- Doesn't do any matching / tp/fp/fn qualification of 3d bboxes


gordonhart:
> - Hard coded references to `s3://kolena-dev-data/` instead of `kolena-public-examples`

We can't use `kolena-dev-data` in public examples, as most are not able to access this bucket.

PR #530, opened 2024-03-01
feat: Attempt to initialize client if uninitialized

consoull:
### Linked issue(s)
Fixes KOL-5272

Updates `kolena_initialized` wrapper to attempt to initialize the client if uninitialized instead of immediately raising an error.

### DIFF
--- a/kolena/_utils/state.py
+++ b/kolena/_utils/state.py
@@ -27,6 +27,7 @@ import requests
 import kolena
 import kolena._api.v1.token as API
 from kolena._utils import krequests
+from kolena._utils.log import info
 from kolena._utils.serde import from_dict
 from kolena.errors import InvalidClientStateError
 from kolena.errors import InvalidTokenError
@@ -100,12 +101,9 @@ class _ClientState:
         self.additional_request_headers = additional_request_headers or self.additional_request_headers
 
     def assert_initialized(self) -> None:
-        if self.base_url is None:
-            raise InvalidClientStateError("missing base_url")
         if self.jwt_token is None:
-            raise UninitializedError("client has not been initialized via kolena.initialize(...)")
-        if self.api_token is None:
-            raise InvalidClientStateError("missing client api_token")
+            info("Attempting to initialize client...")
+            kolena.initialize(verbose=True)
 
     def reset(self) -> None:
         # note that base_url remains set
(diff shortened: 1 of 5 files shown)


consoull:
@gordonhart we have to hold off on updating examples to drop `kolena.initialize()` until we have a release with this change so we can pin the minimum version

kolena/_utils/state.py
```diff
@@ -129,21 +119,17 @@ def get_client_state() -> _ClientState:
...
 
 
-def is_client_initialized() -> bool:
-    try:
-        get_client_state().assert_initialized()
-    except (InvalidClientStateError, UninitializedError):
-        return False
-    return True
+def is_client_uninitialized() -> bool:
+    return get_client_state().jwt_token is None
```
gordonhart, line 123: Should this not follow the same logic in the removed `assert_initialized`?

consoull: I view the `base_url` guard as dead code, we provide a default in the constructor (probably shouldn't be an optional) https://github.com/kolenaIO/kolena/blob/trunk/kolena/_utils/state.py#L54.
What really signals the client is initialized is the presence of a jwt token, the auth token is a required to retrieve the jwt token https://github.com/kolenaIO/kolena/blob/trunk/kolena/_utils/state.py#L186

kolena/_utils/state.py
```diff
@@ -129,21 +119,17 @@ def get_client_state() -> _ClientState:
...
 def kolena_initialized(func: Callable) -> Callable:
-    """Raises InvalidKolenaStateError if not initialized"""
+    """Attempts to initialize client if not initialized"""
 
     @functools.wraps(func)
     def wrapper(*args: Any, **kwargs: Any) -> Any:
-        client_state = get_client_state()
-        client_state.assert_initialized()
+        if is_client_uninitialized():
+            kolena.initialize(verbose=True)
```
ly-kolena: Should we use `verbose=True` here?

kolena/_utils/state.py
```diff
@@ -129,21 +119,17 @@ def get_client_state() -> _ClientState:
     return CLIENT_STATE.get(_client_state)
 
 
-def is_client_initialized() -> bool:
```
ly-kolena, line 132: Nit: from a class' public interface, `is_client_initialized` reads more natural than `is_client_uninitialized`.

consoull: We only use the function to check if the client is uninitialized eg. `if not is_client_initialized()`. I prefer `if is_client_uninitialized()` but don't feel strongly

kolena/_utils/state.py
consoull: Ya I think we should use default args. Will update

ly-kolena (approved):

PR #533, opened 2024-03-05
Update docs to account for auto-initialization

gordonhart:
Follow-on to #530.

Fixes KOL-5458

### DIFF
--- a/docs/installing-kolena.md
+++ b/docs/installing-kolena.md
@@ -48,7 +48,8 @@ Certain metrics computation functionality depends on additional packages like
 
 ## Initialization
 
-Once you have `kolena` installed, initialize a session with `kolena.initialize(api_token=token)`.
+Once you have `kolena` installed, Kolena will automatically initialize with a token present in your environment in the
+`KOLENA_TOKEN` variable or in your `.netrc` file.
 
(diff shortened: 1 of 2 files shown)


docs/installing-kolena.md
```diff
@@ -48,25 +48,27 @@ Certain metrics computation functionality depends on additional packages like
...
+To get an API token, visit the [:kolena-developer-16: Developer](https://app.kolena.com/redirect/developer) page to
+generate a token, then set the `KOLENA_TOKEN` variable in your environment:
 
 ```bash
 export KOLENA_TOKEN="********"
 ```
 
-With the `KOLENA_TOKEN` environment variable set, initialize a client session:
+By default, sessions are automatically initialized when a function requiring initialization is called. To configure
+your session, e.g. to enable extra logging via `verbose=True`, manually initialize a session by running the following:
```
consoull:
```suggestion
your session, e.g. to enable extra logging via `verbose=True`, you may call [`kolena.initialize`](./reference/initialize.md) directly:
```
nit: I like this wording is a slightly better since it encourages the user to check out the initialize options; "..running the following" is more likely to lead to copy paste without exploring imo

consoull (approved):

ly-kolena (approved):

PR #538, opened 2024-03-06
Update object detection 2d example s3 path

consoull:
Updates object detection 2d example s3 path following folder reorganizing

### DIFF
--- a/examples/dataset/object_detection_2d/object_detection_2d/constants.py
+++ b/examples/dataset/object_detection_2d/object_detection_2d/constants.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 BUCKET = "kolena-public-examples"
 DATASET = "coco-2014-val"
+WORKFLOW = "transportation"
 ID_FIELDS = ["locator"]
 MODELS = [
     "yolo_r",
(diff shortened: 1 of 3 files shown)


examples/dataset/object_detection_2d/object_detection_2d/constants.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 BUCKET = "kolena-public-examples"
 DATASET = "coco-2014-val"
+WORKFLOW = "transportation"
```
ly-kolena: Should this be "object-detection-2d", in terms of breaking down by task type?

consoull: Maybe "workflow" is the wrong term here. The folders are `image-retrieval-by-text`, `person-detection`, `transportation`, so "task" may be better suited

ly-kolena: Asking more about should this we match the folder to the example, i.e. "object_detection_2d", or does "transportation" more reflective on the problem statement?

ly-kolena (approved):

PR #553, opened 2024-03-22
Make infer datatype more robust

ly-kolena:

### DIFF
--- a/kolena/dataset/dataset.py
+++ b/kolena/dataset/dataset.py
@@ -99,7 +99,11 @@ def _normalize_url(x: str) -> str:
 
 
 def _infer_datatype_value(x: str) -> str:
-    url = _normalize_url(x)
+    try:
+        url = _normalize_url(x or "")
+    except Exception:
+        return DatapointType.TABULAR.value
+
     mtype, _ = mimetypes.guess_type(url)
     if mtype:
         datatype = _get_datapoint_type(mtype)
--- a/kolena/dataset/dataset.py
+++ b/kolena/dataset/dataset.py
@@ -98,12 +98,11 @@ def _normalize_url(x: str) -> str:
     return url._replace(query="", fragment="").geturl()
 
 
-def _infer_datatype_value(x: str) -> str:
-    try:
-        url = _normalize_url(x or "")
-    except Exception:
+def _infer_datatype_value(x: Any) -> str:
+    if not isinstance(x, str):
         return DatapointType.TABULAR.value
 
+    url = _normalize_url(x or "")
     mtype, _ = mimetypes.guess_type(url)
     if mtype:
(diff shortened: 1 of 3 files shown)


dylangrandmont (approved):
Based on a possible scenario where multiple id_fields may be specified and within those combinations one of the fields may be empty (empty string or `None`), I think this change makes sense.

kolena/dataset/dataset.py
```diff
@@ -99,7 +99,11 @@ def _normalize_url(x: str) -> str:
 
 
 def _infer_datatype_value(x: str) -> str:
```
yifanwu-kolena: i think we need to update the type hint to reflect the change

kolena/dataset/dataset.py
```diff
@@ -99,7 +99,11 @@ def _normalize_url(x: str) -> str:
...
-    url = _normalize_url(x)
+    try:
+        url = _normalize_url(x or "")
+    except Exception:
+        return DatapointType.TABULAR.value
```
yifanwu-kolena: i think we can check the type of x, if not string just fallback to `DatapointType.TABULAR` to make the logic cleaner

tests/unit/dataset/test_dataset.py
```diff
@@ -50,6 +50,8 @@
         ("test.pcd", DatapointType.POINT_CLOUD),
         ("gcp://summary.pdf", DatapointType.DOCUMENT),
         ("//my.mp3", DatapointType.AUDIO),
+        (None, DatapointType.TABULAR),
+        (123, DatapointType.TABULAR),
     ],
 )
 def test__infer_datatype_value(uri: str, expected: str) -> None:
```
yifanwu-kolena: type hint

kolena/dataset/dataset.py
ly-kolena: Given that this is leveraging library, and we don't have all possible scenarios that could fail, I'm using try/exception over trying to if-condition all fail scenarios.

kolena/dataset/dataset.py
ly-kolena: With the "implied" permissiveness of DataFrame input could be heterogeneous, in reality we're handling `Any`.

kolena/dataset/dataset.py
yifanwu-kolena: if we use try-exception at here, assuming the type of x could be `Any`
- we will need to update `_normalize_url` to accept Any
- `urlparse` only accepts string, then we will do the type checking in `_normalize_url`

that's why i suggest to do the type checking at `_infer_datatype_value` level

ly-kolena: updated as suggested

yifanwu-kolena (approved):

PR #543, opened 2024-03-14
Add support for serializing/deserializing deeply nested DataObjects

consoull:
### Linked issue(s)
Fixes KOL-5377

### What change does this PR introduce and why?
Adds support for serializing deeply nested DataObjects

### Please check if the PR fulfills these requirements

- [x] Include reference to internal ticket and/or GitHub issue "Fixes #NNNN" (if applicable)
- [x] Relevant tests for the changes have been added
- [ ] Relevant docs have been added / updated

PR #555, opened 2024-03-25
fix: `dataframe_to_csv` only serialize list/dict to JSON

consoull:
### Linked issue(s)
KOL-5841

Fixes a bug in `dataframe_to_csv` where primitives where being dumped to JSON. Add guard to only dump lists/dicts

### DIFF
--- a/kolena/io.py
+++ b/kolena/io.py
@@ -42,7 +42,7 @@ def _deserialize_dataobject(x: Any) -> Any:
         return [_deserialize_dataobject(item) for item in x]
 
     if isinstance(x, dict):
-        if data_type := x.get(DATA_TYPE_FIELD):
+        if data_type := x.pop(DATA_TYPE_FIELD, None):
             if typed_dataobject := _DATA_TYPE_MAP.get(data_type):
                 return typed_dataobject._from_dict(x)
         else:
@@ -52,7 +52,7 @@ def _deserialize_dataobject(x: Any) -> Any:
 
 
 def _serialize_dataobject_str(x: Any) -> Any:
-    if isinstance(x, list) or isinstance(x, dict):
+    if isinstance(x, list) or isinstance(x, dict) or isinstance(x, DataObject):
         return json.dumps(x, cls=DataObjectJSONEncoder)
     return x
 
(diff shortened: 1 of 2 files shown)


kolena/io.py
```diff
@@ -52,7 +52,9 @@ def _deserialize_dataobject(x: Any) -> Any:
 
 
 def _serialize_dataobject_str(x: Any) -> Any:
-    return json.dumps(x, cls=DataObjectJSONEncoder)
+    if isinstance(x, list) or isinstance(x, dict):
```
ly-kolena: Can we add tests to prevent future regression?

consoull: Yes. I am working on it

consoull: Added a unit test for this method. After speaking with @yifanwu-kolena I'll be addressing the issue with `_deserialize_dataobject_str` calling `json.loads` for all strings when it should be more selective in a separate PR to limit the scope of this fix

kolena/io.py
```diff
@@ -42,7 +42,7 @@ def _deserialize_dataobject(x: Any) -> Any:
         return [_deserialize_dataobject(item) for item in x]
 
     if isinstance(x, dict):
-        if data_type := x.get(DATA_TYPE_FIELD):
+        if data_type := x.pop(DATA_TYPE_FIELD, None):
```
consoull, line 45: I realized this changed prior behaviour when looking back at https://github.com/kolenaIO/kolena/pull/543/files#diff-903d27b9445c08b2a9562ef9d804bc4c5f3ff9e06054752dc952144c7da2b000L45

kolena/io.py
```diff
@@ -52,7 +52,9 @@ def _deserialize_dataobject(x: Any) -> Any:
 
 
 def _serialize_dataobject_str(x: Any) -> Any:
-    return json.dumps(x, cls=DataObjectJSONEncoder)
+    if isinstance(x, list) or isinstance(x, dict) or isinstance(x, DataObject):
```
ly-kolena: nit: can be `isinstance(x, (list, dict, DataObject))`.

consoull: Thanks!

ly-kolena (approved):

PR #560, opened 2024-03-26
chunk uploads more aggressively

sandalns:
Explore reducing the BatchSize so that we do not reach the 5GB s3 limit.


kolena/_utils/consts.py
```diff
@@ -18,7 +18,7 @@
 
 class BatchSize(int, Enum):
     UPLOAD_CHIPS = 1_000
-    UPLOAD_RECORDS = 10_000_000
+    UPLOAD_RECORDS = 5_000_000
     UPLOAD_RESULTS = 1_000_000
```
munkyshi, line 22: Is `UPLOAD_RESULTS` still used? I thought it would be this variable, but I don't see it being used anywhere.

gordonhart: Looks like `UPLOAD_CHIPS` and `UPLOAD_RESULTS` are both unused.

PR #546, opened 2024-03-20
chore: Prioritize columns that begin with 'id_' or end with '_id' over others

munkyshi:
### Linked issue(s)
Fixes: KOL-4429

### What change does this PR introduce and why?
Prioritize columns that begin with 'id_' or end with '_id' over 'locator' and 'text'

We were seeing datasets with fields like 'image_id' default to using 'locator' as the ID field

### Please check if the PR fulfills these requirements

- [x] Include reference to internal ticket and/or GitHub issue "Fixes #NNNN" (if applicable)
- [x] Relevant tests for the changes have been added
- [ ] Relevant docs have been added / updated

PR #563, opened 2024-04-01
Invalidate python CircleCI cache to pull in mkdocs fix

consoull:
Invalidates python CircleCI cache to pull in mkdocs fix for google fonts API change https://github.com/squidfunk/mkdocs-material/issues/6983.

### DIFF
--- a/examples/dataset/semantic_segmentation/semantic_segmentation/upload_dataset.py
+++ b/examples/dataset/semantic_segmentation/semantic_segmentation/upload_dataset.py
@@ -25,7 +25,7 @@ from kolena.dataset import upload_dataset
 def run(args: Namespace) -> None:
     df_dataset = pd.read_csv(args.dataset_csv, storage_options={"anon": True}, converters={"captions": pd.eval})
     df_dataset["mask"] = df_dataset["mask"].apply(lambda mask: SegmentationMask(locator=mask, labels={1: "PERSON"}))
-    upload_dataset(args.dataset_name, df_dataset)
+    upload_dataset(args.dataset_name, df_dataset, id_fields=["locator"])
 
 
 def main() -> None:
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -53,7 +53,7 @@ jobs:
     steps:
       - checkout
       - restore_cache:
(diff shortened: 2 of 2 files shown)


examples/dataset/semantic_segmentation/semantic_segmentation/upload_dataset.py
```diff
@@ -25,7 +25,7 @@
 def run(args: Namespace) -> None:
     df_dataset = pd.read_csv(args.dataset_csv, storage_options={"anon": True}, converters={"captions": pd.eval})
     df_dataset["mask"] = df_dataset["mask"].apply(lambda mask: SegmentationMask(locator=mask, labels={1: "PERSON"}))
-    upload_dataset(args.dataset_name, df_dataset)
+    upload_dataset(args.dataset_name, df_dataset, id_fields=["locator"])
```
consoull, line 28: Needed to update to fix a separate breakage caused by https://github.com/kolenaIO/kolena/pull/546

gordonhart: Before we land this — should we update the behavior to prefer `locator` when present?

consoull: I would rather scope this change down to fixing the CI. Changing the `id_field` inference order is not backwards compatible
cc: @munkyshi

gordonhart: Agreed, but from my perspective, we already broke it by changing it to prefer `*_id` or `id_*` over `locator`.

munkyshi: Ugh, very true -- I hadn't really considered the breakage, sorry.

IMO preferring `*_id` is still the ideal end behavior (I'd prefer to go with an explicitly specified ID over assuming the locator uniquely qualifies a datapoint). What's the ID field that's taking precedence over `locator` in this case?

munkyshi: Oh, I see -- the results are keyed by `locator` so even though `image_id` is unique, it doesn't work for results resolution.

IMO I'd rather be explicit about `locator` -- open to counterarguments

munkyshi (approved):

PR #565, opened 2024-04-03
test run to chunk dataframe based on exported file size

sandalns:
### Linked issue(s)
Fixes KOL-5876
Updates TestRun to chunk dataframes for inferences, test_sample_metrics and aggregate_metrics. Previously we were uploading the entire data frame. This also batches by creating a temporary parquet file with 1000 rows and using that to determine the number of rows to produce a file that is 250 MB.


kolena/workflow/test_run.py
```diff
@@ -242,7 +245,10 @@ def upload_inferences(self, inferences: List[Tuple[TestSample, Inference]]) -> N
         df_serializable = df_validated.as_serializable()
 
         init_response = init_upload()
-        upload_data_frame_chunk(df_serializable, init_response.uuid)
+        batch_size = (
+            len(df_serializable) // MB_BATCH_LIMIT // (get_preflight_export_size(df_serializable) // MB_CONVERSION)
+        )
+        upload_data_frame(df_serializable, batch_size, init_response.uuid)
```
yifanwu-kolena: could we create a wrapper func for this logic? it should be able to reuse in the future as well

sandalns: yes doing that and adding tests

kolena/_utils/batched_load.py
```diff
@@ -79,6 +82,18 @@ def upload_data_frame_chunk(df_chunk: pd.DataFrame, load_uuid: str) -> None:
 DFType = TypeVar("DFType", bound=LoadableDataFrame)
 
 
+def upload_smart_chunk_data_frame(df: pd.DataFrame, uuid: str, rows: int = 1000) -> None:
```
ly-kolena: One idea is in `upload_data_frame_chunk`, where we are already doing the `to_parquet` and would have actual compressed size from `BytesIO` buffer, to split if the chunk size exceeds limit.

yifanwu-kolena: i think we should have similar naming like `upload_data_frame`, maybe `upload_data_frame_in_smart_chunks`. we might want to use `load_uuid` as args as well for consistency.

it will be great to add some comments for this func

kolena/_utils/batched_load.py
```diff
@@ -79,6 +82,18 @@ def upload_data_frame_chunk(df_chunk: pd.DataFrame, load_uuid: str) -> None:
...
+    batch_size = (BATCH_LIMIT // get_preflight_export_size(df, rows)) * rows
+    upload_data_frame(df, batch_size, uuid)
+
+
+def get_preflight_export_size(df: pd.DataFrame, rows: int) -> int:
```
yifanwu-kolena: could we add `_` prefix for this func? b/c it should be only used in this file

kolena/_utils/batched_load.py
```diff
@@ -79,6 +82,18 @@ def upload_data_frame_chunk(df_chunk: pd.DataFrame, load_uuid: str) -> None:
...
+    df_subset = df[:rows]
+    with tempfile.NamedTemporaryFile() as temp:
+        df_subset.to_parquet(temp.name)
+        return os.stat(temp.name).st_size
```
yifanwu-kolena: could we add some defensive programming to raise error when the size is 0?

i understand that it should never be 0 in practice, but from the programming perspective it's possible to be 0

sandalns: will update

kolena/_utils/batched_load.py
sandalns: will add

kolena/_utils/batched_load.py
sandalns: will do

kolena/_utils/batched_load.py
```diff
@@ -79,6 +82,18 @@ def upload_data_frame_chunk(df_chunk: pd.DataFrame, load_uuid: str) -> None:
...
 
+def upload_smart_chunk_data_frame(df: pd.DataFrame, uuid: str, rows: int = 1000) -> None:
+    batch_size = (BATCH_LIMIT // get_preflight_export_size(df, rows)) * rows
+    upload_data_frame(df, batch_size, uuid)
+
+
+def get_preflight_export_size(df: pd.DataFrame, rows: int) -> int:
+    df_subset = df[:rows]
+    with tempfile.NamedTemporaryFile() as temp:
+        df_subset.to_parquet(temp.name)
```
consoull: This seems pretty heavy handed. Won't this be quite slow for large datasets? Can we get away with using `df.memory_usage(deep=True)` as a proxy for the serialized size?

kolena/_utils/batched_load.py
```diff
@@ -38,6 +39,8 @@
 
 VALIDATION_COUNT_LIMIT = 100
 STAGE_STATUS__LOADED = "LOADED"
+MB_CONVERSION = 1024 * 1024
```
consoull: nit: pow(1024, 2)

kolena/_utils/batched_load.py
yifanwu-kolena: i think the memory usage of a dataframe will be inflated comparing with the file size, and the inflated ratio is varied depends on the data inside the df, so it's hard to figure out the magic number of the ratio to make sure the parquet file size in the sweet spot: 100MB-250MB.

current way will not be fully accurate due to parquet internal data encodings overhead, but i think it will be more accurate than memory usage

kolena/_utils/batched_load.py
```diff
@@ -79,6 +82,18 @@ def upload_data_frame_chunk(df_chunk: pd.DataFrame, load_uuid: str) -> None:
 DFType = TypeVar("DFType", bound=LoadableDataFrame)
 
 
+def upload_smart_chunk_data_frame(df: pd.DataFrame, uuid: str, rows: int = 1000) -> None:
+    batch_size = (BATCH_LIMIT // get_preflight_export_size(df, rows)) * rows
```
yifanwu-kolena: could we add a short circuit logic to just use the len(df) as batch_size when the memory usage of the df is relative small(like 2GB) to handle [this comment](https://github.com/kolenaIO/kolena/pull/565/files#r1553990628)?

kolena/_utils/batched_load.py
sandalns: Going to take a look at that and see how it works.

kolena/_utils/batched_load.py
sandalns: done

kolena/_utils/batched_load.py
sandalns: done

kolena/_utils/batched_load.py
```diff
@@ -79,6 +82,26 @@ def upload_data_frame_chunk(df_chunk: pd.DataFrame, load_uuid: str) -> None:
...
+        len(df)
+        if df_memory <= SHORT_CIRCUT_LIMIT and len(df) > 0
+        else (BATCH_LIMIT // _get_preflight_export_size(df, rows)) * rows
+    )
+    upload_data_frame(df, batch_size, uuid)
+
+
+def _get_preflight_export_size(df: pd.DataFrame, rows: int) -> int:
+    df_subset = df[:rows]
+    with tempfile.NamedTemporaryFile() as temp:
```
ly-kolena: I think we can use `io.ByteIO` here, not incurring disk penalty

sandalns: done

kolena/workflow/test_run.py
```diff
@@ -242,8 +242,7 @@ def upload_inferences(self, inferences: List[Tuple[TestSample, Inference]]) -> N
         df_serializable = df_validated.as_serializable()
 
         init_response = init_upload()
-        upload_data_frame_chunk(df_serializable, init_response.uuid)
-
+        upload_data_frame_in_smart_chunks(df_serializable, init_response.uuid)
```
ly-kolena: I think the `upload_data_frame_chunk` in these places should have used `upload_data_frame` in the first place. For this PR, we should be updating occurences of `upload_data_frame` with the "new method".

I would suggest renaming current `upload_data_frame` to internal method, and `upload_data_frame_in_smart_chunks` to `upload_data_frame`. (note: signature changed tho).

cc @gordonhart for any concerns.

sandalns: Will update

kolena/_utils/batched_load.py
```diff
@@ -79,6 +81,26 @@ def upload_data_frame_chunk(df_chunk: pd.DataFrame, load_uuid: str) -> None:
 DFType = TypeVar("DFType", bound=LoadableDataFrame)
 
 
+def upload_data_frame(df: pd.DataFrame, load_uuid: str, rows: int = 1000) -> None:
+    df_memory = df.memory_usage(index=True, deep=True).sum()
+    batch_size = (
```
ly-kolena, line 99: Can we add unit test to verify the chunking logic, maybe with some mock and reduced size?

sandalns: Will add

sandalns: added unit tests

ly-kolena (approved):

PR #573, opened 2024-04-15
Use projected bounding boxes in KITTI example

bagalaster:
### Linked issue(s)

References KOL-6029

Adds world-to-pixel projection matrices to images to allow projected 3d annotations to be rendered on images in Kolena.

### DIFF
--- a/examples/dataset/object_detection_3d/object_detection_3d/upload_dataset.py
+++ b/examples/dataset/object_detection_3d/object_detection_3d/upload_dataset.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 from argparse import ArgumentParser
 from argparse import Namespace
-from collections import Counter
 
 import pandas as pd
 from object_detection_3d.constants import BUCKET
@@ -21,13 +20,9 @@ from object_detection_3d.constants import DATASET
 from object_detection_3d.constants import DEFAULT_DATASET_NAME
 from object_detection_3d.constants import ID_FIELDS
 from object_detection_3d.constants import TASK
-from object_detection_3d.utils import create_velo_to_pixel_matrix
+from object_detection_3d.utils import load_data
 
-from kolena.annotation import LabeledBoundingBox
-from kolena.annotation import LabeledBoundingBox3D
-from kolena.asset import ImageAsset
 from kolena.dataset import upload_dataset
-from kolena.workflow.asset import PointCloudAsset
 
 # KITTI only supports evaluation of the first three classes but "Van" and "Person_sitting" GTs
 # are used to avoid penalizing inferences labeled as "Car" and "Pedestrian" respectively.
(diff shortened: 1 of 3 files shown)


gordonhart:
Rendering both 2D and 3D bboxes on the image is quite noisy. What's necessary to rip out the 2D ground truth bounding boxes entirely?

Screenshot 2024-04-15 at 12 22 15 PM

bagalaster:
I agree, I'm trying to get rid of these before this is marked ready for review - there is [logic](https://github.com/kolenaIO/kolena/blob/trunk/examples/dataset/object_detection_3d/object_detection_3d/vendored/kitti_eval.py#L51) in the buried fairly deep in the evaluation that assumes that there are 2d bounding box ground truths. 

Long term, I think we should apply a more general approach to evaluation for 3D OD, similar to the method we have for 2D OD today. This approach should only rely on 3D OD inferences, which would allow us to completely rip out 2D annotations.

Short-term, as part of this PR, I am planning on omitting the 2D annotations from the data uploaded in `upload_dataset` and re-ingesting the raw 2D annotations for use in evaluation in `upload_results` (as opposed to downloading them via `download_dataset` as we currently do).

I also considered modifying the code inside of `kitti_eval` to not depend on 2D annotations, but I think this would be more trouble than it's worth, especially if we already think we need different evaluation logic anyways.

examples/dataset/object_detection_3d/object_detection_3d/upload_dataset.py
```diff
@@ -47,61 +43,6 @@
 }
 
 
-def load_data(df_raw: pd.DataFrame) -> pd.DataFrame:
```
bagalaster, line 50: Moved to `utils` as this is re-used in `upload_results`

examples/dataset/object_detection_3d/object_detection_3d/upload_results.py
```diff
@@ -292,16 +290,22 @@ def compute_metrics_with_difficulty(difficulty: int) -> pd.DataFrame:
...
+                    nMismatchedInferences=len(FP_3D),
                     thresholds=current_optimal_thresholds,
-                    FP_2D=FP_2D,
                     FP_3D=FP_3D,
-                    TP_2D=TP_2D,
                     TP_3D=TP_3D,
-                    FN_2D=FN_2D,
                     FN_3D=FN_3D,
+                    images_with_inferences=[
+                        ImageAsset(
```
bagalaster, line 301: A couple of things to consider
- Should we include (un)matched_inferences and unmatched_ground_truth on these images too? These would not be automatically hidden as they are at the top level and may clutter the image.
- This leads to one field on the dataset with just ground truths projected on all images and one field on the model result with both ground truths and model inferences projected on all images. Is there a different approach to consider?

examples/dataset/object_detection_3d/object_detection_3d/utils.py
```diff
@@ -97,3 +104,85 @@ def limit_period(val: np.array, offset: Optional[float] = 0.5, period: Optional[
...
+                "image_id": record.image_id,
+                "locator": record.left_image,
+                "image_bboxes": bboxes_2d,
+                "images": [
+                    ImageAsset(
+                        locator=record.left_image,
+                        side="left",  # type: ignore[call-arg]
+                        # type: ignore[call-arg]
+                        projection=create_velo_to_pixel_matrix(record.Tr_velo_to_cam, record.P2),
+                        velodyne_bboxes=bboxes_3d,  # type: ignore[call-arg]
```
munkyshi, line 162: Do we need to attach the 3d bboxes to the image assets? I thought they'd come from the parent level?

examples/dataset/object_detection_3d/object_detection_3d/upload_results.py
munkyshi: Synced up on slack -- seems like the right approach for this longer term is to get the top level 3d bboxes projected into the 2d image assets (if they have a projection matrix). This looks to be non-trivial, so for the kitti example (for now), we'll have the left/right view visible only for model results (and include gts on those images).

bagalaster: These changes^ are made now

munkyshi (approved):

PR #577, opened 2024-04-22
Support thresholded metrics

ly-kolena:

### DIFF
--- a/kolena/dataset/evaluation.py
+++ b/kolena/dataset/evaluation.py
@@ -251,6 +251,7 @@ def upload_results(
     :param model: The name of the model.
     :param results: Either a DataFrame or a list of tuples, where each tuple consists of
                     an eval configuration and a DataFrame.
+    :param thresholded_fields: Columns in result DataFrame containing data associated with different thresholds.
 
     :return: None
     """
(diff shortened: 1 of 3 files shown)


kolena/dataset/evaluation.py
```diff
@@ -238,4 +254,4 @@ def upload_results(
 
     :return: None
     """
-    _upload_results(dataset, model, results)
+    _upload_results(dataset, model, results, thresholded_fields=thresholded_fields)
```
jared-jewitt, line 258: I believe you'll also want to update the doc comment to include the new optional `thresholded_fields` param

ly-kolena: Added docstring.

kolena/dataset/evaluation.py
```diff
@@ -227,6 +242,7 @@ def upload_results(
     dataset: str,
     model: str,
     results: Union[DataFrame, List[Tuple[EvalConfig, DataFrame]]],
+    thresholded_fields: Optional[List[str]] = None,
```
munkyshi, line 245: Haven't been paying the closest attention here, but is there any reason we can't infer that a field is thresholded based on the dataframe? Having another field here increases the complexity of our API.

ly-kolena: 1. There could be multiple fields that store thresholded metrics kind of data.
2. For users using thresholded metrics, they are already in the "more advanced" scenario who have to have the knowledge, thus specifying the fields is comparatively not much extra burden.
3. Personally, I prefer to avoid inferring based on naming convention.

However, we do face the same issue regardless when it comes to importing through UI. Currently thinking to support configuration with this infer from UI side.

munkyshi: Synced with @ly-kolena and Mehrad offline -- being explicit with the name in the public client, and having the web app do the "smart inference" seems like a fine approach. No change needed here.

munkyshi (approved):

PR #580, opened 2024-04-24
Added support for upload JSONL and update documentation

diegokolena:
### Linked issue(s)
- KOL-5568

### DIFF
--- a/kolena/io.py
+++ b/kolena/io.py
@@ -145,7 +145,8 @@ def dataframe_from_jsonl(*args: Any, **kwargs: Any) -> pd.DataFrame:
     :return: DataFrame.
     """
     df = pd.read_json(*args, **kwargs, lines=True)
-    return df
+    df_post = _dataframe_object_serde(df, _deserialize_dataobject)
+    return df_post
 
 
 def dataframe_to_jsonl(df: pd.DataFrame, *args: Any, **kwargs: Any) -> str:
--- a/kolena/io.py
+++ b/kolena/io.py
@@ -135,34 +135,6 @@ def dataframe_from_json(*args: Any, **kwargs: Any) -> pd.DataFrame:
     return df_post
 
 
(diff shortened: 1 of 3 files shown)


kolena/io.py
```diff
@@ -135,6 +135,34 @@ def dataframe_from_json(*args: Any, **kwargs: Any) -> pd.DataFrame:
...
+    :param args: positional arguments for `pandas.DataFrame.read_json`.
+    :param kwargs: keyword arguments for `pandas.DataFrame.read_json`.
+    :return: DataFrame.
+    """
+    df = pd.read_json(*args, **kwargs, lines=True)
+    df_post = _dataframe_object_serde(df, _deserialize_dataobject)
+    return df_post
+
+
+def dataframe_to_jsonl(df: pd.DataFrame, *args: Any, **kwargs: Any) -> str:
```
ly-kolena: Do we need this? Users should be knowledgeable to just call `df.to_json(lines=True)`.

kolena/io.py
```diff
@@ -135,6 +135,34 @@ def dataframe_from_json(*args: Any, **kwargs: Any) -> pd.DataFrame:
     return df_post
 
 
+def dataframe_from_jsonl(*args: Any, **kwargs: Any) -> pd.DataFrame:
```
ly-kolena: Same question, users should be able to just use existing `dataframe_from_json(<file>, lines=True)`.

kolena/io.py
diegokolena: I agree it's not necessary, but I thought we could provide this helper function since we already have one for dataframe_to_csv, among others.

ly-kolena: I'm in favor of not addition new helpers unless there's something existing ones cannot do. Thoughts?

diegokolena: Ok to remove, if users find necessary we can bring in the future.

ly-kolena (approved):

PR #581, opened 2024-04-24
Update object-detection to thresholded metrics

ly-kolena:

### DIFF
--- a/examples/dataset/object_detection_3d/object_detection_3d/utils.py
+++ b/examples/dataset/object_detection_3d/object_detection_3d/utils.py
@@ -12,7 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from collections import Counter
+from typing import Any
+from typing import Dict
 from typing import List
+from typing import NamedTuple
 from typing import Optional
 from typing import Tuple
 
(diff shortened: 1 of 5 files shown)


examples/dataset/object_detection_3d/object_detection_3d/utils.py
```diff
@@ -181,3 +184,65 @@ def load_data(df_raw: pd.DataFrame) -> pd.DataFrame:
         )
 
     return pd.DataFrame(records)
+
+
+def _compute_thresholded_metrics(
```
bagalaster, line 189: What's the reason for re-implementing these evaluation functions here? Could we leverage `multiclass_datapoint_metrics` from `kolena/_experimental/object_detection/dataset.py` instead?

ly-kolena: The underlying data structures are not the same. for object-detection 2d, we're using our own implementation and `MatchInfereces` etc. classes, but for object-detection 3d, we're using mmdetection3d.

munkyshi (approved):
Read through the code a couple of times -- this all seems reasonable to me. I'd like to run through an example with this tomorrow, but I don't want to block merging this PR on carving out the time to do that.

jared-jewitt:
Should we be updating this section on the docs? https://docs.kolena.com/dataset/formatting-your-datasets/#formatting-results-for-object-detection

ly-kolena:
> Should we be updating this section on the docs? https://docs.kolena.com/dataset/formatting-your-datasets/#formatting-results-for-object-detection

Any objection to follow up in a separate PR? We can discuss how to present this feature and where to put there.

PR #584, opened 2024-04-30
Make BoundingBox derived fields more resilient

munkyshi:
### Linked issue(s)

Fixes: KOL-6104

Users have accidentally swapped top_left and bottom_right points, or input top_right and bottom_left as the bbox fields.

This change makes our derived fields (width, height, area, aspect ratio) resilient to these types of errors -- the calculation still works as long as we are provided with opposite corners.

### DIFF
--- a/kolena/annotation.py
+++ b/kolena/annotation.py
@@ -91,8 +91,8 @@ class BoundingBox(Annotation):
         return _AnnotationType.BOUNDING_BOX
 
     def __post_init__(self) -> None:
-        object.__setattr__(self, "width", self.bottom_right[0] - self.top_left[0])
-        object.__setattr__(self, "height", self.bottom_right[1] - self.top_left[1])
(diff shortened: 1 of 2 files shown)


kolena/annotation.py
```diff
@@ -91,8 +91,8 @@ def _data_type() -> _AnnotationType:
         return _AnnotationType.BOUNDING_BOX
 
     def __post_init__(self) -> None:
-        object.__setattr__(self, "width", self.bottom_right[0] - self.top_left[0])
```
ly-kolena, line 94: what features on our platform does this break? If not, I think it's better to show the data error then hide them.

munkyshi: If it's an actual data error, it should still be noticeable in the UI -- this just makes us robust to a easy to make mistake for bbox input where the resolution should be straightforward. We've had users run into this, and I'd prefer to avoid explaining that they need to re-push their data to fix something like this.

ly-kolena (approved):

PR #586, opened 2024-05-02
created new annotation for labeled text segments

grant-Kolena:
to support [this](https://linear.app/kolena/issue/KOL-6082/ability-to-highlight-relevant-text-in-text-files-for-rag-evaluation)

I have created this sample dataset for the color text segment [label](https://trunk.kolena.io/grant-tenant/dataset/studio?datasetId=220&offset=0)


kolena/annotation.py
```diff
@@ -382,6 +387,123 @@ class ScoredLabeledTimeSegment(TimeSegment):
...
+    """
+    Segments of text in the associated text field
+    """
+
+    text_field: str
+    """Text field column name containing the text segments."""
+
+    segments: List[Tuple[int, int]]
+    """List of text segments. The first item of each tuple is the start index of the segment,
+    and the second item is the end index."""
```
gordonhart: Why is this a list?

kolena/annotation.py
```diff
@@ -382,6 +387,123 @@ class ScoredLabeledTimeSegment(TimeSegment):
...
+    """
+
+    text_segments: List[TextSegment]
+    """List of text segments."""
+
+    label: str
+    """The label for the specified segments."""
+
+    color: Optional[str] = None
+    """Color of the segments for display."""
```
gordonhart: Optional values with defaults break extending this dataclass with your own custom fields — better to omit.

kolena/annotation.py
```diff
@@ -382,6 +387,123 @@ class ScoredLabeledTimeSegment(TimeSegment):
...
+            for field_name, text in texts.items():
+                segment = find_substring_indices(field_name, text, keywords)
+                if segment:
+                    segments.append(segment)
+            if segments:
+                labeled_segments.append(LabeledTextSegments(segments, label, colors.get(label, None)))
+        return labeled_segments
+
+    @staticmethod
+    def extract_labeled_text_segments_from_keywords(
```
gordonhart: Odd to have this attached as a static method to this annotation class. Should this be defined elsewhere?

kolena/annotation.py
```diff
@@ -382,6 +387,123 @@ class ScoredLabeledTimeSegment(TimeSegment):
     """The score associated with this time segment."""
 
 
+@dataclass(frozen=True, config=ValidatorConfig)
+class TextSegment(DataObject):
```
gordonhart: Should extend `Annotation`

kolena/annotation.py
```diff
@@ -382,6 +387,123 @@ class ScoredLabeledTimeSegment(TimeSegment):
...
+        for start, end in self.segments:
+            if start < 0:
+                raise ValueError(f"Start index must be non-negative ({start} provided)")
+            if end < 0:
+                raise ValueError(f"End index must be non-negative ({end} provided)")
+            if start > end:
+                raise ValueError(f"Start index must be less than or equal to end index ({start} > {end})")
+
+
+def find_substring_indices(field_name: str, text: str, substrings: list[str]) -> Optional[TextSegment]:
```
gordonhart: This shouldn't be publicly exposed

kolena/annotation.py
```diff
@@ -382,6 +387,123 @@ class ScoredLabeledTimeSegment(TimeSegment):
...
+
+        indices.extend([(match.start(), match.end()) for match in matches])
+
+    if indices:
+        return TextSegment(field_name, indices)
+    return None
+
+
+@dataclass(frozen=True, config=ValidatorConfig)
+class LabeledTextSegments(Annotation):
```
gordonhart: Should extend `TextSegments`

kolena/annotation.py
```diff
@@ -382,6 +387,123 @@ class ScoredLabeledTimeSegment(TimeSegment):
     """The score associated with this time segment."""
 
 
+@dataclass(frozen=True, config=ValidatorConfig)
```
gordonhart, line 386: `_data_type` needs to be defined here

kolena/annotation.py
```diff
@@ -55,6 +59,7 @@ class _AnnotationType(DataType):
     BITMAP_MASK = "BITMAP_MASK"
     LABEL = "LABEL"
     TIME_SEGMENT = "TIME_SEGMENT"
+    LabeledTextSegments = "LABELED_TEXT_SEGMENTS"
```
gordonhart: 1. Remove `Labeled`
2. Enum value should be capitalized, `TEXT_SEGMENTS`

kolena/annotation.py
grant-Kolena: We are going for a structure like this, top level is label, the next level is the segments for a specific field, there is a [RFC](https://www.notion.so/kolena/Highlight-Relevant-Test-Fields-b9481e261d6c4c058f66bec328c715ae?pvs=4) for this, and @GoodLuckJane believes this makes the FE work a lot easier

```
[
  {
    "text_segments": [
      {
        "text_field": "text",
        "segments": [
          [
            214,
            217
          ],
          [
            724,
            727
          ]
        ]
      },
      {
        "text_field": "text_summary",
        "segments": [
          [
            72,
            75
          ],
          [
            39,
            41
          ]
        ]
      }
    ],
    "label": "pronoun",
    "color": "red",
    "data_type": "ANNOTATION/LABELED_TEXT_SEGMENTS"
  }
]
```

kolena/annotation.py
grant-Kolena: got it, will remove the optional field. Colors will be mandatory then

kolena/annotation.py
grant-Kolena: Sure I can move it to util or something

kolena/annotation.py
grant-Kolena: I am actually following the example for [Curve](https://github.com/kolenaIO/kolena/blob/a72f0933e5c3a4c63db67dac1856bacc1ce7616b/kolena/workflow/plot.py#L75) and [CurvePlot](https://github.com/kolenaIO/kolena/blob/a72f0933e5c3a4c63db67dac1856bacc1ce7616b/kolena/workflow/plot.py#L110), Curve is the building block for CurvePlot, and Curve extended DataObject while CurvePlot extended Plot. TextSegments will never be used on its own I believe

kolena/annotation.py
grant-Kolena: my bad

kolena/annotation.py
grant-Kolena: same reasoning as above, Curve is never used alone so no `_data_type` is defined

kolena/annotation.py
grant-Kolena: Do you think this is the right pattern?

kolena/annotation.py
gordonhart: Let's let the frontend figure out colors.

kolena/annotation.py
```diff
@@ -55,6 +57,7 @@ class _AnnotationType(DataType):
     BITMAP_MASK = "BITMAP_MASK"
     LABEL = "LABEL"
     TIME_SEGMENT = "TIME_SEGMENT"
+    TEXT_SEGMENTS = "TEXT_SEGMENTS"
```
gordonhart: Should be singular (`TEXT_SEGMENT`) now that the annotation is simplified.

kolena/annotation.py
```diff
@@ -382,6 +385,127 @@ class ScoredLabeledTimeSegment(TimeSegment):
...
+
+    """
+    df[labeled_text_segments_column] = df[text_fields].apply(
+        lambda texts: _extract_labeled_text_segments_from_keywords_single_row(
+            texts.to_dict(),
+            keyword_labels,
+        ),
+        axis=1,
+    )
+    return df
```
gordonhart: I would move these elsewhere — they don't really belong in `annotation.py`.

kolena/annotation.py
grant-Kolena: got it

kolena/utils.py
```diff
@@ -0,0 +1,80 @@
...
+            labeled_segments.extend(_find_substring_indices(field_name, text, keywords, label))
+    return labeled_segments
+
+
+def extract_labeled_text_segments_from_keywords(
+    df: pd.DataFrame,
+    text_fields: List[str],
+    keyword_labels: Dict[str, List[str]],
+    labeled_text_segments_column: str = "labeled_text_segments",
+) -> pd.DataFrame:
```
gordonhart: What's the reason for including this function along with this PR?

grant-Kolena: I am trying to provide a simpler way for the user to use the feature, instead of calculate the indexes and transform into our format, just provide us with keywords and tell us what label we should use. I think it could be useful when the user just want to highlight some keywords. 

This is part of the requirement as well
```
Setup a utility function to allow for string match use case
```

munkyshi: Discussion continued in https://linear.app/kolena/issue/KOL-6082/ability-to-highlight-relevant-text-in-text-files-for-rag-evaluation#comment-b2e6f63f

kolena/annotation.py
```diff
@@ -382,6 +383,67 @@ class ScoredLabeledTimeSegment(TimeSegment):
...
+    end: int
+    """End index of the text segment."""
+
+    def __post_init__(self) -> None:
+        if self.start < 0:
+            raise ValueError(f"Start index must be non-negative ({self.start} provided)")
+        if self.end < 0:
+            raise ValueError(f"End index must be non-negative ({self.end} provided)")
+        if self.start > self.end:
+            raise ValueError(f"Start index must be less than or equal to end index ({self.start} > {self.end})")
```
ly-kolena: what does it mean to have `start == end`?

grant-Kolena: Zero length segment, probably not useful as well

ly-kolena (approved):

gordonhart (approved):

PR #588, opened 2024-05-03
Add IoU property to object detection inferences

bagalaster:
### Linked issue(s)

Fixes KOL-6202

This PR attaches the best IoU found over the ground truths to the inferences in the calculation of object detection results. This enables users to differentiate between different kinds of false positives (i.e. localization errors vs background detection).

Additionally, this PR exposes new functions `iter_object_detection_results` and `create_object_detection_results` to allow users to generate object detection results without uploading them. This is expected to be useful for users who would like to be able to post-process their object detection results before upload.

### DIFF
--- a/kolena/_experimental/object_detection/__init__.py
+++ b/kolena/_experimental/object_detection/__init__.py
@@ -29,7 +29,7 @@ from .workflow import ThresholdConfiguration
 
 from .evaluator import ObjectDetectionEvaluator
 from .dataset import upload_object_detection_results
-from .dataset import create_object_detection_results
+from .dataset import iter_object_detection_results
 
 __all__ = [
     "TestSample",
@@ -46,6 +46,6 @@ __all__ = [
     "TestSuiteMetrics",
     "ThresholdConfiguration",
     "ObjectDetectionEvaluator",
-    "create_object_detection_results",
+    "iter_object_detection_results",
     "upload_object_detection_results",
 ]
--- a/kolena/_experimental/object_detection/__init__.py
+++ b/kolena/_experimental/object_detection/__init__.py
@@ -29,6 +29,7 @@ from .workflow import ThresholdConfiguration
 
 from .evaluator import ObjectDetectionEvaluator
 from .dataset import upload_object_detection_results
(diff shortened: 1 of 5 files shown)


kolena/_experimental/object_detection/__init__.py
```diff
@@ -45,5 +47,7 @@
     "TestSuiteMetrics",
     "ThresholdConfiguration",
     "ObjectDetectionEvaluator",
+    "iter_object_detection_results",
+    "create_object_detection_results",
```
munkyshi: Are these needed as part of this IoU PR? I'm a bit concerned about expanding the client surface area with these `iter` and `create` functions (especially `create`, since it seems to be just a surface level wrapper around `iter`).

kolena/metrics/_geometry.py
```diff
@@ -141,22 +149,25 @@ def _match_inferences_single_class_pascal_voc(
...
             inf_gt_iou = iou(gt, inf)
-            # track the highest IoU over the threshold
-            if inf_gt_iou >= iou_threshold and inf_gt_iou > best_gt_iou:
-                best_gt_iou = inf_gt_iou
-                best_gt = g
-
-        if best_gt is None or (best_gt in taken_gts and best_gt < len(ground_truths)):
+            if gt not in (ignored_ground_truths or []):
+                best_gt_iou = max(best_gt_iou, inf_gt_iou)
+                # track the highest IoU over the threshold
```
munkyshi: nit: looks like this got extra indented?

munkyshi:
Code change looks good! Mostly want to make sure we land on what we expose as public API before we merge this.

kolena/_experimental/object_detection/__init__.py
bagalaster: Discussed offline, will expose `compute_object_detection_results` returning a `DataFrame` and keep `_iter_object_detection_results` returning an `Iterator[DataFrame]` as a private function for consistency with how we do this elsewhere (e.g. `download_dataset`)

munkyshi (approved):

PR #589, opened 2024-05-06
feat: Add quality-standard result download

consoull:
### Linked issue(s)
Fixes KOL-5814

Adds a function to download the quality-standard result given a dataset and list of models tested on that dataset. The result is a multi-index DataFrame with index `(stratificaiton, test_case)` and columns `(model, eval_config, metric_group, metric)`.


consoull:
‎quality_standard_diagram_light ‎001
An example of the returned DataFrame. Would like to include this in the documentation somewhere. Unfortunatley it doesn't look like you can embed an image in the API docs with https://mkdocstrings.github.io/griffe/. Would need to create a separate page

consoull:
Thinking I can look to update https://docs.kolena.com/dataset/core-concepts/quality-standard/ with the diagram as a follow up

tests/integration/_experimental/test_quality_standard.py
```diff
@@ -0,0 +1,100 @@
...
+    df_columns: pd.MultiIndex = quality_standard_df.columns
+    assert df_columns.names == ["model", "eval_config", "metric_group", "metric"]
+    assert df_columns.levels == [[model_name], ["null"], [metric_group_name], [metric_name]]
+
+    df_index: pd.MultiIndex = quality_standard_df.index
+    assert df_index.names == ["stratification", "test_case"]
+    assert all(df_index.levels[0] == ["Dataset", test_case_name])
+    assert all(df_index.levels[1] == ["new york", "waterloo"])
+
+    assert quality_standard_df.loc[("Dataset", np.nan), (model_name, "null", metric_group_name, metric_name)] == 0.0
```
yifanwu-kolena: is string "null" expected? how's this look like for a non-null eval config?

consoull: A json blob. I'll update the test to add a second model with an eval config

consoull: Pandas is stringifying the null here, but yes it is expected to be an empty cell if there is no eval config

yifanwu-kolena (approved):

PR #594, opened 2024-05-10
Remove telemetry

kevinguo455:
### Linked issue(s)
https://linear.app/kolena/issue/KOL-6307/remove-telemetry-from-python-sdk

Removes client telemetry. We do not get value out of this and it adds complexity to Python client.

### DIFF
--- a/kolena/_utils/state.py
+++ b/kolena/_utils/state.py
@@ -24,9 +24,8 @@ from typing import Optional
 
 import requests
 
-import kolena
 import kolena._api.v1.token as API
-from kolena._utils import krequests
+import kolena._utils
 from kolena._utils.serde import from_dict
 from kolena.errors import InvalidTokenError
 from kolena.errors import UnauthenticatedError
@@ -147,7 +146,7 @@ def get_token(
         proxies=proxies,
     )
     try:
-        krequests.raise_for_status(r)
+        kolena._utils.krequests.raise_for_status(r)
     except UnauthenticatedError as e:
         raise InvalidTokenError(e)
 
--- a/kolena/_utils/state.py
+++ b/kolena/_utils/state.py
@@ -24,8 +24,8 @@ from typing import Optional
 
 import requests
 
+import kolena
 import kolena._api.v1.token as API
-import kolena._utils
 from kolena._utils.serde import from_dict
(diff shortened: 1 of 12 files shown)


tests/unit/test_krequests.py
```diff
@@ -51,10 +50,10 @@ def clean_client_state() -> Iterator[None]:
         ({"timeout": (1, 60)}, None, None),  # attempt to override default args
         ({}, {"http": "dummy-proxy"}, {}),
         ({}, {}, {"some-header-key": "some-header-val"}),
-        ({}, {}, {"X-Kolena-Telemetry": "off"}),  # attempt to override default headers with client state
-        ({"headers": {"X-Kolena-Telemetry": "off"}}, {}, {}),  # attempt to override default headers with kwargs
+        ({}, {}, {}),  # attempt to override default headers with client state
```
ly-kolena: Let's update these cases with new values to preserve the original purposes.

kolena/_utils/state.py
```diff
@@ -24,9 +24,8 @@
 
 import requests
 
-import kolena
```
ly-kolena, line 27: What's the reason for this change? There are a couple usages in this file with `kolena.xxx`. Let's revert the change if it is not needed.

kevinguo455: The pre-commit hooks removed this line. Now that I look at it, I'm not entirely sure why the `kolena.xxx` usages still work with `import kolena` removed.

ly-kolena (approved):

kolena/_utils/state.py
```diff
@@ -153,7 +146,7 @@ def get_token(
         proxies=proxies,
     )
     try:
-        krequests.raise_for_status(r)
+        kolena._utils.krequests.raise_for_status(r)
```
yifanwu-kolena, line 149: this [commit](https://github.com/kolenaIO/kolena/pull/594/commits/79a5d59adc77b1b2c8a53e4fba7dd77f4e22abf3) againsts with here, might need to revert this diff and add back `from kolena._utils import krequests` as before to keep the explicit importing aligning with the usage in this file

yifanwu-kolena: nvm, my comment is invalid

PR #597, opened 2024-05-17
feat: better format dataset result download response

nankolena:
### Linked issue(s)
Resolves KOL-5941

Formatting the `download_results()` response better. Instead of having to rely on indexing the output such as:
```
response = download_results(dataset, model)
datapoints = response[0]
eval_config = response[1][0][0]
results = response[1][0][1]
```

We can now do:
```
response = download_results(dataset, model)
datapoints = response[0]
eval_config = response[1][0].eval_config
results = response[1][0].results
```

This change is backward compatible with the existing pattern of regular tuples, and the same change is applied to `upload_results()`

Updated API ref page:
image

### DIFF
--- a/kolena/dataset/evaluation.py
+++ b/kolena/dataset/evaluation.py
@@ -17,6 +17,7 @@ from typing import Any
 from typing import Dict
 from typing import Iterator
 from typing import List
+from typing import NamedTuple
 from typing import Optional
 from typing import Tuple
 from typing import Union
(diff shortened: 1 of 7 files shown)


kolena/dataset/evaluation.py
```diff
@@ -148,10 +159,15 @@ def download_results(
 
     :param dataset: The name of the dataset.
     :param model: The name of the model.
-    :return: Tuple of DataFrame of datapoints and list of tuples,
-             each containing an evaluation configuration and the corresponding DataFrame of results.
+    :return: Tuple of DataFrame of datapoints and list of [`EvalConfigResults`][kolena.dataset.EvalConfigResults].
     """
     log.info(f"downloading results for model '{model}' on dataset '{dataset}'")
+    existing_dataset = _load_dataset_metadata(dataset)
```
ly-kolena, line 164: non-blocking: wonder if we want to add a flag like `raise_not_exists` to this internal method to reduce code dupes.

ly-kolena (approved):

yifanwu-kolena (approved):

kolena/dataset/evaluation.py
nankolena: Refactored in https://github.com/kolenaIO/kolena/pull/597/commits/246e01bfd3f28c0e3f1be454c613fbfe3a6ec3e5. Let me know what you think. I initially tried to add a boolean flag to the func arg, but we'd still have code dup given that this function returns `Optional[EntityData]`, meaning all usages would still need to do check if the returned val is None.

There's only one place that actually did not need to raise the not found error, so I instead made a `try...except` block there but keep the not found check in this function itself.

PR #601, opened 2024-05-24
Assign lower rank to all archived pages in docs

marko-polo-cheno:
Improve documentation searching to prioritize datasets dev guide and API reference.


gordonhart (approved):
Good addition, thanks @marko-polo-cheno.

PR #610, opened 2024-05-30
Preserve inference properties when matching

bagalaster:
### Linked issue(s)

Related to KOL-6304

Preserves additional properties added when adding `iou` to inference matches produced by `match_inferences`.

`dataclasses.replace` only preserves properties which exist in the class definition, but not any extra properties that may be allowed (via `extra='allow'` in the pydantic validation configuration).

This PR ensures that inference annotations retain any additional properties after being matched.

### DIFF
--- a/kolena/metrics/_geometry.py
+++ b/kolena/metrics/_geometry.py
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import dataclasses
 from collections import defaultdict
 from typing import Dict
 from typing import Generic
@@ -98,8 +97,9 @@ Inf = TypeVar("Inf", bound=Union[ScoredBoundingBox, ScoredPolygon, ScoredLabeled
 
 
 def _inf_with_iou(inf: Inf, iou_val: float) -> Inf:
-    args_without_iou = {key: value for key, value in inf._to_dict().items() if key != "iou"}
-    return dataclasses.replace(inf, **args_without_iou, iou=iou_val)  # type: ignore[call-arg]
+    args_without_iou = {key: value for key, value in inf._to_dict().items() if key not in ["iou", "data_type"]}
+    out: Inf = inf.__class__(**args_without_iou, iou=iou_val)  # type: ignore[call-arg,assignment]
+    return out
 
 
 @dataclass(frozen=True)
--- a/kolena/metrics/_geometry.py
+++ b/kolena/metrics/_geometry.py
@@ -98,7 +98,8 @@ Inf = TypeVar("Inf", bound=Union[ScoredBoundingBox, ScoredPolygon, ScoredLabeled
 
 
 def _inf_with_iou(inf: Inf, iou_val: float) -> Inf:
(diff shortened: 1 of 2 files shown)


kolena/metrics/_geometry.py
```diff
@@ -98,7 +97,13 @@ def as_shapely_polygon(obj: Union[BoundingBox, Polygon]) -> ShapelyPolygon:
 
 
 def _inf_with_iou(inf: Inf, iou_val: float) -> Inf:
-    return dataclasses.replace(inf, iou=iou_val)  # type: ignore[call-arg]
+    input_args = {
+        key: value
+        for key, value in inf._to_dict().items()
+        if key not in ["width", "height", "area", "aspect_ratio", "iou", "data_type"]
+    }
+    out: Inf = inf.__class__(**input_args, iou=iou_val)  # type: ignore[call-arg,assignment]
```
ly-kolena: Any issue with just using `inf.__class__(**inf._to_dict(), iou=iou_val)`?

tests/unit/metrics/test_geometry.py
```diff
@@ -689,6 +689,30 @@ def test__match_inferences(
         assert inf.iou == pytest.approx(expected_iou, abs=1e-5)
 
 
+def test__match_inferences__preserves_extra_props_tp() -> None:
```
ly-kolena: nit: would simply testing `_inf_with_iou` cover these cases?

tests/unit/metrics/test_geometry.py
```diff
@@ -689,6 +689,30 @@ def test__match_inferences(
...
+    ground_truths = [BoundingBox(top_left=(1, 1), bottom_right=(6, 6))]
+    inferences = [
+        # type: ignore[call-arg]
+        ScoredLabeledBoundingBox(score=0.9, label="cow", top_left=(1, 1), bottom_right=(6, 6), flavor="cherry"),
+    ]
+    matches = match_inferences(ground_truths, inferences)
+    matched_inf = matches.matched[0][1]
+    assert hasattr(matched_inf, "flavor")
```
ly-kolena: nit: this should be already covered by the next assertion

kolena/metrics/_geometry.py
bagalaster: We will still have to filter out `iou` since it may already be defined on the input inference and cause an error. `data_type` would also become exposed as a property, when it ordinarily wouldn't be.

No real issue with passing the other properties - just seemed a bit strange to pass them explicitly when they're going to be derived from the bbox coords anyways.

ly-kolena: My concern is hard coding fields here, and this might lead to unintentional drift. How about this

```
def _inf_with_iou(inf: Inf, iou_val: float) -> Inf:
    exclude = ["iou"] + [f.name for f in dataclasses.fields(inf) if not f.init]
    obj = {k: v for k, v in vars(inf).items() if k not in exclude}
    out = inf.__class__(**obj, iou=iou_val)
    return out
```

side note: not impacting the use case here, but `_to_dict` is mainly used for serialization, so technically, if a member is also a dataobject, the returned value would also be a serialized version instead of the original object.

consoull (approved):

ly-kolena (approved):

PR #609, opened 2024-05-30
Added descriptions to metrics glossary pages

marko-polo-cheno:
### Linked issue(s)
Fixes KOL-6605
### What change does this PR introduce and why?
Added descriptions to metrics glossary pages

PR #612, opened 2024-05-31
Add docs for correlation metrics

brian386:

Adds documentation for Pearson's Correlation Coefficient, and Spearman's Correlation Coefficient.


y27choi (commented):
These pages look really great, @brian386.  👍

A couple of questions I have are:
1. Does "monotonic relationship" need any further explanation? I'm thinking a small callout next to the first paragraph of `metrics/spearman-correlation` could be nice.
2. As I was reading both pages, I assumed that the example data were exactly the same but it seems like one of the predicted temperatures is different. Was that intentional? I think it would be nice to have a matching example so you can kind of compare how these metrics can differ.

brian386:
> These pages look really great, @brian386. 👍
> 
> A couple of questions I have are:
> 
> 1. Does "monotonic relationship" need any further explanation? I'm thinking a small callout next to the first paragraph of `metrics/spearman-correlation` could be nice.
> 2. As I was reading both pages, I assumed that the example data were exactly the same but it seems like one of the predicted temperatures is different. Was that intentional? I think it would be nice to have a matching example so you can kind of compare how these metrics can differ.

Thanks! For 1, I explain monotonic relationship in the second part of the sentence: "where as one variable increases, the other also increases." Should this be explained further or worded differently? For 2, I changed the numbers to match - I wasn't paying too much attention to that before.

docs/metrics/pearson-correlation.md
```diff
@@ -0,0 +1,54 @@
+# Pearson Correlation Coefficient
```
gordonhart: Add a `description` to the front matter please — see other metrics guides or #609 for examples.

docs/metrics/spearman-correlation.md
```diff
@@ -0,0 +1,58 @@
+# Spearman's Rank Correlation Coefficient
```
gordonhart: Same comment about `description`.

docs/metrics/spearman-correlation.md
```diff
@@ -0,0 +1,58 @@
...
+Spearman's rank correlation coefficient is a statistical measure that assesses the strength and direction of the
+monotonic relationship between two variables. Unlike Pearson's correlation, Spearman's correlation does not assume a
+linear relationship and can be used with ordinal data. The values range from -1 to +1. A value close to +1 indicates a
+strong positive monotonic correlation, where as one variable increases, the other also increases. A value close to -1
+signifies a strong negative monotonic correlation, meaning as one variable increases, the other decreases. A value of
+0 indicates no monotonic correlation between the variables.
+
+## Implementation Details
+
+The Spearman's rank correlation coefficient is calculated by ranking the data points and then applying Pearson's
```
gordonhart: > Pearson's correlation

Add hyperlink to the other doc?

docs/metrics/index.md
```diff
@@ -86,6 +86,13 @@ biases, and its intended uses.
     emphasizing larger errors more significantly than smaller ones. It is mainly different from
     [MAE](mean-absolute-error.md) in that larger errors are penalized more heavily.
 
+- [Pearson Correlation Coefficient](pearson-correlation.md)
```
gordonhart: Do these page titles need "Coefficient" or can that be omitted? Looking at "Cohen's Kappa" which is the closest analog in the other metrics, it seems that the more standard thing to do would be to just call these "Pearson Correlation" and "Spearman's Rank Correlation".

docs/metrics/index.md
```diff
@@ -86,6 +86,13 @@ biases, and its intended uses.
...
+
+    ---
+
+    Pearson's correlation coefficient is a statistical measure that evaluates the strength and
```
gordonhart: We should be consistent — is it "Pearson Correlation" or "Pearson's Correlation"?

docs/metrics/pearson-correlation.md
```diff
@@ -0,0 +1,54 @@
...
+\bar{x} &= \frac{25 + 35 + 30 + 40}{4} = 32.5 \\[1em]
+\bar{y} &= \frac{30 + 28 + 27 + 35}{4} = 30 \\[1em]
+r &= \frac{\sum_{i=1}^{4} (x_i - \bar{x})(y_i - \bar{y})}{\sqrt{\sum_{i=1}^{4} (x_i - \bar{x})^2 \sum_{i=1}^{4} (y_i -
+\bar{y})^2}} \\[1em]
+&= \frac{(25 - 32.5)(30 - 30) + (35 - 32.5)(28 - 30) + (30 - 32.5)(27 - 30) +
+(40 - 32.5)(35 - 30)}{\sqrt{[(25 - 32.5)^2 +
+(35 - 32.5)^2 + (30 - 32.5)^2 + (40 - 32.5)^2] [(30 - 30)^2 + (28 - 30)^2 + (27 - 30)^2 + (35 - 30)^2]}} \\[1em]
+&\approx 0.58
+\end{align}
+$$
```
gordonhart, line 49: This page should link somewhere to Spearman's Rank Correlation to indicate that they are related and that you'd often use both at once.

docs/metrics/pearson-correlation.md
```diff
@@ -0,0 +1,54 @@
...
+Temperature Correlation:
+
+<div class="grid" markdown>
+
+| Ground Truth Temperature (&deg;C) | Predicted Temperature (&deg;C) |
+| --- | --- |
+| 25  | 30  |
+| 35  | 28  |
+| 30  | 27  |
+| 40  | 35  |
```
gordonhart: I'd recommend removing one of these pairs to make the sample calculations easier to read.

gordonhart: Same comment for the other doc too.

docs/metrics/spearman-correlation.md
```diff
@@ -0,0 +1,58 @@
+# Spearman's Rank Correlation Coefficient
+
+Spearman's rank correlation coefficient is a statistical measure that assesses the strength and direction of the
+monotonic relationship between two variables. Unlike Pearson's correlation, Spearman's correlation does not assume a
```
gordonhart: `Pearson's Correlation` should be linked to other doc

docs/metrics/index.md
ly-kolena: quick google looks like `Pearson Correlation` is used by wiki and scipy.

docs/metrics/pearson-correlation.md
```diff
@@ -0,0 +1,56 @@
...
+
+$$
+\begin{align}
+\bar{x} &= \frac{25 + 35 + 30}{3} = 30 \\[1em]
+\bar{y} &= \frac{30 + 28 + 27}{3} = 28.33 \\[1em]
+r &= \frac{\sum_{i=1}^{4} (x_i - \bar{x})(y_i - \bar{y})}{\sqrt{\sum_{i=1}^{3} (x_i - \bar{x})^2 \sum_{i=1}^{3} (y_i -
+\bar{y})^2}} \\[1em]
+&= \frac{(25 - 30)(30 - 28.33) + (35 - 30)(28 - 28.33) + (30 - 30)(27 - 28.33)}{\sqrt{[(25 - 30)^2 +
+(35 - 30)^2 + (30 - 30)^2] [(30 - 28.33)^2 + (28 - 28.33)^2 + (27 - 28.33)^2 ]}} \\[1em]
+&\approx -0.65
```
gordonhart: Should we use an example with a better model, i.e. one that produces a positive correlation? Same for Spearman's

gordonhart (approved):
Looks good — one remaining point, otherwise ready to ship. Well done @brian386

docs/metrics/pearson-correlation.md
brian386: Oh yeah, I was also thinking about this - I think temperatures vs. predicted temperatures is not the best example, since correlation metrics are used to measure correlation between 2 variables, not necessarily predicted vs ground truth (we expect the correlation to always be positive so it's not so meaningful). For example, we might calculate the correlation between the person's IQ and their height to see if there is correlation. But I couldn't think of any realistic example with relatively small numbers. Maybe I could rename the columns to "Temperature in place A" and "Temperature in place B"?

y27choi: I think what @gordonhart meant here is to update the predicted temperature so it will yield a positive correlation. Regarding the column renaming, I prefer to keep it "temperatures vs. predicted temperatures" as this encourages people to use this correlation coefficient as an evaluation metric.

brian386: Sounds good, I changed it now

y27choi (approved):

PR #615, opened 2024-06-06
Ignore ground truths with ignore flag set when matching

bagalaster:
### Linked issue(s)

Towards KOL-6692

Ignores bounding boxes with `ignore=True` when matching for OD. This allows users to set ignored ground truths while still being able to rely on `upload_object_detection_results` to do the work of calculating TP/FP/FN


kolena/_experimental/object_detection/dataset.py
```diff
@@ -212,6 +213,7 @@ def _compute_metrics(
     *,
     ground_truth: str,
     inference: str,
+    ignore_gt_property: Optional[str] = None,
```
munkyshi: Is there a reason to make this configurable, rather than just using `ignore`? Wonder if it makes the API needlessly verbose.

bagalaster: Transparency would be the biggest argument in favor IMO. This makes it clear to the user what field is being used. I could easily see a situation where user uses the name `ignored` instead of `ignore`, for example. In that case, hard-coding the magic property `ignore` would make it pretty difficult to discover the issue.

bagalaster: We could alternatively have an `options: Optional[Dict]` argument or similar, and fetch this name from `options["ignore_gt_property"]`. I think I actually like that better than what I have here, come to think of it.

kolena/_experimental/object_detection/dataset.py
```diff
@@ -240,10 +242,20 @@ def _compute_metrics(
...
         inferences = record[idx[inference]]
+        ignored_ground_truths = [
+            gt
+            for gt in ground_truths
+            if ignore_gt_property is not None
+            and hasattr(gt, ignore_gt_property)
+            and isinstance(getattr(gt, ignore_gt_property), bool)
+            and getattr(gt, ignore_gt_property)
+        ]
+        unignored_ground_truths = [gt for gt in ground_truths if gt not in ignored_ground_truths]
```
munkyshi, line 255: I'm not sure this is quite what we're targeting here -- my read is that ignored bboxes should still be matched, but not surface as FNs if unmatched. Maybe something we should clarify before committing to.

kolena/_experimental/object_detection/dataset.py
munkyshi: We should add a docstring for this param

kolena/_experimental/object_detection/dataset.py
bagalaster: That's the behavior that this accomplishes. `_match_inferences_single_class_pascal_voc` [matches against](https://github.com/kolenaIO/kolena/blob/ea38bdce1e71facc3bf783d5a14b6bc85ad9dc32/kolena/metrics/_geometry.py#L147) `ground_truths + ignored_ground_truths`, but [doesn't include](https://github.com/kolenaIO/kolena/blob/ea38bdce1e71facc3bf783d5a14b6bc85ad9dc32/kolena/metrics/_geometry.py#L176) `ignored_ground_truths` in the FN and doesn't include any of the inferences that match with anything in `ignored_ground_truths` in the FP.

bagalaster: This is also verified in the unit test

munkyshi: Ah, great (thank you)

munkyshi: Ah, great (thank you)

munkyshi (approved):

PR #617, opened 2024-06-13
Fix treatment of single class vs multiclass thresholded objects

bagalaster:
### Linked issue(s)

Fixes KOL-6703

Fixes an edge case in `upload_object_detection_results` where all ground truths and inferences have the same label.

### DIFF
--- a/kolena/_experimental/object_detection/dataset.py
+++ b/kolena/_experimental/object_detection/dataset.py
@@ -150,14 +150,21 @@ def multiclass_datapoint_metrics(
     ]
     scores = [inf["score"] for inf in tp] + [inf.score for inf in fp]
     labels = sorted(
+        {_safe_get_label(inf) for _, inf in object_matches.matched}
+        .union(
+            {_safe_get_label(inf) for inf in object_matches.unmatched_inf},
+        )
+        .union(
+            {_safe_get_label(gt) for gt, _ in object_matches.unmatched_gt},
+        )
+        .difference({None}),
+    )
+    inference_labels = (
         {inf.label for _, inf in object_matches.matched}
         .union(
-            {inf.label for inf in object_matches.unmatched_inf},
+            {_safe_get_label(inf) for inf in object_matches.unmatched_inf},
         )
-        .union({gt.label for gt, _ in object_matches.unmatched_gt}),
-    )
-    inference_labels = {inf.label for _, inf in object_matches.matched}.union(
-        {inf.label for inf in object_matches.unmatched_inf},
+        .difference({None})
     )
     fields = [
         ScoredLabel(label=label, score=thresholds[label])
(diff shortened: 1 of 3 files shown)


kolena/_experimental/object_detection/dataset.py
```diff
@@ -305,14 +321,21 @@ def _compute_metrics(
...
-        return len(labels) >= 2
-    except AttributeError:
-        return False
+    labels = {_safe_get_label(gt) for gt in itertools.chain.from_iterable(_filter_null(ground_truth))}.union(
+        {_safe_get_label(inf) for inf in itertools.chain.from_iterable(_filter_null(inference))},
+    )
+    if None in labels:
+        return len(labels) >= 3
+
+    return len(labels) >= 2
```
bagalaster, line 338: This is now more robust to the case where some objects in the ground truths/inferences are unlabeled. Previously, any unlabeled objects would cause us to treat this as a single class problem.

kolena/_experimental/object_detection/dataset.py
```diff
@@ -305,15 +311,28 @@ def _compute_metrics(
...
-        return len(labels) >= 2
-    except AttributeError:
+    labels = {_safe_get_label(gt) for gt in itertools.chain.from_iterable(_filter_null(ground_truth))}.union(
+        {_safe_get_label(inf) for inf in itertools.chain.from_iterable(_filter_null(inference))},
+    )
+    if None in labels:
         return False
 
+    return len(labels) >= 2
+
```
bagalaster, line 337: I personally think this logic is wrong. I think we should ignore objects without labels and check whether there is more than 1 unique label, instead of just giving up and defaulting to single class when there is an object without a label.

I kept the existing behavior for now because I'm not sure what consequences changing this behavior has for downstream handling of thresholded metrics. These are only style tweaks that make changing the behavior marginally easier in the future.

munkyshi: I don't think there is a case where we mix unlabeled vs labeled annotations for OD, which is why it short circuits like this -- I agree that we should ideally check all valid labels, rather than declare that it's a single class problem.

tests/unit/_experimental/object_detection/test_dataset.py
```diff
@@ -23,59 +23,154 @@
...
+            # inferences
+            [
+                [LabeledBoundingBox(label="cat", top_left=[3, 3], bottom_right=[9, 9])],
+                [BoundingBox(top_left=[11, 10], bottom_right=[15, 15])],
+                [],
+                math.nan,
+                None,
+            ],
+            # expected
+            False,
```
munkyshi, line 115: Is this the short circuit issue that you flagged above?

munkyshi (approved):

tests/unit/_experimental/object_detection/test_dataset.py
bagalaster: Correct

PR #583, opened 2024-04-26
How to do Object Detection in Kolena - Task Doc Template 

marko-polo-cheno:
### Linked issue(s)

Fixes KOL-5750

### What change does this PR introduce and why?
The beginning of a series of How-to guides in Kolena. This one is for object detection.

Will need thoughts on document organization. From this [thread](https://linear.app/kolena/issue/KOL-6038/clean-up-docskolenacom#comment-d9a5c219), I think all future `How-to`'s would live under the quickstart page. `Quickstart > How to test Object Detection models`

PR #643, opened 2024-07-11
Extend OD matching functions to accept an optional required_match_fields parameter

munkyshi:
### Linked issue(s)
Fixes: KOL-6947

Extend OD matching functions to accept an optional required_match_fields parameter

This allows for restricting matches to situations when the ground truth and inference agree on a field.
One example use case for this is with video data, where the ground truth and inference should agree
on a `frame_id` for them to be considered a proper match.

### DIFF
--- a/kolena/_experimental/object_detection/dataset.py
+++ b/kolena/_experimental/object_detection/dataset.py
@@ -226,6 +226,7 @@ def _compute_metrics(
     threshold_strategy: Union[Literal["F1-Optimal"], float, Dict[str, float]] = 0.5,
     min_confidence_score: float = 0.5,
     batch_size: int = 10_000,
+    required_match_fields: Optional[List[str]] = None,
 ) -> Iterator[pd.DataFrame]:
     """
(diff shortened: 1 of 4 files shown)


munkyshi:
One additional update that I'd like to do afterwards is to provide a simplified example for `pedestrian_crossing_detection` that uses the `upload_object_detection_results` function, rather than its convoluted matching logic.

munkyshi:
I'd also like to eventually update the OD docs (initially https://github.com/kolenaIO/kolena/pull/583) to have a blurb about using this with `frame_id` for the video use case.

munkyshi:
The more I think about this, the more it may make sense to build out a separate `upload_video_object_detection_results` -- this still loses out on computing frame level metrics (as well as aggregating them into cross frame level precision / recall / fpr / f1), which could be useful. Those implementations might make sense as a follow-up PR.

bagalaster (approved):

yifanwu-kolena (approved):

PR #659, opened 2024-08-09
Add skeleton fields to Keypoints and Keypoints3D annotations

munkyshi:
### Linked issue(s)
N/A (related KOL-7143)

Add skeleton fields to Keypoints and Keypoints3D annotations

This was a previously undocumented feature, where users could specify a `skeleton` and have keypoints rendered with connecting lines in the web application. This PR documents this and makes it an official part of the `kolena` API.


kolena/annotation.py
```diff
@@ -190,6 +191,9 @@ class Keypoints(Annotation):
     points: List[Tuple[float, float]]
     """The sequence of discrete `(x, y)` pixel coordinates comprising this keypoints annotation."""
 
+    skeleton: Optional[List[Tuple[float, float]]] = None
+    """A sequence of paired indices indicating connections between `points`."""
```
gordonhart: The reason these were left off initially is because dataclasses with optional fields don't play nicely with subclassing. This change will break any usages like this:

```py
from dataclasses import dataclass
from kolena.annotation import Keypoints

@dataclass(frozen=True)
class MyKeypoints(Keypoints):
    my_field: str
```

The above script runs on the latest release, but on this branch fails with:

```
TypeError: non-default argument 'my_field' follows default argument
```

gordonhart (approved):

kolena/annotation.py
```diff
@@ -20,7 +20,9 @@
 | [`BoundingBox3D`][kolena.annotation.BoundingBox3D] | Limited to `PointCloud` data |
 | [`Polygon`][kolena.annotation.Polygon] | Limited to `Image` or `Video` data |
 | [`Polyline`][kolena.annotation.Polyline] | Limited to `Image` or `Video` data |
+| [`Polyline3D`][kolena.annotation.Polyline3D] | Limited to `PointCloud` data |
 | [`Keypoints`][kolena.annotation.Keypoints] | Limited to `Image` or `Video` data |
+| [`Keypoints3D`][kolena.annotation.Keypoints3D] | Limited to `PointCloud` data |
```
gordonhart: Can these be visualized without a point cloud?

PR #662, opened 2024-08-13
give the user an option to choose whether to use the intersection or union of datapoint when calculating metrics with quality standard

grant-Kolena:
This is to address Mehrad's comment [here](https://linear.app/kolena/issue/KOL-6996/[kodiak]-[voxel]-ability-to-see-qs-metrics-for-the-portion-of-the#comment-5e5a74df), we want the user to select whether to use the intersection or union of datapoint when calculating metrics with quality standard.

### DIFF
--- a/kolena/_experimental/quality_standard.py
+++ b/kolena/_experimental/quality_standard.py
@@ -48,11 +48,29 @@ def _format_quality_standard_result_df(quality_standard_result: dict) -> pd.Data
     return df.pivot(columns=["model", "eval_config", "metric_group", "metric"], values="value")
 
 
+def _download_quality_standard_result(
+    dataset: str,
+    models: List[str],
+    metric_groups: Union[List[str], None] = None,
+    intersect_results: bool = True,
+) -> pd.DataFrame:
+    response = krequests.get(
+        Path.RESULT,
+        params=dict(dataset_name=dataset, models=models, metric_groups=metric_groups),
+        api_version="v2",
+    )
+    krequests.raise_for_status(response)
+    return _format_quality_standard_result_df(
+        response.json(),
+    )
+
+
 @with_event(event_name=EventAPI.Event.FETCH_QUALITY_STANDARD_RESULT)
 def download_quality_standard_result(
     dataset: str,
     models: List[str],
     metric_groups: Union[List[str], None] = None,
+    intersect_results: bool = True,
 ) -> pd.DataFrame:
     """
     Download quality standard result given a dataset and list of models.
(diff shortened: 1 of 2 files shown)


kolena/_experimental/quality_standard.py
```diff
@@ -48,27 +48,46 @@ def _format_quality_standard_result_df(quality_standard_result: dict) -> pd.Data
...
-    )
-    krequests.raise_for_status(response)
-    return _format_quality_standard_result_df(response.json())
+    if intersect_results:
+        return _download_quality_standard_result(dataset, models, metric_groups, intersect_results)
+    else:
+        result_dfs = []
+        for model in models:
+            result_dfs.append(_download_quality_standard_result(dataset, [model], metric_groups, intersect_results))
+        return pd.concat(result_dfs, axis=1)
```
yifanwu-kolena, line 93: why not use `exclude_null_results` via api but handle at the client level?

grant81: For the qs request from client, we always want to exclude_null_results, this is just to choose whether the user want to only calculate metrics from datapoints that has result for all models. This is to have parity with the future FE change

grant-Kolena: This is also me BTW

tests/integration/_experimental/test_quality_standard.py
```diff
@@ -104,6 +111,11 @@ def test__download_quality_standard_result(
         quality_standard_df,
         download_quality_standard_result(dataset_name, [model_name], [metric_group_name]),
     )
+    assert_frame_equal(
+        download_quality_standard_result(dataset_name, [model_name, model_name2]),
+        download_quality_standard_result(dataset_name, [model_name, model_name2], intersect_results=False),
+        check_like=True,
+    )
```
yifanwu-kolena: if we don't upload partial results for model2, i think this test is not that usefully

grant-Kolena: ok, let me just separate the tests

tests/integration/_experimental/test_quality_standard.py
```diff
@@ -164,3 +164,115 @@ def test__download_quality_standard_result(
...
+        dataset_name,
+        model_name,
+        model1_results,
+    )
+    model_2_results = model1_results[2:]
+    _upload_results(
+        dataset_name,
+        model_name2,
+        model_2_results,
+    )
```
yifanwu-kolena: could we normalize the affix style between `model_name`, `model_name2`, `model1_results` and `model_2_results`?

yifanwu-kolena (approved):

PR #665, opened 2024-08-19
fix: validate dataset is non-empty when uploading

nankolena:
### Linked issue(s)
Towards [KOL-7149](https://linear.app/kolena/issue/KOL-7149/improve-the-sdk-experience-with-dataset-upload), will also make a server-side change

To maintain parity with the frontend which does not allow creation of dataset that is empty of has empty names. Currently, empty datasets created via the SDK will not show in the frontend and users are not able to delete them.

### DIFF
--- a/kolena/dataset/_common.py
+++ b/kolena/dataset/_common.py
@@ -65,6 +65,11 @@ def _validate_dataframe_ids_uniqueness(df: pd.DataFrame, id_fields: List[str]) -
         raise InputValidationError("id fields must be hashable") from e
 
 
+def validate_dataframe_not_empty(df: pd.DataFrame) -> None:
+    if df.empty:
+        raise InputValidationError("dataframe is empty")
+
+
 def validate_dataframe_ids(df: pd.DataFrame, id_fields: List[str]) -> None:
     for id_field in id_fields:
         if id_field not in df.columns:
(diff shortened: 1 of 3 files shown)


kolena/dataset/_common.py
```diff
@@ -65,6 +65,16 @@ def _validate_dataframe_ids_uniqueness(df: pd.DataFrame, id_fields: List[str]) -
         raise InputValidationError("id fields must be hashable") from e
 
 
+def validate_dataframe_not_empty(df: pd.DataFrame) -> None:
+    if df.empty:
+        raise InputValidationError("dataframe is empty")
+
+
+def validate_name_not_empty(name: str) -> None:
+    if not name.replace(" ", ""):
```
yifanwu-kolena: i think we should use [`strip`](https://docs.python.org/3.10/library/stdtypes.html#str.strip) at here

yifanwu-kolena: or we can add the normalization and validation in the sdk-server as well

nankolena: True, `strip()` is simpler - changing in a sec. I'm gonna make a server side change too but wanted to change client side first

yifanwu-kolena (approved):

PR #667, opened 2024-08-20
feat: add support for 3d mesh assets + add optional `frame_rate` to video assets

nankolena:
### Linked issue(s)
Resolves KOL-7209
Resolves KOL-7208

- Add support for 3d mesh asset type ~~(depends on https://github.com/kolenaIO/marina/pull/3960)~~
- Add `frame_rate` for video assets (related to https://github.com/kolenaIO/marina/pull/3953)

Verified documentation generated:
Aug-20-2024 12-32-35

### DIFF
--- a/kolena/asset.py
+++ b/kolena/asset.py
@@ -23,7 +23,7 @@ The following asset types are available:
 - [`PointCloudAsset`][kolena.asset.PointCloudAsset]
 - [`VideoAsset`][kolena.asset.VideoAsset]
 - [`AudioAsset`][kolena.asset.AudioAsset]
-- [`Mesh3dAsset`][kolena.asset.Mesh3dAsset]
+- [`MeshAsset`][kolena.asset.MeshAsset]
 
 """
 from abc import ABCMeta
@@ -43,7 +43,7 @@ class _AssetType(DataType):
     POINT_CLOUD = "POINT_CLOUD"
     VIDEO = "VIDEO"
     AUDIO = "AUDIO"
-    MESH_3D = "MESH_3D"
+    MESH = "MESH"
 
     @staticmethod
     def _data_category() -> DataCategory:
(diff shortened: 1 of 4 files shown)


kolena/asset.py
```diff
@@ -157,4 +164,29 @@ def _data_type() -> _AssetType:
         return _AssetType.AUDIO
 
 
-_ASSET_TYPES = [ImageAsset, PlainTextAsset, BinaryAsset, PointCloudAsset, BaseVideoAsset, VideoAsset, AudioAsset]
+@dataclass(frozen=True, config=ValidatorConfig)
+class Mesh3dAsset(Asset):
```
gordonhart: Any reason not to call this `MeshAsset`?

nankolena: This is because the current implementation is specific to 3d mesh types, and there might be a possibility that we'd want to introduce support for 2d meshes, in which case we would want to stay consistent with [annotation naming](https://github.com/kolenaIO/kolena/blob/87a2f2d2843083c79db240b1e7cb9efc0df099f4/kolena/annotation.py#L48), for example `POLYLINE` vs `POLYLINE_3D`. I've raised https://github.com/kolenaIO/marina/pull/3960 to update UI logic.

gordonhart: What is a 2D mesh?

nankolena: There could be different applications such as
- 2D finite element analysis: https://www.comsol.com/multiphysics/mesh-refinement
- Creative assets such as in Unity 2D: https://docs.unity3d.com/ScriptReference/Collider2D.CreateMesh.html

I'm coming from a more defensive angle in case we do find an use case for 2d mesh in the future, in which case the naming would otherwise be confusing if we have `MESH` for 3d and `MESH_2D` for 2d.

gordonhart (approved):

PR #682, opened 2024-09-13
do not run embedding reduction when running integration test

grant-Kolena:

### DIFF
--- a/tests/integration/_experimental/search/test_embeddings.py
+++ b/tests/integration/_experimental/search/test_embeddings.py
@@ -17,8 +17,8 @@ import numpy as np
 import pandas as pd
 import pytest
 
-from kolena._experimental.search import upload_dataset_embeddings
 from kolena._experimental.search import upload_embeddings
+from kolena._experimental.search.embeddings import _upload_dataset_embeddings
 from kolena.dataset import upload_dataset
 from kolena.errors import InputValidationError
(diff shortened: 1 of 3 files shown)


tests/integration/_experimental/search/test_embeddings.py
```diff
@@ -100,17 +100,18 @@ def dataset_name() -> str:
     ],
 )
 def test__upload_dataset_embeddings(embedding: np.ndarray, dataset_name: str) -> None:
-    upload_dataset_embeddings(
+    _upload_dataset_embeddings(
```
nankolena, line 103: Are we testing this internal method because we do not want to surface the `run_embedding_reduction` arg publicly yet?

grant-Kolena: this is causing issues due to different embedding length was uploaded for the same dataset. Also we don't want to trigger an aws job unnecessarily every time an integration test runs.

nankolena: Yup, I'm just wondering why we are testing this internal function instead of the public one, and it seems like they are identical except that the internal one can now take this extra arg that we do not yet want to surface publicly?

nankolena (approved):

yifanwu-kolena (approved):

PR #683, opened 2024-09-15
Migrate to uv for package management

munkyshi:
### Linked issue(s)

KOL-7307

Replace `poetry` with `uv` for dependency management and tooling.

Installing `kolena` is potentially a long process, and the dependency resolution time with `poetry` has crept up to the point of causing our workflow tests to flake. `uv` seems to offer a much faster interface for this, while preserving simplicity of development.

Also removes codecov for now, as it's something that hasn't been actively used / monitored.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -38,17 +38,16 @@ jobs:
         with:
           ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
 
-      - name: Install Poetry
-        run: pip install poetry==1.5.1
+      - uses: astral-sh/setup-uv@v2
+
+      - name: Setup environment
+        run: uv venv && source .venv/bin/activate
 
       - name: Fetch tags to enable autoversioning
         run: git fetch --prune --unshallow --tags
 
-      - name: Update package version to PEP 440-compliant production release tag
-        run: poetry version $(git describe --tags --abbrev=0)
-
-      - name: Install Poetry dependencies
-        run: poetry install
+      - name: Install dependencies
+        run: uv pip install --all-extras -r pyproject.toml
 
       - name: Install docs 'insiders' dependencies
         run: ./docs/setup_insiders.sh
@@ -56,7 +55,7 @@ jobs:
       - name: Publish 'kolena' documentation to trunk (S3)
         if: inputs.environment != 'production'
         run: |
-          poetry run mkdocs build --verbose --strict --config-file mkdocs.insiders.yml
+          uv run mkdocs build --verbose --strict --config-file mkdocs.insiders.yml
           aws s3 sync ./site "s3://trunk-docs.kolena.io" --delete
           aws cloudfront create-invalidation --no-cli-pager --distribution-id ${{ secrets.TRUNK_DOC_DISTRIBUTION_ID }} --paths "/*"
           aws s3 sync ./site "s3://trunk-docs.kolena.com" --delete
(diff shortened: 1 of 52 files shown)


grant-Kolena:
Nice! Glad to see the build is working again

grant-Kolena (approved):

pyproject.toml
```diff
@@ -23,61 +25,37 @@ classifiers = [# classifiers for license, versions set automatically during Poet
...
-pytest-depends = "^1.0.1"
-mkdocs = "^1.4.3"
-cairosvg = "^2.7.0"
-mkdocs-material = ">=9.2,<10"  # insiders fork installed out-of-band in docs/setup_insiders.sh
-mkdocstrings = { version = ">0.20,<1", extras = ["python"] }
-mkdocs-git-committers-plugin-2 = "^1"
-mkdocs-git-revision-date-localized-plugin = "^1"
-
-[tool.poetry.scripts]
-kolena = 'kolena._utils.cli:run'
```
yifanwu-kolena: i think we need to find the counterpart of uv, b/c it provides a kolena cli in the python venv after installing kolena package

munkyshi: @yifanwu-kolena what is this used for / who uses it (is it something publicly shared / documented)? I couldn't find a great counterpart, and I'm not sure it's worth blocking this based on the cli.

munkyshi: Ah, I see: https://docs.kolena.com/workflow/advanced-usage/packaging-for-automated-evaluation/ -- do we need to preserve this capability?

munkyshi: Actually, this seems to work out of the box with uv!
```
❯ uv run kolena --help
Ignoring existing lockfile due to change in supported environments
Usage: kolena [OPTIONS] COMMAND [ARGS]...

Options:
  -h, --help  Show this message and exit.

Commands:
  evaluator
  repository
```

yifanwu-kolena: this is cool, but how can uv know the cli entrypoint? 🤔 

based on [the doc](https://github.com/astral-sh/uv/blob/039b68367e6481690ca001c44e61500bd1e42570/docs/concepts/projects.md?plain=1#L210-L255), we can define `[project.scripts]`

munkyshi: Hmm, I wonder if there was something leftover from my local state with `[tool.poetry.scripts]`

Nice catch, re-added as `[project.scripts]`

yifanwu-kolena: maybe there are some heuristic magic behind the scene, but i believe explicit is better than implicit

.github/workflows/docs.yml
```diff
@@ -38,25 +38,24 @@ jobs:
...
       - name: Fetch tags to enable autoversioning
         run: git fetch --prune --unshallow --tags
 
-      - name: Update package version to PEP 440-compliant production release tag
-        run: poetry version $(git describe --tags --abbrev=0)
-
-      - name: Install Poetry dependencies
-        run: poetry install
+      - name: Install dependencies
+        run: uv pip install --all-extras -r pyproject.toml
```
yifanwu-kolena, line 50: does this handle the dynamic release tag in uv? my concern is that we might publish the version as `0.999.0`

docs/automations/set-up-natural-language-search.md
```diff
@@ -70,16 +70,11 @@ export KOLENA_TOKEN="********"
...
-    ```shell
-    poetry source add --priority=supplemental kolena-embeddings "https://gateway.kolena.cloud/repositories"
-    ```
-
-    Then run the following command, making sure to replace <KOLENA_TOKEN> with the token retrieved from the developer page:
+    Run the following command, making sure to replace <KOLENA_TOKEN> with the token retrieved from the developer page:
     ```shell
-    poetry config http-basic.kolena-embeddings <KOLENA_TOKEN> ""
+    uv add --extra-index-url="https://<KOLENA_TOKEN>@gateway.kolena.cloud/repositories" kolena-embeddings
     ```
```
yifanwu-kolena, line 78: just want to verify that we tested this manually already

yifanwu-kolena (approved):
🚢

docs/automations/set-up-natural-language-search.md
munkyshi: Yep, I ran this manually and was able to install `kolena-embeddings`

PR #686, opened 2024-09-17
feat: experimental method to download results by model tag

nankolena:
### Linked issue(s)

Resolves KOL-6900
Corresponding backend change: https://github.com/kolenaIO/shipyard/pull/2450

CI is expected to fail before the above change is in production. Planning to cut a minor version after 1.36 release.

- During `upload_results`, allow optionally specifying a list of string tags to be associated with the model. A model can be associated with multiple tags, and a tag can by associated with multiple models.
- Creating a method under experimental folder to support downloading results by specifying the model tag. Technically speaking, the backend change supports directly downloading results even if a tag is associated with multiple models. That however will make the function signature nested and complicated. Given that the immediate use case we see is a one-to-one mapping between a model and its human-readable tag, limiting the function by inspecting that only 1 model is associated with the provided tag and keeping the same return type as `download_results`. In case multiple models share the same tag, the exception message will contain the model names to help users identify which ones to use.

Also verified docs update:
image

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -494,7 +494,7 @@ workflows:
           matrix:
             parameters:
               subproject: [ age_estimation, automatic_speech_recognition, classification, keypoint_detection, question_answering, rain_forecast, semantic_segmentation, speaker_diarization, object_detection_2d, semantic_textual_similarity, person_detection, crossing_pedestrian_detection, named_entity_recognition ]
-              resource-class: [ medium ]
+              resource-class: [ medium ]    # TODO: revert after KOL-7307 is resolved [KOL-7444]
               python-version: [ "3.9.18" ]
       - example-test-workflow:
           matrix:
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -447,6 +447,16 @@ workflows:
           enabled: true
           requires:
             - ci-base-<< matrix.python-version >>-<< matrix.extras >>
+      - integration-test:
+          name: integration-test-experimental-dataset-<< matrix.python-version >>-<< matrix.extras >>
+          matrix:
+            parameters:
+              python-version: [ "3.9.18" ]
+              extras: [ "none" ]
+          pytest-group: _experimental/dataset
+          enabled: true
+          requires:
+            - ci-base-<< matrix.python-version >>-<< matrix.extras >>
       - integration-test-dataset:
           name: integration-test-dataset-<< matrix.python-version >>-<< matrix.extras >>
           matrix:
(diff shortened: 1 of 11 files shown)


.circleci/continue_config.yml
```diff
@@ -447,6 +447,16 @@ workflows:
           enabled: true
           requires:
             - ci-base-<< matrix.python-version >>-<< matrix.extras >>
+      - integration-test:
```
munkyshi, line 450: Let's revert this as soon as #683 merges

nankolena: Yup, created a ticket in linear to keep track: https://linear.app/kolena/issue/KOL-7444/use-small-resource-instance-for-client-repo-tests

kolena/_experimental/dataset/evaluation.py
```diff
@@ -0,0 +1,81 @@
...
+    return [model.name for model in load_by_tag_response.models]
+
+
+@with_event(EventAPI.Event.FETCH_DATASET_MODEL_RESULT_BY_TAG)
+def download_results_by_tag(
+    dataset: str,
+    model_tag: str,
+    commit: Optional[str] = None,
+    include_extracted_properties: bool = False,
+) -> tuple[pd.DataFrame, list[EvalConfigResults]]:
```
munkyshi: I don't think typing like `tuple` and `list` existed until 3.9, which means we need to use `typing.Tuple` and `typing.List` as long as we support 3.8.

https://peps.python.org/pep-0585/

kolena/_experimental/dataset/evaluation.py
```diff
@@ -0,0 +1,81 @@
...
+        json=asdict(LoadByTagRequest(tag=tag)),
+    )
+    response.raise_for_status()
+
+    load_by_tag_response = from_dict(LoadByTagResponse, response.json())
+    return [model.name for model in load_by_tag_response.models]
+
+
+@with_event(EventAPI.Event.FETCH_DATASET_MODEL_RESULT_BY_TAG)
+def download_results_by_tag(
```
munkyshi, line 52: We should add a docstring to indicate that this is currently restricted to if the tag is unique to that model

.circleci/continue_config.yml
munkyshi: Whoops, wrong line.

.circleci/continue_config.yml
```diff
@@ -484,7 +494,7 @@ workflows:
           matrix:
             parameters:
               subproject: [ age_estimation, automatic_speech_recognition, classification, keypoint_detection, question_answering, rain_forecast, semantic_segmentation, speaker_diarization, object_detection_2d, semantic_textual_similarity, person_detection, crossing_pedestrian_detection, named_entity_recognition ]
-              resource-class: [ small ]
+              resource-class: [ medium ]
```
munkyshi: Let's revert this as soon as https://github.com/kolenaIO/kolena/pull/683 merges. Also worth adding a comment saying so, assuming we can do it without an extra CI trigger.

munkyshi (approved):
Some small nits, but approving to unblock after they are addressed.

PR #695, opened 2024-09-25
Update build steps to exclude examples, docs

gordonhart:
Should fix the [failed builds](https://github.com/kolenaIO/kolena/actions/runs/11025258075) during last night's release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,7 +49,7 @@ jobs:
         run: sed -i "s/0.999.0/$(git describe --tags --abbrev=0)/g" pyproject.toml
 
       - name: Build 'kolena' Python package
-        run: uv build --sdist
+        run: uvx hatch build -t sdist
 
       - name: Publish 'kolena' documentation to production (S3)
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -94,7 +94,7 @@ jobs:
           sed -i '0,/kolena/{s/kolena/kolena-client/}' pyproject.toml kolena/__init__.py
 
           uv pip install --all-extras -r pyproject.toml
-          uv build --sdist
+          uvx hatch build -t sdist
```
millerpeterson, line 97: FWIW `uv build --sdist` works too.

millerpeterson (approved):

.github/workflows/release.yml
gordonhart: Hmm I looked at the `--verbose` output for both commands and it's not-insignificantly different between those two commands. I generally trust running `hatch` directly more while `uv` is in its early releases — going to leave these changes up for now.

grant-Kolena (approved):

PR #710, opened 2024-10-11
Add basic example for polygon-based instance segmentation in datasets

munkyshi:
### Linked issue(s)
N/A

Add basic example for polygon-based instance segmentation in datasets

### DIFF
--- a/examples/dataset/instance_segmentation/pyproject.toml
+++ b/examples/dataset/instance_segmentation/pyproject.toml
@@ -1,0 +1,27 @@
+[project]
+name = "instance_segmentation"
+version = "0.1.0"
+description = "Example Kolena integration for Instance Segmentation"
+authors = [
+    { name = "Kolena Engineering", email = "eng@kolena.com" }
+]
+license = "Apache-2.0"
+requires-python = ">=3.9,<3.12"
+
+dependencies = [
+    "kolena[metrics]>=1.23.0,<2",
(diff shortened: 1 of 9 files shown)


examples/dataset/instance_segmentation/pyproject.toml
```diff
@@ -0,0 +1,27 @@
...
+authors = [
+    { name = "Kolena Engineering", email = "eng@kolena.com" }
+]
+license = "Apache-2.0"
+requires-python = ">=3.9,<3.12"
+
+dependencies = [
+    "kolena[metrics]>=1.23.0,<2",
+    "s3fs>=2023.5.0,<2024",
+    "fsspec>=2023.5.0,<2024",
```
nankolena, line 14: non-blocking - I noticed similar pattern in other examples too that capped the versions for these, do you remember the rationale for this?

nankolena (approved):

examples/dataset/instance_segmentation/pyproject.toml
munkyshi: Whoops, had an auto-merge on -- this was basically to match the effect of `poetry` capping the major version with `^2023.5.0` -- I suppose in this case, the major version doesn't necessarily mean an API break, and it's safe to leave these uncapped, but I didn't change it to keep consistent with OD.

PR #714, opened 2024-10-17
chore: update doc to reflect SDK support

brianshen3:
### Linked issue(s)
Fixes KOL-7637
Updated OD docs to reflect model results upload is only supported by SDK. And the client will need to reupload results if metrics change.

### DIFF
--- a/docs/dataset/object-detection.md
+++ b/docs/dataset/object-detection.md
@@ -233,40 +233,11 @@ For details on what is recommended within your model results, see the relevant d
 
 ### Step 4: Upload Object Detection Model Results
 
-=== "Web App"
-
-    We will upload object detection model results of [YOLO X](https://github.com/Megvii-BaseDetection/YOLOX)
-    for a single class object detection task.
-    To upload new model results, from the `Details` tab of the dataset, click on `Upload Model Results`
-    in the upper right.
-    Then, select `Upload From Cloud Storage`. Using the explorer, navigate to
-    `s3://kolena-public-examples/coco-2014-val/person-detection/results/` and select the `yolo_x` CSV.
-
-    You will now see a preview of how Kolena will ingest the model results. Give your model a name,
-    and click `Import` to upload the model results.
-
-    <figure markdown>
-    ![Model Results Upload](../assets/images/task-od-upload-results-light.gif#only-light)
-    ![Model Results Upload](../assets/images/task-od-upload-results-dark.gif#only-dark)
-    <figcaption markdown>Model Results Upload</figcaption>
-    </figure>
-
-    <div class="grid" markdown>
-    <div markdown>
-    <figure markdown>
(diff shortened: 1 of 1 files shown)


docs/dataset/object-detection.md
```diff
@@ -233,40 +233,11 @@ For details on what is recommended within your model results, see the relevant d
 
 ### Step 4: Upload Object Detection Model Results
 
-=== "Web App"
```
nankolena, line 236: Thanks for picking this up.

This is sort of gray area in that, in this example it actually can work because the files in `s3://kolena-public-examples/coco-2014-val/person-detection/results/` were actually exported from the Kolena web app after someone ran `upload_object_detection_results` -- i.e. this is a special case where the data in the csv files are indeed outputs from the sdk clients. 

I think there's merit in this example to keep the web app upload part so that an OD customer can set up a dataset fairly easily via the UI, but maybe add a note to say that generally only sdk client uploads would work for OD. Let me know what you think!

brianshen3: It's a good point. Let me think about this

brianshen3: I updated the docs based on your feedback. Could you take a look when you have a chance?

docs/dataset/object-detection.md
```diff
@@ -283,7 +290,7 @@ For details on what is recommended within your model results, see the relevant d
     [uploading object detection model results](../dataset/advanced-usage/dataset-formatting/computer-vision.md#uploading-model-results).
 
     ```shell
-    uv run python3 object_detection_2d/upload_results.py faster_rcnn
+    poetry run python3 object_detection_2d/upload_results.py faster_rcnn
```
nankolena: Did you have to use `poetry` on this example? I thought we migrated all to `uv` in this repo

docs/dataset/object-detection.md
```diff
@@ -164,11 +164,9 @@ example script for details on how an object detection dataset can be generated a
 
 ### Step 2: Upload your Object Detection Dataset
 
-Model evaluation on Kolena starts with [datasets](../dataset/core-concepts/index.md#kolena-dataset-20-dataset).
-Upload your dataset of datapoints (e.g. locators to
+Model evaluation on Kolena starts with datasets. Upload your dataset of datapoints (e.g. locators to
```
nankolena: Are these changes that remove hyperlinks intentional?

docs/dataset/object-detection.md
nankolena: Thanks - this change looks good! I left 2 small comments to other changes in this PR.

nankolena (approved):

docs/dataset/object-detection.md
brianshen3: Updated this. I used uv

PR #726, opened 2024-11-19
chore: created function to remove result rows greater than 16mb 

brianshen3:
### Linked issue(s)
Towards KOL-7746
created function to remove result rows greater than 16mb to fix issues with uploading data that doesn't fit into snowflake limit. And fixed log by adding missing new line.

### DIFF
--- a/kolena/_experimental/object_detection/dataset.py
+++ b/kolena/_experimental/object_detection/dataset.py
@@ -474,6 +474,7 @@ def upload_object_detection_results(
     min_confidence_score: float = 0.01,
     batch_size: int = 10_000,
     required_match_fields: Optional[List[str]] = None,
+    row_limit: bool = False,
 ) -> None:
     """
     Compute metrics and upload results of the model computed by
@@ -502,6 +503,8 @@ def upload_object_detection_results(
         the inference and ground truth for them to be considered a match.
     :return:
     """
+    if row_limit:
+        df = remove_oversized_rows(df)
     eval_config = dict(
         iou_threshold=iou_threshold,
         threshold_strategy=threshold_strategy,
@@ -509,7 +512,7 @@ def upload_object_detection_results(
     )
     results = _iter_object_detection_results(
         dataset_name,
-        remove_oversized_rows(df),
+        df,
         ground_truths_field=ground_truths_field,
         raw_inferences_field=raw_inferences_field,
         gt_ignore_property=gt_ignore_property,
--- a/kolena/_experimental/object_detection/dataset.py
+++ b/kolena/_experimental/object_detection/dataset.py
@@ -431,7 +431,7 @@ def compute_object_detection_results(
     return pd.concat(list(results_iter))
 
 
-def get_object_size(obj):
+def get_object_size(obj: Any) -> int:
     """Get approximate size of a Python object in bytes."""
     return len(json.dumps(str(obj)).encode("utf-8"))
 
--- a/kolena/_experimental/object_detection/dataset.py
+++ b/kolena/_experimental/object_detection/dataset.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import itertools
+import json
 from collections import defaultdict
 from typing import Any
 from typing import cast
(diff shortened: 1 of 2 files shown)


nankolena (commented):
Do we need this? I feel like there's not enough datapoint that warrants the need for this, while it has the potential of failing "silently" if the customer is not paying attention to the logs. My thinking is that we'd want to fail explicitly for now instead

brianshen3:
> Do we need this? I feel like there's not enough datapoint that warrants the need for this, while it has the potential of failing "silently" if the customer is not paying attention to the logs. My thinking is that we'd want to fail explicitly for now instead



> Do we need this? I feel like there's not enough datapoint that warrants the need for this, while it has the potential of failing "silently" if the customer is not paying attention to the logs. My thinking is that we'd want to fail explicitly for now instead

The issue I am stuck on this with Plus AI. Hamed and the team isn't able to resolve this on their end. I shared with them the code to resolve this but they seem to ignore and choose a code on their end which didn't work. I do think you bring a good point. I will add this as a option to the function call instead of a default run. This would solve the issue with PlusAI and we can have this for future use cases. I will let Hamed know he has that option and the issue is still on his end.

nankolena:
> > Do we need this? I feel like there's not enough datapoint that warrants the need for this, while it has the potential of failing "silently" if the customer is not paying attention to the logs. My thinking is that we'd want to fail explicitly for now instead
> 
> > Do we need this? I feel like there's not enough datapoint that warrants the need for this, while it has the potential of failing "silently" if the customer is not paying attention to the logs. My thinking is that we'd want to fail explicitly for now instead
> 
> The issue I am stuck on this with Plus AI. Hamed and the team isn't able to resolve this on their end. I shared with them the code to resolve this but they seem to ignore and choose a code on their end which didn't work. I do think you bring a good point. I will add this as a option to the function call instead of a default run. This would solve the issue with PlusAI and we can have this for future use cases. I will let Hamed know he has that option and the issue is still on his end.

I see, instead of merging this to the public client repo, can we instead sharing this commit if/when we run into this? I'm reluctant having this in the public client repo unless really necessary, and it sounds so far like an edge case

brianshen3:
> > > Do we need this? I feel like there's not enough datapoint that warrants the need for this, while it has the potential of failing "silently" if the customer is not paying attention to the logs. My thinking is that we'd want to fail explicitly for now instead
> > 
> > 
> > > Do we need this? I feel like there's not enough datapoint that warrants the need for this, while it has the potential of failing "silently" if the customer is not paying attention to the logs. My thinking is that we'd want to fail explicitly for now instead
> > 
> > 
> > The issue I am stuck on this with Plus AI. Hamed and the team isn't able to resolve this on their end. I shared with them the code to resolve this but they seem to ignore and choose a code on their end which didn't work. I do think you bring a good point. I will add this as a option to the function call instead of a default run. This would solve the issue with PlusAI and we can have this for future use cases. I will let Hamed know he has that option and the issue is still on his end.
> 
> I see, instead of merging this to the public client repo, can we instead sharing this commit if/when we run into this? I'm reluctant having this in the public client repo unless really necessary, and it sounds so far like an edge case

Okay, I will take another approach. I will let Hamed know we decided it was a edge case. I will ask him to share his upload code. I will reiteriate the solution to him again.

nankolena:
> > > > Do we need this? I feel like there's not enough datapoint that warrants the need for this, while it has the potential of failing "silently" if the customer is not paying attention to the logs. My thinking is that we'd want to fail explicitly for now instead
> > > 
> > > 
> > > > Do we need this? I feel like there's not enough datapoint that warrants the need for this, while it has the potential of failing "silently" if the customer is not paying attention to the logs. My thinking is that we'd want to fail explicitly for now instead
> > > 
> > > 
> > > The issue I am stuck on this with Plus AI. Hamed and the team isn't able to resolve this on their end. I shared with them the code to resolve this but they seem to ignore and choose a code on their end which didn't work. I do think you bring a good point. I will add this as a option to the function call instead of a default run. This would solve the issue with PlusAI and we can have this for future use cases. I will let Hamed know he has that option and the issue is still on his end.
> > 
> > 
> > I see, instead of merging this to the public client repo, can we instead sharing this commit if/when we run into this? I'm reluctant having this in the public client repo unless really necessary, and it sounds so far like an edge case
> 
> Okay, I will take another approach. I will let Hamed know we decided it was a edge case. I will ask him to share his upload code. I will reiteriate the solution to him again.

Sounds good! Sorry for lacking the context here, but feel free to share this code snippet with him too. My perspective is that we should keep an eye on this LOB size usage and if results upload fails we should be aware instead of having them skipped

PR #742, opened 2025-01-13
feat: added logic to handle labeled answers

brianshen3:
Handle compute metrics for labeled answers in the script.

### DIFF
--- a/examples/dataset/retrieval_augmented_generation/retrieval_augmented_generation/metrics.py
+++ b/examples/dataset/retrieval_augmented_generation/retrieval_augmented_generation/metrics.py
@@ -16,17 +16,15 @@ from retrieval_augmented_generation.constants import ID_FIELDS
 
 
 def is_doc_retrieved(retrieved_contents: list, doc_names: list) -> bool:
-    # Handle case where doc_names is a list
+    # NOTE: if one doc is retrieved, it is considered correct
     return any(any(doc_name in content.locator for content in retrieved_contents) for doc_name in doc_names)
 
 
 def is_page_retrieved(retrieved_contents: list, doc_names: list, relevant_pages: list) -> bool:
-    # Create pairs of (doc_name, page_number) from retrieved contents
+    # NOTE: if one page of any doc is retrieved, it is considered correct
     retrieved_pairs = [
         (content.locator.split("/")[-1].replace(".pdf", ""), content.page_number) for content in retrieved_contents
     ]
-
-    # Check if any of the relevant page pairs match with retrieved pairs
     return any(pair in retrieved_pairs for pair in relevant_pages)
 
 
--- a/examples/dataset/retrieval_augmented_generation/retrieval_augmented_generation/metrics.py
+++ b/examples/dataset/retrieval_augmented_generation/retrieval_augmented_generation/metrics.py
@@ -27,7 +27,7 @@ def is_page_retrieved(retrieved_contents: list, doc_names: list, relevant_pages:
     ]
 
     # NOTE: all relevant pages must be retrieved to be considered correct.
-    return set(relevant_pages).issubset(retrieved_pages)
+    return set(relevant_pages).issubset(retrieved_pages) and is_doc_retrieved(retrieved_contents, doc_names)
 
 
 def compute_metrics(df_dataset: pd.DataFrame, df_results: pd.DataFrame) -> pd.DataFrame:
--- a/examples/dataset/retrieval_augmented_generation/retrieval_augmented_generation/metrics.py
+++ b/examples/dataset/retrieval_augmented_generation/retrieval_augmented_generation/metrics.py
@@ -15,22 +15,47 @@ import pandas as pd
 from retrieval_augmented_generation.constants import ID_FIELDS
 
 
-def is_doc_retrieved(retrieved_contents: list, doc_name: str) -> bool:
-    return any([doc_name in content.locator for content in retrieved_contents])
+def is_doc_retrieved(retrieved_contents: list, doc_names: list) -> bool:
+    # Handle case where doc_names is a list
+    return any(any(doc_name in content.locator for content in retrieved_contents) for doc_name in doc_names)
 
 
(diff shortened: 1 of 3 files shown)


marko-polo-cheno:
With a partially labelled dataset and `uv run python3 retrieval_augmented_generation/upload_results.py query_decomp --dataset-name financebenchmarknew --evaluate`, I am seeing `kolena> Failed to import module for data type 'SPECIAL/TIMESTAMP'`.

Also seeing metrics indicate the correct page is retrieved, but the wrong document is retrieved which is contradictory.
image

brianshen3:
Fixed the timestamp error. Was due to new data type needed to be imported as it was read in the download.

As for the metrics error. It was because originally there was only one doc. But with labelling it could be a list of docs. Updated the logic to if retrieved_docs is True then is_page can be true. Let me know if that logic makes sense @marko-polo-cheno

kolena/_utils/datatypes.py
```diff
@@ -99,7 +99,7 @@ def data_category_to_module_name(self) -> str:
         if self == DataCategory.ANNOTATION:
             return "kolena.annotation"
         if self == DataCategory.SPECIAL:
-            return "kolena._experimental.data_type.special"
+            return "kolena._experimental.special_data_type"
```
brianshen3: @nankolena There were errors regarding the imports. I couldn't find `kolena._experimental.data_type.special`

I fixed it to the one aligned in the docs. Let me know if this is right track.

nankolena: Ah good call, missed this when refactoring the module, thanks for picking this up! Once this is release we should bump the min kolena version in this example

kolena/_utils/datatypes.py
```diff
@@ -116,7 +116,7 @@ def _get_data_type(name: str) -> Optional[Type["TypedDataObject"]]:
             module = DataCategory(data_category).data_category_to_module_name()
             importlib.import_module(module)
         except Exception as e:
-            log.error(f"Failed to import module for data type '{name}'", e)
+            log.error(f"Failed to2 import module for data type '{name}'", e)
```
nankolena: nit - clean up

kolena/_utils/datatypes.py
nankolena: Actually lemme make a separate change for this so we can cut a minor release off of it quickly

brianshen3: Ok I will move the change outside this PR

examples/dataset/retrieval_augmented_generation/retrieval_augmented_generation/metrics.py
```diff
@@ -15,22 +15,47 @@
...
+            if x is not None
+            else [],
+        )
+        # Extract page numbers from retrieved contents
+        df_dataset["relevant_pages"] = df_dataset["labeling_task"].apply(
+            lambda x: [content.page_number for content in x.get("retrieved_contents", []) or []]
+            if x is not None
+            else [],
+        )
+        ground_truth_columns = ["doc_name", "relevant_pages", "financebench_id"]
```
marko-polo-cheno: Can we move `ground_truth_columns` outside of the if-statement?

Something like this might be cleaner
```
def extract_doc_names(labeling_task):
    if labeling_task is None:
        return []
    return [
        content.locator.split("/")[-1].replace(".pdf", "")
        for content in labeling_task.get("retrieved_contents", []) or []
    ]

def extract_relevant_pages(labeling_task):
    if labeling_task is None:
        return []
    return [
        content.page_number
        for content in labeling_task.get("retrieved_contents", []) or []
    ]

df_dataset["doc_name"] = df_dataset["labeling_task"].apply(extract_doc_names)
df_dataset["relevant_pages"] = df_dataset["labeling_task"].apply(extract_relevant_pages)

```

examples/dataset/retrieval_augmented_generation/retrieval_augmented_generation/metrics.py
```diff
@@ -15,22 +15,47 @@
 from retrieval_augmented_generation.constants import ID_FIELDS
 
 
-def is_doc_retrieved(retrieved_contents: list, doc_name: str) -> bool:
-    return any([doc_name in content.locator for content in retrieved_contents])
+def is_doc_retrieved(retrieved_contents: list, doc_names: list) -> bool:
+    # Handle case where doc_names is a list
+    return any(any(doc_name in content.locator for content in retrieved_contents) for doc_name in doc_names)
```
marko-polo-cheno, line 20: Should specify that we mark a model result as True/correct if just one document from the ground truth is found in the retrieved docs and likewise for the page metric.

In other words, any non-zero overlap for two sets of citations is treated as correct. Further improvements for the future would probably include an IOU score.

marko-polo-cheno (approved):

kolena/_utils/datatypes.py
nankolena: Cut [1.51.1 client version](https://github.com/kolenaIO/kolena/releases/tag/1.51.1) just now, release might take a bit of time. Could you rebase and update this example to `>=1.51.1` later?

brianshen3: No rush, its okay

examples/dataset/retrieval_augmented_generation/retrieval_augmented_generation/metrics.py
brianshen3: that's a good suggestion, I will update to reflect that

brianshen3: Updated to have note for the logic we use

examples/dataset/retrieval_augmented_generation/retrieval_augmented_generation/metrics.py
brianshen3: moved to cleaner logic. I also aligned to use doc_names instead of doc_name.

kolena/_utils/datatypes.py
brianshen3: Updated the versioning.

nankolena (approved):

PR #750, opened 2025-01-24
feat: to use file name field in embedding extraction

brianshen3:

### DIFF
--- a/examples/dataset/search_embeddings/search_embeddings/upload_embeddings.py
+++ b/examples/dataset/search_embeddings/search_embeddings/upload_embeddings.py
@@ -32,7 +32,7 @@ from kolena.dataset import download_dataset
 
 
 BUCKET = "kolena-public-examples"
-DATASET = "coco-2014-val"
+DATASET = "coco-stuff-10k"
 IMAGE_S3_DIR = f"s3://{BUCKET}/{DATASET}/data/images/"  # noqa: E231
 LOCATOR_FIELD = "locator"
 FILE_NAME_FIELD = "file_name"
--- a/examples/dataset/search_embeddings/search_embeddings/upload_embeddings.py
+++ b/examples/dataset/search_embeddings/search_embeddings/upload_embeddings.py
@@ -46,8 +46,8 @@ def image_locators_from_s3_path(
     locators_and_filepaths: List[Tuple[str, Optional[str]]] = []
     for _, row in df.iterrows():
         locator = row[LOCATOR_FIELD]
-        file_name = row[FILE_NAME_FIELD]
-        file_extension = row[FILE_EXTENSION_FIELD]
+        file_name = row.get(FILE_NAME_FIELD, None)
+        file_extension = row.get(FILE_EXTENSION_FIELD, None)
         if not locator.startswith("s3://"):
             raise ValueError(f"invalid input path: {locator}")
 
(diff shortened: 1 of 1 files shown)


examples/dataset/search_embeddings/search_embeddings/upload_embeddings.py
```diff
@@ -33,28 +33,40 @@
 
 BUCKET = "kolena-public-examples"
 DATASET = "coco-stuff-10k"
-IMAGE_S3_DIR = f"s3://{BUCKET}/{DATASET}/data/images/"
+IMAGE_S3_DIR = f"s3://{BUCKET}/{DATASET}/data/images/"  # noqa: E231
```
nankolena, line 36: curious - do we really need this? I'm surprised if flake8 raise style errors on string literal

examples/dataset/search_embeddings/search_embeddings/upload_embeddings.py
```diff
@@ -33,28 +33,40 @@
...
-    s3_locators: List[str],
+    df: pd.DataFrame,
     local_dir: Optional[str] = None,
 ) -> List[Tuple[str, Optional[str]]]:
     locators_and_filepaths: List[Tuple[str, Optional[str]]] = []
-    for locator in s3_locators:
+    for _, row in df.iterrows():
+        locator = row[LOCATOR_FIELD]
+        file_name = row[FILE_NAME_FIELD]
+        file_extension = row[FILE_EXTENSION_FIELD]
```
nankolena: These columns may or may not exist right? Should we do `.get(xx, None)` instead?

examples/dataset/search_embeddings/search_embeddings/upload_embeddings.py
```diff
@@ -33,28 +33,40 @@
...
             locators_and_filepaths.append((locator, None))
         else:
-            relative_locator = os.path.relpath(locator[5:], IMAGE_S3_DIR[5:])
-            target = os.path.join(local_dir, relative_locator)
-            if not os.path.exists(target):
-                raise ValueError(f"missing local file: {target}")
-            locators_and_filepaths.append((locator, target))
-
+            # Target is the local filepath to the image
+            if file_name is None or file_extension is None:
```
nankolena, line 58: If I understand the goal here correctly by reading [your comment](https://linear.app/kolena/issue/KOL-7903/up-setting-up-kolena-embedding-extraction#comment-b93db5b2), the issue we try to address here is that one customer had issues using this example code out of the box because they have a slightly different setup in that their local file names are not the same with those in cloud storage. In their dataset, in addition to the image `locator`, they will have `file_name` and `file_extension` to reflect their local file structure, and we'll be using these additional columns to find the local file. Is this correct?

If so I think the changes you proposed should work for the customer. We can share the commit with the customer but shouldn't update the public example to cater to this specific set up because the public example is meant to be a generic one to get a user started. If a new customer were to look at this example after this change, they will likely be confused why it's expecting for example `file_name` and `file_extension`

examples/dataset/search_embeddings/search_embeddings/upload_embeddings.py
brianshen3: Yes, flake8 thinks it's part of a function
Screenshot 2025-01-24 at 2 39 46 PM

examples/dataset/search_embeddings/search_embeddings/upload_embeddings.py
brianshen3: Thanks Nan. That makes sense. I will close out the PR later.

nankolena: For sure - lmk if you want a second pair of eyes to go through things before sharing with the customer!

PR #752, opened 2025-02-03
feat: support editing dataset description programmatically

nankolena:
### Linked issue(s)
Resolves KOL-7913

Add new parameter to `upload_dataset` call to include a string `description` to support programmatically specifying the dataset description during creation/update.

This corresponds to https://github.com/kolenaIO/shipyard/pull/2606 from the backend. Integration test is expected to fail before the backend change is in production. However, tested locally with my dev workspace [here](https://trunk.kolena.io/nan/dataset?datasetId=38).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -95,7 +95,7 @@ dev-dependencies = [
     # insiders fork installed out-of-band in docs/setup_insiders.sh
     "mkdocs-material>=9.2,<10",
     "mkdocstrings>=0.25,<1",
-    "mkdocstrings-python>=1.11.1,<2",
+    "mkdocstrings-python==1.11.1",
     "mkdocs-git-committers-plugin-2>=1,<2",
     "mkdocs-git-revision-date-localized-plugin>=1,<2",
 ]
--- a/kolena/_api/v2/dataset.py
+++ b/kolena/_api/v2/dataset.py
@@ -38,6 +38,7 @@ class RegisterRequest:
     append_only: bool = False
     tags: Optional[List[str]] = None
     dataset_tags: Optional[List[str]] = None
+    description: Optional[str] = None
 
 
 @dataclass(frozen=True)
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -95,7 +95,7 @@ dev-dependencies = [
     # insiders fork installed out-of-band in docs/setup_insiders.sh
     "mkdocs-material>=9.2,<10",
     "mkdocstrings>=0.25,<1",
-    "mkdocstrings-python>=1.11.1,<2",
+    "mkdocstrings-python==1.11.1",
```
nankolena, line 98: Unsure why we need this exactly. Without this pinned, I was getting [doc build error](https://app.circleci.com/pipelines/github/kolenaIO/kolena/5149/workflows/4d767b54-8697-426c-83be-55a7ad6ede5b/jobs/130504):
```
TypeError: Unable to evaluate type annotation "Annotated[str | None, Field(group='docstrings', parent='docstring_options', description='The default docstring style to use if no other style is detected.')]". If you are making use of the new typing syntax (unions using `|` since Python 3.10 or builtins subscripting since Python 3.9), you should either replace the use of new syntax with the existing `typing` constructs or install the `eval_type_backport` package.
```

Though I was not able to reproduce this locally. This technically should be no-ops because after `1.11.1`, [mkdocsstrings-python dropped support for py 3.8 that we still need to support.](https://github.com/mkdocstrings/python/releases/tag/1.12.0).

Worth noting that this specific error message seems to come from [a newer version of mkdocstrings-python 1.42.0](https://github.com/mkdocstrings/python/commit/5ebeda6fce1b1bc7cb3f5e27a5a90ac394a3de0c#diff-0c40e032c32782e0f18420cd22b0f0ecf8c3bfed21a3ed465b229b21d1f1f2a1)

grant-Kolena (approved):

PR #754, opened 2025-02-07
chore: moved upload embeddings out of experimental

brianshen3:
### Linked issue(s)
[KOL-7957](https://linear.app/kolena/issue/KOL-7957/move-upload-embeddings-out-of-the-experimental-section)
Move `upload_dataset_embeddings` out of _experimental

- [x] Relevant docs have been added / updated (Going to do this separate PR/ticket)


kolena/_experimental/search/embeddings.py
```diff
@@ -11,131 +11,21 @@
...
 from kolena._utils.instrumentation import with_event
-from kolena._utils.state import API_V2
-from kolena.dataset._common import COL_DATAPOINT_ID_OBJECT
-from kolena.dataset._common import validate_dataframe_ids
-from kolena.dataset.dataset import _load_dataset_metadata
-from kolena.dataset.dataset import _to_serialized_dataframe
-from kolena.errors import InputValidationError
-
-
-def upload_embeddings(key: str, embeddings: List[Tuple[str, np.ndarray]]) -> None:
```
nankolena, line 48: This is the function to upload embeddings for workflows. Logically it does make sense to move this out of experimental, but then it should not be part of `/dataset` because it's not used for datasets. To avoid complexities I think it's fine to keep this here. What do you think?

kolena/_experimental/search/embeddings.py
```diff
@@ -11,131 +11,21 @@
...
-        api_version=API_V2,
-        data=json.dumps(dataclasses.asdict(request)),
-    )
-    krequests.raise_for_status(res)
-    data = from_dict(data_class=UploadDatasetEmbeddingsResponse, data=res.json())
-    log.success(f"uploaded embeddings for dataset '{dataset_name}' and key '{key}' on {data.n_datapoints} datapoints")
+from kolena.dataset.search import upload_dataset_embeddings as new_upload_dataset_embeddings
 
 
 @with_event(event_name=EventAPI.Event.UPLOAD_DATASET_EMBEDDINGS)
```
nankolena: nit - should remove this, otherwise the event would be recorded twice because this calls the other `upload_dataset_embeddings`

kolena/_experimental/search/embeddings.py
```diff
@@ -11,131 +11,21 @@
...
 def upload_dataset_embeddings(dataset_name: str, key: str, df_embedding: pd.DataFrame) -> None:
     """
     Upload a list of search embeddings for a dataset.
 
+    .. deprecated:: 1.54.2
+        Use :func:`kolena.dataset.search.upload_dataset_embeddings` instead.
+
```
nankolena: Nice job adding deprecation notice! A couple questions:
- Should this be 1.55.0 (the expected next release)?
- Its not rendered nicely, is there some syntax issue?
image

nankolena (commented):
I think you are missing some changes to the docs index too because it's currently still shown under the experimental page. Probably need to [remove this block](https://github.com/kolenaIO/kolena/blob/ed91015a9fc6f17f6652819d2cced84ea9d30471/docs/reference/experimental/index.md?plain=1#L7-L10) and check if you need to add anything for the dataset page

image

kolena/_experimental/search/embeddings.py
```diff
@@ -144,4 +34,4 @@ def upload_dataset_embeddings(dataset_name: str, key: str, df_embedding: pd.Data
     :raises NotFoundError: The given dataset does not exist.
     :raises InputValidationError: The provided input is not valid.
     """
-    _upload_dataset_embeddings(dataset_name, key, df_embedding)
+    new_upload_dataset_embeddings(dataset_name, key, df_embedding)
```
nankolena, line 92: nit - consider adding a warning message here so that users will know this is deprecated.

kolena/_experimental/search/embeddings.py
brianshen3: Yes, lets keep it here for now. If needed I can fix it later.

brianshen3:
> I think you are missing some changes to the docs index too because it's currently still shown under the experimental page. Probably need to [remove this block](https://github.com/kolenaIO/kolena/blob/ed91015a9fc6f17f6652819d2cced84ea9d30471/docs/reference/experimental/index.md?plain=1#L7-L10) and check if you need to add anything for the dataset page
> 
> image

Updated to show properly under dataset. Removed showing from experimental to avoid confusion and follow current doc behaviour.

kolena/_experimental/search/embeddings.py
brianshen3: Added deprecation warnings and use new endpoint

kolena/_experimental/search/embeddings.py
brianshen3: Removed from experimental to avoid confusions. Updated to 1.55.0 as it's expected to be part of next release. Good catch.

kolena/_experimental/search/embeddings.py
brianshen3: Updated removing the duplicate event. Good catch.

docs/reference/dataset/index.md
```diff
@@ -7,3 +7,7 @@
 
 ::: kolena.dataset.dataset
 ::: kolena.dataset.evaluation
+::: kolena.dataset.search
+    options:
+        show_root_heading: true
+        members: ["upload_dataset_embeddings"]
```
nankolena: nit - let's remove these options to be consistent with others, currently this is the only method that's a level under

image

kolena/_experimental/search/embeddings.py
nankolena: Making sure we're on the same page here - is the plan to revert the change on `upload_embeddings`?

brianshen3: Yes, I will revert it.

docs/reference/dataset/index.md
brianshen3: Fixed this

kolena/_experimental/search/embeddings.py
nankolena: This doesn't look reverted still, maybe consider something like these changes? https://github.com/kolenaIO/kolena/compare/brianshen/kol-7957-move-upload-embeddings-out-of-the-experimental-section...nan/kol-7957

brianshen3: Thanks for the branch reference. I followed that and updated the PR.

tests/integration/dataset/search/test_embeddings.py
```diff
@@ -18,8 +18,8 @@
 import pytest
```
nankolena: nit: also bump this test file up similar to the structure in `kolena/dataset/embeddings.py`, and then we should be able to remove the `integration-test-dataset-search-embedding-<< matrix.python-version >>-<< matrix.extras >>` from the circleci config file because the following should take care of it: https://github.com/kolenaIO/kolena/blob/ed91015a9fc6f17f6652819d2cced84ea9d30471/.circleci/continue_config.yml#L268-L274

docs/reference/dataset/index.md
```diff
@@ -7,3 +7,7 @@
 
 ::: kolena.dataset.dataset
 ::: kolena.dataset.evaluation
+::: kolena.dataset.embeddings
+    options:
+        members: ["upload_dataset_embeddings"]
+        show_root_heading: false
```
nankolena, line 13: nit: I feel like these options are not needed here, could you verify whether removing them will not affect anything and if so get rid of these lines?

nankolena (commented):
Only a couple small nits - should be good once those are addressed!

docs/reference/dataset/index.md
brianshen3: The reason I added the option was because the DatasetEmbeddingsDataFrameSchema shows up.
Screenshot 2025-02-12 at 10 11 24 AM

tests/integration/dataset/search/test_embeddings.py
brianshen3: I removed the `integration-test-dataset-search-embedding-<< matrix.python-version >>-<< matrix.extras >>`. Seems to be a duplicate.

nankolena: Could you also bump this integration test one folder up and get rid of the `/search` dir similar to how you did for `kolena/dataset/embeddings.py`?

nankolena (approved):

